PR #36, opened 2020-03-23
set fill value to max of full dataset + 1

jonathanburns:
Hello, I'm somewhat new to this repo 👋 . 

It looks like if the global max is in not in the current cross_validation split,  using `X[train_indices]` would lead to different fill values in different CV splits. My impression is that this is undesired, but correct me if I am wrong.

Similarly, `all_nan_columns` should be the same across all CV splits, I think?


LMZimmer:
Hi there,
I believe the idea was that if NaNs occur for a feature in the validation set but not the train set, we can still train on the train set with that feature (in particular for `.refit`). However, this might through errors when trying to validate so I also think this is cleaner.

PR #78, opened 2021-01-23
Allow specifying the network type in include

franchuterivera:
- Fixes issue where doing pop removes element from the instances
- Allows to put in networks in include
- Adds softmax to output prediction


test/test_pipeline/components/test_setup_networks.py
```diff
@@ -0,0 +1,65 @@
...
+
+        # Make sure that performance was properly captured
+        assert run_summary.performance_tracker['train_loss'][1] > 0
+        assert run_summary.total_parameter_count > 0
+        assert 'accuracy' in run_summary.performance_tracker['train_metrics'][1]
+
+        # Commented out the next line as some pipelines are not
+        # achieving this accuracy with default configuration and 10 epochs
+        # To be added once we fix the search space
+        # assert run_summary.performance_tracker['val_metrics'][fit_dictionary['epochs']]['accuracy'] >= 0.8
```
franchuterivera, line 63: can you create an issue for this? Yeah I think this is one of the most important things to fix

test/test_pipeline/components/test_setup_networks.py
```diff
@@ -0,0 +1,65 @@
...
+from autoPyTorch.pipeline.tabular_classification import TabularClassificationPipeline
+
+
+# Disable
+def blockPrint():
+    sys.stdout = open(os.devnull, 'w')
+
+
+# Restore
+def enablePrint():
```
franchuterivera, line 17: what is this haha

test/test_api/test_api.py
```diff
@@ -27,8 +27,8 @@
 # Test
 # ========
 @pytest.mark.parametrize('openml_id', (40981, ))
-@pytest.mark.parametrize('resampling_strategy', (HoldoutValTypes.holdout_validation,
-                                                 CrossValTypes.k_fold_cross_validation, ))
+@pytest.mark.parametrize('resampling_strategy', (CrossValTypes.k_fold_cross_validation,
```
franchuterivera, line 30: why removing holdout?

ravinkohli: :P I'll fix this, I did this for debugging

test/test_pipeline/components/test_setup_networks.py
ravinkohli: I added this as the traditional classifiers had way too many things printed but then I added it to base model I think when the classifiers are fitted. I'll remove them from here

test/test_pipeline/components/test_setup_networks.py
ravinkohli: sure

test/test_pipeline/components/test_setup_networks.py
ravinkohli: oh but this is useless here :P

PR #80, opened 2021-01-28
Search space update

ravinkohli:
Added HyperparameterSearchSpaceUpdates to alter the configuration space of the pipeline. See examples/example_tabular_classification.py for more info on how to use it.

### DIFF
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -182,11 +182,10 @@ class BaseTask:
 
         self.search_space_updates = search_space_updates
         if search_space_updates is not None:
-            assert isinstance(self.search_space_updates,
-                              HyperparameterSearchSpaceUpdates), "Expected search space updates to be of instance" \
-                                                                 " HyperparameterSearchSpaceUpdates got {}".format(
-                type(self.search_space_updates)
-            )
+            if not isinstance(self.search_space_updates,
+                              HyperparameterSearchSpaceUpdates):
+                raise ValueError("Expected search space updates to be of instance"
+                                 " HyperparameterSearchSpaceUpdates got {}".format(type(self.search_space_updates)))
 
     @abstractmethod
     def _get_required_dataset_properties(self, dataset: BaseDataset) -> Dict[str, Any]:
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -180,9 +180,13 @@ class BaseTask:
 
         self.stop_logging_server = None  # type: Optional[multiprocessing.synchronize.Event]
 
-        if search_space_updates is not None and \
-                isinstance(search_space_updates, HyperparameterSearchSpaceUpdates):
-            self.search_space_updates = search_space_updates
+        self.search_space_updates = search_space_updates
+        if search_space_updates is not None:
+            assert isinstance(self.search_space_updates,
+                              HyperparameterSearchSpaceUpdates), "Expected search space updates to be of instance" \
+                                                                 " HyperparameterSearchSpaceUpdates got {}".format(
+                type(self.search_space_updates)
+            )
 
     @abstractmethod
     def _get_required_dataset_properties(self, dataset: BaseDataset) -> Dict[str, Any]:
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -180,7 +180,9 @@ class BaseTask:
 
         self.stop_logging_server = None  # type: Optional[multiprocessing.synchronize.Event]
 
-        self.search_space_updates = search_space_updates
+        if search_space_updates is not None and \
+                isinstance(search_space_updates, HyperparameterSearchSpaceUpdates):
+            self.search_space_updates = search_space_updates
 
     @abstractmethod
     def _get_required_dataset_properties(self, dataset: BaseDataset) -> Dict[str, Any]:
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -256,7 +256,8 @@ class BaseTask:
                 info=self._get_required_dataset_properties(dataset))
             return get_configuration_space(info=dataset.get_dataset_properties(dataset_requirements),
                                            include=self.include_components,
-                                           exclude=self.exclude_components)
+                                           exclude=self.exclude_components,
+                                           search_space_updates=self.search_space_updates)
         raise Exception("No search space initialised and no dataset passed. "
                         "Can't create default search space without the dataset")
 
@@ -820,7 +821,8 @@ class BaseTask:
                 pipeline_config={**self.pipeline_options, **budget_config},
                 ensemble_callback=proc_ensemble,
                 logger_port=self._logger_port,
-                start_num_run=num_run
+                start_num_run=num_run,
+                search_space_updates=self.search_space_updates
             )
             try:
                 self.run_history, self.trajectory, budget_type = \
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -46,6 +46,7 @@ from autoPyTorch.pipeline.components.training.metrics.base import autoPyTorchMet
 from autoPyTorch.pipeline.components.training.metrics.utils import calculate_score, get_metrics
 from autoPyTorch.utils.backend import Backend, create
 from autoPyTorch.utils.common import FitRequirement, replace_string_bool_to_bool
+from autoPyTorch.utils.hyperparameter_search_space_update import HyperparameterSearchSpaceUpdates
 from autoPyTorch.utils.logging_ import (
     PicklableClientLogger,
     get_named_client_logger,
@@ -135,6 +136,7 @@ class BaseTask:
             include_components: Optional[Dict] = None,
             exclude_components: Optional[Dict] = None,
             backend: Optional[Backend] = None,
+            search_space_updates: Optional[HyperparameterSearchSpaceUpdates] = None
     ) -> None:
         self.seed = seed
         self.n_jobs = n_jobs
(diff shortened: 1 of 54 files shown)


autoPyTorch/pipeline/base_pipeline.py
```diff
@@ -52,21 +53,23 @@ class BasePipeline(Pipeline):
...
     def __init__(
-        self,
-        config: Optional[Configuration] = None,
-        steps: Optional[List[Tuple[str, autoPyTorchChoice]]] = None,
-        dataset_properties: Optional[Dict[str, Any]] = None,
-        include: Optional[Dict[str, Any]] = None,
-        exclude: Optional[Dict[str, Any]] = None,
-        random_state: Optional[np.random.RandomState] = None,
-        init_params: Optional[Dict[str, Any]] = None
+            self,
```
franchuterivera, line 56: I think this is over indented -- I do not know why flake did not fail here...

autoPyTorch/pipeline/base_pipeline.py
```diff
@@ -138,9 +141,9 @@ def predict(self, X: np.ndarray, batch_size: Optional[int] = None
         return self.named_steps['network'].predict(loader)
 
     def set_hyperparameters(
-        self,
-        configuration: Configuration,
-        init_params: Optional[Dict] = None
+            self,
```
franchuterivera, line 144: same here -- overindented

autoPyTorch/pipeline/base_pipeline.py
```diff
@@ -250,12 +253,12 @@ def __repr__(self) -> str:
         return string
 
     def _get_base_search_space(
-        self,
-        cs: ConfigurationSpace,
-        dataset_properties: Dict[str, Any],
-        include: Optional[Dict[str, Any]],
-        exclude: Optional[Dict[str, Any]],
-        pipeline: List[Tuple[str, autoPyTorchChoice]]
+            self,
```
franchuterivera, line 256: same here -- overindented

autoPyTorch/pipeline/base_pipeline.py
```diff
@@ -281,6 +284,12 @@ def _get_base_search_space(
                 raise ValueError('Invalid key in exclude: %s; should be one '
                                  'of %s' % (key, keys))
 
+        if self.search_space_updates is not None and isinstance(self.search_space_updates,
```
franchuterivera, line 287: Ho can it not be an instance of HyperparameterSearchSpaceUpdates? Asking for instance isn't very fast... probably we should query this in the base task once.

ravinkohli: okay that makes sense.

autoPyTorch/pipeline/base_pipeline.py
ravinkohli: I didnt know about it and got no error. I'll fix it

autoPyTorch/api/base_task.py
```diff
@@ -178,6 +180,14 @@ def __init__(
 
         self.stop_logging_server = None  # type: Optional[multiprocessing.synchronize.Event]
 
+        self.search_space_updates = search_space_updates
+        if search_space_updates is not None:
+            assert isinstance(self.search_space_updates,
```
franchuterivera, line 185: Now this should be a ValueError... In case the user or someone wants to catch this

autoPyTorch/pipeline/base_pipeline.py
```diff
@@ -302,9 +310,11 @@ def _get_base_search_space(
             # if the node isn't a choice we can add it immediately because it
             #  must be active (if it wasn't, np.sum(matches) would be zero
             if not is_choice:
+                # for mypy
+                assert not isinstance(node, autoPyTorchChoice)
```
franchuterivera, line 314: Is this really needed? no need to change, but I wonder why you need this because is_choice is the same as this statement.

autoPyTorch/pipeline/components/base_component.py
```diff
@@ -252,3 +252,36 @@ def __str__(self) -> str:
...
+            dict -- Mapping of search space updates. Keys don't contain the prefix.
+        """
+        if prefix is None:
+            return self._cs_updates
+        result: Dict[str, Tuple] = dict()
+
+        # iterate over all search space updates of this node and filter the ones out, that have the given prefix
+        for key in self._cs_updates.keys():
+            if key.startswith(prefix):
+                result[key[len(prefix):]] = self._cs_updates[key]
```
franchuterivera, line 286: In the base choice, you added `len(prefix):+1` whereas here `len(prefix):`. Why is that so? Can you add a comment in the code for this?
Also, can you update the comment:
```
# iterate over all search space updates of this node and filter the ones out, that have the given prefix
```

Shouldn't it indicate that you are including the ones with the prefix not filtering them out?

autoPyTorch/pipeline/components/setup/augmentation/image/Resize.py
```diff
@@ -35,7 +35,8 @@ def fit(self, X: Dict[str, Any], y: Any = None) -> BaseImageAugmenter:
 
     @staticmethod
     def get_hyperparameter_search_space(
-            dataset_properties: Optional[Dict[str, str]] = None
+            dataset_properties: Optional[Dict[str, str]] = None,
+            use_augmenter: Tuple[Tuple, bool] = ((True, False), True),
     ) -> ConfigurationSpace:
         cs = ConfigurationSpace()
         use_augmenter = CategoricalHyperparameter('use_augmenter', choices=[True, False], default_value=True)
```
franchuterivera, line 42: I think here you missed to use `use_augmenter`

ravinkohli: i sure did, thanks

autoPyTorch/pipeline/components/setup/network_backbone/base_network_backbone_choice.py
```diff
@@ -47,10 +46,10 @@ def get_components(self) -> Dict[str, autoPyTorchComponent]:
         return components
 
     def get_available_components(
-        self,
-        dataset_properties: Optional[Dict[str, str]] = None,
-        include: List[str] = None,
-        exclude: List[str] = None,
+            self,
```
franchuterivera, line 49: I think there is an overidentation here

autoPyTorch/pipeline/components/setup/network_backbone/base_network_backbone_choice.py
```diff
@@ -116,11 +115,11 @@ def get_available_components(
         return components_dict
 
     def get_hyperparameter_search_space(
-        self,
-        dataset_properties: Optional[Dict[str, str]] = None,
-        default: Optional[str] = None,
-        include: Optional[List[str]] = None,
-        exclude: Optional[List[str]] = None,
+            self,
```
franchuterivera, line 118: and here also

test/test_pipeline/test_tabular_classification.py
```diff
@@ -186,3 +192,26 @@ def test_get_fit_requirements(self, fit_dictionary):
         assert isinstance(fit_requirements, list)
         for requirement in fit_requirements:
             assert isinstance(requirement, FitRequirement)
+
+    def test_apply_search_space_updates(self, fit_dictionary, search_space_updates):
```
franchuterivera, line 196: I think this test is great, do not change it, but also add 2 additional tests:

- Take the whole configuration space, and convert it to dict, store it. Use it to create a new configuration space update where we only leave one choice dynamically for choices, and for uniform/int hyperparameters, make the range 0 and 1. Then get the configuration space using get configuration space, convert it to dict and make sure that the ranges change accordingly.

2. Put a broken configuration update so that it raises an error

autoPyTorch/utils/hyperparameter_search_space_update.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def save_as_file(self, path: str) -> None:
+        with open(path, "w") as f:
+            for update in self.updates:
+                print(update.node_name, update.hyperparameter, str(update.value_range),  # noqa
+                      str(update.default_value) + (" log" if update.log else ""),
+                      file=f)
+
+
+def parse_hyperparameter_search_space_updates(updates_file: Optional[str]
```
franchuterivera, line 50: Can you add a test where you write the file to disk, read it back and use it to update the config space?

franchuterivera (changes requested):
Thanks a lt, this looks really good and helpful to test the new code. I added some incremental enhancements request on top of it.

autoPyTorch/pipeline/components/base_component.py
ravinkohli: Its because component hyperparameters will not have a delimiter. I'll add this to a comment

test/test_pipeline/test_tabular_classification.py
ravinkohli: could you explain the first point? I think the .__choice__ hyperparameter is chosen using include exclude. Is that what you want to test? Both include and search space update run together?

franchuterivera: Yeah, test them both together but also want to make sure that every thing can be updated. For instance, there was a typo with one config space... The only way to catch this is if we create a test that changes every element of the config space and see if it changed. Makes sense?

autoPyTorch/pipeline/base_pipeline.py
ravinkohli: It assumes the node is autopytorch choice and then gives an arg-type error

test/test_pipeline/test_tabular_classification.py
ravinkohli: yeah I have added this in the latest push

PR #81, opened 2021-02-01
Network Cleanup

bastiscode:
- split image and time series networks into separate files
- unified `get_hyperparameter_search_space` function signatures
- removed old `supported_tasks` attributes from backbones and heads
- updated doc strings in base backbone, base head and some backbones


autoPyTorch/pipeline/components/setup/network_backbone/ConvNetImageBackbone.py
```diff
@@ -0,0 +1,127 @@
...
+from ConfigSpace.hyperparameters import (
+    CategoricalHyperparameter,
+    UniformIntegerHyperparameter
+)
+
+from torch import nn
+
+from autoPyTorch.pipeline.components.setup.network_backbone.base_network_backbone import NetworkBackboneComponent
+
+_activations: Dict[str, nn.Module] = {
```
franchuterivera, line 14: Why not import it from [here](https://github.com/automl/Auto-PyTorch/blob/ddc0f3d890c5a377132108ea03cb23bc093ce17e/autoPyTorch/pipeline/components/setup/network_backbone/ShapedMLPBackbone.py#L17)

bastiscode: Right, I will change it

franchuterivera (changes requested):
Can you create testing for the different backbones?

What I mean by that is that we can create a pipeline so that we include the backbones `TCNBackbone` for example using something like:

```
pipeline = TabularClassificationPipeline(
        dataset_properties=fit_dictionary['dataset_properties'],
        include={
            'network_backbone': ['MLPBackbone'],
        }
    )
```

This is to make sure that the code is not broken and we can at least instantiate a pipeline. we can further improve/check score and so on, but at least to make sure that each backbone can be instantiated.

PR #82, opened 2021-02-02
Make sure the performance of pipeline is at least 0.8

franchuterivera:
Makes sure that every tabular classification network is able to fit on dummy data and produce a good validation performance.


ravinkohli (approved):
I think it can be merged once the checks are passed. Maybe the network is overfitting as the accuracy after 4 epochs is already higher than 80 for ShapedResNetBackbone

ravinkohli (approved):

PR #84, opened 2021-02-02
Reduce Deadlock Probability

franchuterivera:
Use the logger port instead of the logger for the TAE execution.

Add show_models() for debug purposes 

Remove tensorboard output because killing a run in the process of writing to disk halts the complete search process and python does not handle the recovery nice. This is something we should look fixing in pynisher.

Minor fixes like empty cuda in when not needed.

### DIFF
--- a/autoPyTorch/pipeline/components/setup/traditional_ml/classifier_models/classifiers.py
+++ b/autoPyTorch/pipeline/components/setup/traditional_ml/classifier_models/classifiers.py
@@ -106,6 +106,13 @@ class LGBModel(BaseClassifier):
         y_pred = self.model.predict(X_test)
         return y_pred
 
+    @staticmethod
+    def get_properties(dataset_properties: Optional[Dict[str, Any]] = None) -> Dict[str, str]:
+        return {
+            'shortname': 'LGBMClassifier',
+            'name': 'LGBMClassifier',
+        }
+
 
 class CatboostModel(BaseClassifier):
 
@@ -167,6 +174,13 @@ class CatboostModel(BaseClassifier):
         y_pred = self.model.predict(X_test)
         return y_pred
 
+    @staticmethod
+    def get_properties(dataset_properties: Optional[Dict[str, Any]] = None) -> Dict[str, str]:
+        return {
+            'shortname': 'CatBoostClassifier',
+            'name': 'CatBoostClassifier',
+        }
+
 
 class RFModel(BaseClassifier):
 
@@ -225,6 +239,13 @@ class RFModel(BaseClassifier):
             return self.model.predict_proba(X_test)
         y_pred = self.model.predict(X_test)
         return y_pred
+
+    @staticmethod
+    def get_properties(dataset_properties: Optional[Dict[str, Any]] = None) -> Dict[str, str]:
+        return {
+            'shortname': 'RandomForestClassifier',
+            'name': 'RandomForestClassifier',
+        }
 
 
 class ExtraTreesModel(BaseClassifier):
@@ -285,6 +306,13 @@ class ExtraTreesModel(BaseClassifier):
         y_pred = self.model.predict(X_test)
         return y_pred
 
+    @staticmethod
+    def get_properties(dataset_properties: Optional[Dict[str, Any]] = None) -> Dict[str, str]:
+        return {
+            'shortname': 'ExtraTreesClassifier',
+            'name': 'ExtraTreesClassifier',
+        }
+
 
 class KNNModel(BaseClassifier):
 
@@ -338,6 +366,13 @@ class KNNModel(BaseClassifier):
         y_pred = self.model.predict(X_test)
         return y_pred
 
+    @staticmethod
+    def get_properties(dataset_properties: Optional[Dict[str, Any]] = None) -> Dict[str, str]:
+        return {
+            'shortname': 'KNeighborsClassifier',
+            'name': 'KNeighborsClassifier',
+        }
+
 
 class SVMModel(BaseClassifier):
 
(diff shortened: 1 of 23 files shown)


autoPyTorch/pipeline/components/setup/traditional_ml/classifier_models/classifiers.py
```diff
@@ -106,6 +106,13 @@ def predict(self, X_test: np.ndarray, predict_proba: bool = False) -> np.ndarray
         y_pred = self.model.predict(X_test)
         return y_pred
 
+    @staticmethod
+    def get_properties(dataset_properties: Optional[Dict[str, Any]] = None) -> Dict[str, str]:
+        return {
+            'shortname': 'LGBMClassifier',
+            'name': 'LGBMClassifier',
```
ravinkohli, line 113: could we have the full name here?

autoPyTorch/pipeline/components/setup/traditional_ml/classifier_models/classifiers.py
```diff
@@ -226,6 +240,13 @@ def predict(self, X_test: np.ndarray, predict_proba: bool = False) -> np.ndarray
         y_pred = self.model.predict(X_test)
         return y_pred
 
+    @staticmethod
+    def get_properties(dataset_properties: Optional[Dict[str, Any]] = None) -> Dict[str, str]:
+        return {
+            'shortname': 'RandomForestClassifier',
```
ravinkohli, line 246: could we make it RFClassifier

autoPyTorch/pipeline/components/setup/traditional_ml/classifier_models/classifiers.py
```diff
@@ -338,6 +366,13 @@ def predict(self, X_test: np.ndarray, predict_proba: bool = False) -> np.ndarray
         y_pred = self.model.predict(X_test)
         return y_pred
 
+    @staticmethod
+    def get_properties(dataset_properties: Optional[Dict[str, Any]] = None) -> Dict[str, str]:
+        return {
+            'shortname': 'KNeighborsClassifier',
```
ravinkohli, line 372: can this be KNNClassifier

autoPyTorch/pipeline/components/setup/traditional_ml/classifier_models/classifiers.py
```diff
@@ -338,6 +366,13 @@ def predict(self, X_test: np.ndarray, predict_proba: bool = False) -> np.ndarray
...
+            'name': 'KNeighborsClassifier',
```
ravinkohli, line 373: can this be K Nearest Neighbours Classifier

autoPyTorch/pipeline/components/setup/traditional_ml/classifier_models/classifiers.py
```diff
@@ -384,3 +419,10 @@ def predict(self, X_test: np.ndarray, predict_proba: bool = False) -> np.ndarray
             return self.model.predict_proba(X_test)
         y_pred = self.model.predict(X_test)
         return y_pred
+
+    @staticmethod
+    def get_properties(dataset_properties: Optional[Dict[str, Any]] = None) -> Dict[str, str]:
+        return {
+            'shortname': 'SVC',
+            'name': 'SVC',
```
ravinkohli, line 427: Can this be Support Vector Classification

autoPyTorch/pipeline/tabular_classification.py
```diff
@@ -225,6 +227,39 @@ def _get_pipeline_steps(self, dataset_properties: Optional[Dict[str, Any]],
...
+
+        It should be a representation that follows:
+        [{'PreProcessing': <>, 'Estimator': <>}]
+
+        Returns:
+            Dict: contains the pipeline representation in a short format
+        """
+        preprocessing = []
+        estimator = []
+        skip_steps = ['data_loader', 'trainer', 'lr_scheduler', 'optimizer', 'network_init',
```
ravinkohli, line 243: Maybe we can have a verbose option where we include- trainer, lr_scheduler, optimizer and network_init?

autoPyTorch/pipeline/tabular_regression.py
```diff
@@ -172,6 +173,39 @@ def _get_pipeline_steps(self, dataset_properties: Optional[Dict[str, Any]]) -> L
...
+
+        It should be a representation that follows:
+        [{'PreProcessing': <>, 'Estimator': <>}]
+
+        Returns:
+            Dict: contains the pipeline representation in a short format
+        """
+        preprocessing = []
+        estimator = []
+        skip_steps = ['data_loader', 'trainer', 'lr_scheduler', 'optimizer', 'network_init',
```
ravinkohli, line 189: Same here

ravinkohli (commented):
The PR looks great, I have just added minor comments that will give a bit more information to the user.

autoPyTorch/pipeline/tabular_classification.py
franchuterivera: I was coding this and seeing the outcome, and it just doesn't add much information to 'what the best estimator is' because it's part of the construction of the model, but not part of the model itself.

I have a better proposal -- I would like to create a command like [this](https://github.com/EpistasisLab/tpot/blob/6448bdb71ba08b4a0447c640d2f05a05e1affc21/tpot/export_utils.py#L56).

Tpot is able to print the python code on how to train the model that you did. So if your goal were to see what happened (and for debug purposes) it would be great to produce a file that contains PyTorch commands with not only the scheduler but also the config (like it is nice to know we know more that the fact that we picked `adam` optimizers). If this is better, then I would like to disentangle this export_pipeline command from this PR and create an issue for this.
What do you think?

ravinkohli: yes this would be ideal to use. Could you add an issue so we don't forget?

PR #85, opened 2021-02-04
Adding tabular regression pipeline

bastiscode:
PR includes:
- tabular regression pipeline
- fixes for other components to get the pipeline to work, such as.:
     - better device handling for base network and trainer
     - casting targets to correct data type (long or float) depending on task type in the trainer
     - handling single batch / unbatched inputs to column transformer
- refactored tests for classification and regression pipeline


autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/TabularColumnTransformer.py
```diff
@@ -91,4 +91,9 @@ def __call__(self, X: Union[np.ndarray, torch.tensor]) -> Union[np.ndarray, torc
         if self.preprocessor is None:
             raise ValueError("cant call {} without fitting the column transformer first."
                              .format(self.__class__.__name__))
+
+        if len(X.shape) == 1:
```
franchuterivera, line 95: Do you think we should support this case? having 1 feature as a column is not a very interesting case to produce targets. Many things fail downstream I think because of that.

I was actually planning to add a check to ensure at least 2D in the inputs.

autoPyTorch/pipeline/components/setup/network/base_network.py
```diff
@@ -20,15 +20,11 @@ class NetworkComponent(autoPyTorchTrainingComponent):
 
     def __init__(
             self,
-            network: Optional[torch.nn.Module] = None,
```
franchuterivera, line 23: I think we should keep the network here.
So what happens with base estimators is, that when clone, they honor the set_params/get_params scheme from scikit learn described [here](https://scikit-learn.org/stable/developers/develop.html). In summary, an object gets re-created using the class-type and whatever get_params returns. Get_params returns the contents of `vars` with some filtering.

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
```diff
@@ -97,11 +98,11 @@ def get_components(self) -> Dict[str, autoPyTorchComponent]:
         return components
 
     def get_hyperparameter_search_space(
-        self,
-        dataset_properties: Optional[Dict[str, str]] = None,
-        default: Optional[str] = None,
-        include: Optional[List[str]] = None,
-        exclude: Optional[List[str]] = None,
+            self,
```
franchuterivera, line 101: I think this is over-indented?

autoPyTorch/pipeline/tabular_classification.py
```diff
@@ -243,7 +243,7 @@ def _get_pipeline_steps(self, dataset_properties: Optional[Dict[str, Any]],
             ("preprocessing", EarlyPreprocessing()),
             ("network_backbone", NetworkBackboneChoice(default_dataset_properties)),
             ("network_head", NetworkHeadChoice(default_dataset_properties)),
-            ("network", NetworkComponent(default_dataset_properties)),
```
franchuterivera, line 246: Do you think we can try to keep the same interface to other components?

autoPyTorch/pipeline/tabular_regression.py
```diff
@@ -59,25 +59,25 @@ class TabularRegressionPipeline(RegressorMixin, BasePipeline):
...
-        self,
-        config: Optional[Configuration] = None,
-        steps: Optional[List[Tuple[str, autoPyTorchChoice]]] = None,
-        dataset_properties: Optional[Dict[str, Any]] = None,
-        include: Optional[Dict[str, Any]] = None,
-        exclude: Optional[Dict[str, Any]] = None,
-        random_state: Optional[np.random.RandomState] = None,
-        init_params: Optional[Dict[str, Any]] = None,
-        search_space_updates: Optional[HyperparameterSearchSpaceUpdates] = None
+            self,
```
franchuterivera, line 62: Isn't this over-indented?

autoPyTorch/pipeline/tabular_regression.py
```diff
@@ -59,25 +59,25 @@ class TabularRegressionPipeline(RegressorMixin, BasePipeline):
...
         super().__init__(
             config, steps, dataset_properties, include, exclude,
             random_state, init_params, search_space_updates)
 
     def fit_transformer(
-        self,
-        X: np.ndarray,
-        y: np.ndarray,
-        fit_params: Optional[Dict[str, Any]] = None
+            self,
```
franchuterivera, line 77: same here

autoPyTorch/pipeline/tabular_regression.py
```diff
@@ -195,7 +195,7 @@ def _get_pipeline_steps(self, dataset_properties: Optional[Dict[str, Any]],
             ("preprocessing", EarlyPreprocessing()),
             ("network_backbone", NetworkBackboneChoice(default_dataset_properties)),
             ("network_head", NetworkHeadChoice(default_dataset_properties)),
-            ("network", NetworkComponent(default_dataset_properties)),
+            ("network", NetworkComponent()),
```
franchuterivera, line 198: Do you think we can keep the same signature as other components?

autoPyTorch/pipeline/tabular_regression.py
```diff
@@ -59,25 +59,25 @@ class TabularRegressionPipeline(RegressorMixin, BasePipeline):
...
+            self,
+            config: Optional[Configuration] = None,
+            steps: Optional[List[Tuple[str, autoPyTorchChoice]]] = None,
+            dataset_properties: Optional[Dict[str, Any]] = None,
+            include: Optional[Dict[str, Any]] = None,
+            exclude: Optional[Dict[str, Any]] = None,
+            random_state: Optional[np.random.RandomState] = None,
+            init_params: Optional[Dict[str, Any]] = None,
+            search_space_updates: Optional[HyperparameterSearchSpaceUpdates] = None
     ):
```
franchuterivera, line 71: I think we have to update the comment from this file from :
```
available classifiers at runtime.
```
```
available regressors at runtime.
```
If you could just search for classification and replace to actually say regression that will be great

test/test_pipeline/test_tabular_regression.py
```diff
@@ -0,0 +1,268 @@
...
+
+from autoPyTorch.pipeline.components.setup.early_preprocessor.utils import get_preprocess_transforms
+from autoPyTorch.pipeline.tabular_regression import TabularRegressionPipeline
+from autoPyTorch.utils.common import FitRequirement
+from autoPyTorch.utils.hyperparameter_search_space_update import HyperparameterSearchSpaceUpdates, \
+    parse_hyperparameter_search_space_updates
+
+
+@pytest.mark.parametrize("fit_dictionary_tabular", ["regression_numerical_only",
+                                                    # "regression_categorical_only",
```
franchuterivera, line 24: This is not supported yet?

test/test_pipeline/test_tabular_regression.py
```diff
@@ -0,0 +1,268 @@
...
+        cs = pipeline.get_hyperparameter_search_space()
+        config = cs.sample_configuration()
+        pipeline.set_hyperparameters(config)
+
+        pipeline.fit(fit_dictionary_tabular)
+
+        prediction = pipeline.predict(
+            fit_dictionary_tabular['backend'].load_datamanager().test_tensors[0])
+        assert isinstance(prediction, np.ndarray)
+        assert prediction.shape == (200, 1)
```
franchuterivera, line 88: Maybe you can take this from the data, else when changing the fit dictionary this will fail.

test/test_pipeline/test_tabular_regression.py
```diff
@@ -0,0 +1,268 @@
...
+
+        # Make sure that performance was properly captured
+        assert run_summary.performance_tracker['train_loss'][1] > 0
+        assert run_summary.total_parameter_count > 0
+        assert 'r2' in run_summary.performance_tracker['train_metrics'][1]
+
+        # Make sure a network was fit
+        assert isinstance(pipeline.named_steps['network'].get_network(), torch.nn.Module)
+
+    def test_pipeline_predict(self, fit_dictionary_tabular):
```
franchuterivera, line 73: can you check here that at least the score is greater than 0.80?

test/test_pipeline/test_tabular_regression.py
```diff
@@ -0,0 +1,268 @@
...
+        """This test makes sure that the pipeline is able to fit
+        given random combinations of hyperparameters across the pipeline"""
+        pipeline = TabularRegressionPipeline(
+            dataset_properties=fit_dictionary_tabular['dataset_properties'])
+
+        cs = pipeline.get_hyperparameter_search_space()
+        config = cs.sample_configuration()
```
franchuterivera, line 80: To check that the score is at least 0.8, change this to default configuration

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/TabularColumnTransformer.py
bastiscode: I added this because the pipeline failed during predict when inputs are single records, not if they have a single feature. So this turns e.g. a vector of length `F` into a vector of shape `[1, F]` (added batch dimension). But if you add a check that ensures this somewhere else, then I can remove this piece of code.

autoPyTorch/pipeline/components/setup/network/base_network.py
bastiscode: Okay, I will add it again

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
bastiscode: Yes, will change it

autoPyTorch/pipeline/tabular_classification.py
bastiscode: I thought this might be cleaner, since the NetworkComponent similar to the EarlyPreprocessing just aggregates parts of the pipeline (backbone and head for the NetworkComponent, all the preprocessing transformers for EarlyPreprocessing). They dont require dataset properties because of that, and because it was already removed at the EarlyPreprocessing I also removed it here. What do you think?

autoPyTorch/pipeline/tabular_regression.py
bastiscode: I will change it.

autoPyTorch/pipeline/tabular_regression.py
bastiscode: see reply above to similar issue in tabular classification pipeline

test/test_pipeline/test_tabular_regression.py
bastiscode: There is an issue with the CategoricalEncoder implementation, thats why I commented it out for now.

test/test_pipeline/test_tabular_regression.py
bastiscode: Yes, thats a good idea. Will change it in the tabular classification aswell.

test/test_pipeline/test_tabular_regression.py
bastiscode: I mean this depends on the data, what score we expect, right? And what score do you mean for regression, R2?

test/test_pipeline/test_tabular_regression.py
bastiscode: You mean that I should not just randomly sample and set a configuration?

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/TabularColumnTransformer.py
franchuterivera: Yeah I find this similar problem and was planning to fix it [here](https://github.com/automl/Auto-PyTorch/blob/d5bfd42f15a186c7e572d3187005fa52f4073a00/autoPyTorch/pipeline/components/training/data_loader/feature_data_loader.py#L14) in my new PR.

I think it makes sense to try to put all feature modification code in the same place, so debug is easy. So having it close to the dataloader/dataset is ideal. What do you think?

test/test_pipeline/test_tabular_regression.py
franchuterivera: Can you tell me what the issue is?

test/test_pipeline/test_tabular_regression.py
franchuterivera: Oh Yeah, something bound in [0, 1] like R2.

The data that I saw you were using was dummy data right? With the default configuration space we should be able to overfit the data (I think?)

test/test_pipeline/test_tabular_regression.py
franchuterivera: If you can that will be great :D, but for instance, this means that you won't have a consistent test to make sure regression code is working. With the default pipeline and after the randomness is removed, you will have a consistent check.

But maybe you get a very small learning rate, and then never achieve a good performance in the given budget.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/TabularColumnTransformer.py
bastiscode: Yes, thats better. Mine was just a quick fix. We should merge your PR first then, so I can rebase my code.

test/test_pipeline/test_tabular_regression.py
bastiscode: Yes, we should be able to easily overfit.

test/test_pipeline/test_tabular_regression.py
bastiscode: Makes sense.

test/test_pipeline/test_tabular_regression.py
bastiscode: I have to check first if it is still there because I think Ravins PR changed something in this regard.

bastiscode: Issue seems to be resolved!

autoPyTorch/api/tabular_regression.py
```diff
@@ -0,0 +1,249 @@
...
+                type(dataset)
+            ))
+        return {'task_type': dataset.task_type,
+                'output_type': dataset.output_type,
+                'issparse': dataset.issparse,
+                'numerical_columns': dataset.numerical_columns,
+                'categorical_columns': dataset.categorical_columns}
+
+    def build_pipeline(self, dataset_properties: Dict[str, Any]) -> TabularRegressionPipeline:
+        return TabularRegressionPipeline(dataset_properties=dataset_properties)
```
ravinkohli, line 104: I think we should also pass include exclude components to this. Also, we are not doing this in tabular classification, could you also change it there?

autoPyTorch/evaluation/utils.py
```diff
@@ -78,3 +81,23 @@ def convert_multioutput_multiclass_to_multilabel(probas: Union[List, np.ndarray]
                 multioutput_probas[:, i] = 0
         probas = multioutput_probas
     return probas
+
+
+class VotingRegressorWrapper(VotingRegressor):
+    """
+    Wrapper around the sklearn voting regressor that properly handles
```
ravinkohli, line 88: I think you left this incomplete.

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -60,9 +59,9 @@ def is_max_time_reached(self) -> bool:
 
 class RunSummary(object):
     def __init__(
-        self,
-        total_parameter_count: float,
-        trainable_parameter_count: float,
+            self,
```
ravinkohli, line 62: This is over indented if you're using pycharm you can fix it by going to setting > Editor > Cose style > python  and then change continuation indent to 4

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -194,18 +192,17 @@ def prepare(
...
-        if self.weighted_loss:
-            weights = self.get_class_weights(output_type, labels)
-            if output_type == BINARY:
-                kwargs['pos_weight'] = weights
-            else:
-                kwargs['weight'] = weights
-
-        criterion = criterion(**kwargs) if weights is not None else criterion()
+        kwargs: Dict[str, Any] = {}
+        # if self.weighted_loss:
```
ravinkohli, line 196: why have these lines been commented out?

examples/example_tabular_regression.py
```diff
@@ -0,0 +1,116 @@
...
+    X_train, X_test, y_train, y_test = model_selection.train_test_split(
+        X,
+        y,
+        random_state=1,
+    )
+
+    # Scale the regression targets to have zero mean and unit variance.
+    # This is important for Neural Networks since predicting large target values would require very large weights.
+    # One can later rescale the network predictions like this: y_pred = y_pred_scaled * y_train_std + y_train_mean
+    y_train_mean = y_train.mean()
```
ravinkohli, line 82: do you think it is better to do this in the target validator class?

autoPyTorch/api/tabular_regression.py
bastiscode: I can add it after this PR is merged, its too much changes for one PR already in my opinion. But you should definitely do this.

autoPyTorch/evaluation/utils.py
bastiscode: Right

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
bastiscode: That is good to know, thanks

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
bastiscode: Because there was this issue with pos_weight and binary classifciation I talked to you about. Until its resolved I commented it out because it would throw errors on classification with two classes.

examples/example_tabular_regression.py
bastiscode: Open for discussion I would say. We could also put it in a separate component in the regression pipeline.

autoPyTorch/api/tabular_regression.py
ravinkohli: I would say it is a bug in our implementation as if someone wants to include a particular component and then calls refit, the pipeline that will be used would be different from what the user wanted

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
ravinkohli: I dont think we can treat binary classification as multi class classification because the traditional classifiers infer the metric to use and other things by checking the targets. Could you share the error that comes?

bastiscode: The targets are not affected. What I suggested for binary classification was instead of predicting 1 output and using BCEWithLogitsLoss to predict 2 outputs and using standard CrossEntropyLoss, so we dont have to make this distinction with pos_weight and other stuff at all. What do you think?

ravinkohli: I think we are already predicting 2 outputs for binary classification see [here](https://github.com/automl/Auto-PyTorch/blob/a29dbee2b8f1b2520073a9ed1ce5d942c6ddefa6/autoPyTorch/datasets/base_dataset.py#L123). I think the confusion comes from the fact that I am checking if the task is binary and then making it pos_weight, but I think I should check if the criterion is BCELogitsWithLoss then we should use pos_weight, or remove this loss altogether. If you want to, could you change this in this PR? otherwise could you make an issue for this?

bastiscode: I think the code you are linking to is exactly the change from me in this PR to avoid using binary cross entropy.

ravinkohli: yes, but even in the previous version [here](https://github.com/automl/Auto-PyTorch/blob/b48b952a2709f869f6e63511cfa0c94a889f3591/autoPyTorch/pipeline/components/setup/network_head/base_network_head.py#L41), its the same thing that is happening

bastiscode: yep right

PR #89, opened 2021-02-09
Handling Input to auto pytorch 

franchuterivera:
- Enable any input type to autopytorch (`if input in (pd.DataFrame, List, sparse, numpy)`)

- Better error to user and plenty of more testing (included moving dataset to pytest)

### DIFF
--- a/autoPyTorch/api/tabular_classification.py
+++ b/autoPyTorch/api/tabular_classification.py
@@ -1,3 +1,5 @@
+import os
+import uuid
 from typing import Any, Callable, Dict, List, Optional, Union
 
 import numpy as np
@@ -86,10 +88,6 @@ class TabularClassificationTask(BaseTask):
             task_type=TASK_TYPES_TO_STRING[TABULAR_CLASSIFICATION],
         )
 
-        # Create a validator object to make sure that the data provided by
-        # the user matches the autopytorch requirements
-        self.InputValidator = TabularInputValidator(is_classification=True)
-
     def _get_required_dataset_properties(self, dataset: BaseDataset) -> Dict[str, Any]:
         if not isinstance(dataset, TabularDataset):
             raise ValueError("Dataset is incompatible for the given task,: {}".format(
@@ -105,24 +103,25 @@ class TabularClassificationTask(BaseTask):
         return TabularClassificationPipeline(dataset_properties=dataset_properties)
 
     def search(
-            self,
-            optimize_metric: str,
-            X_train: Optional[Union[List, pd.DataFrame, np.ndarray]] = None,
-            y_train: Optional[Union[List, pd.DataFrame, np.ndarray]] = None,
-            X_test: Optional[Union[List, pd.DataFrame, np.ndarray]] = None,
-            y_test: Optional[Union[List, pd.DataFrame, np.ndarray]] = None,
-            budget_type: Optional[str] = None,
-            budget: Optional[float] = None,
-            total_walltime_limit: int = 100,
-            func_eval_time_limit: int = 60,
-            traditional_per_total_budget: float = 0.1,
-            memory_limit: Optional[int] = 4096,
-            smac_scenario_args: Optional[Dict[str, Any]] = None,
-            get_smac_object_callback: Optional[Callable] = None,
-            all_supported_metrics: bool = True,
-            precision: int = 32,
-            disable_file_output: List = [],
-            load_models: bool = True,
+        self,
+        optimize_metric: str,
+        X_train: Optional[Union[List, pd.DataFrame, np.ndarray]] = None,
+        y_train: Optional[Union[List, pd.DataFrame, np.ndarray]] = None,
+        X_test: Optional[Union[List, pd.DataFrame, np.ndarray]] = None,
+        y_test: Optional[Union[List, pd.DataFrame, np.ndarray]] = None,
+        dataset_name: Optional[str] = None,
+        budget_type: Optional[str] = None,
+        budget: Optional[float] = None,
+        total_walltime_limit: int = 100,
+        func_eval_time_limit: int = 60,
+        traditional_per_total_budget: float = 0.1,
+        memory_limit: Optional[int] = 4096,
+        smac_scenario_args: Optional[Dict[str, Any]] = None,
+        get_smac_object_callback: Optional[Callable] = None,
+        all_supported_metrics: bool = True,
+        precision: int = 32,
+        disable_file_output: List = [],
+        load_models: bool = True,
     ) -> 'BaseTask':
         """
         Search for the best pipeline configuration for the given dataset.
@@ -133,9 +132,8 @@ class TabularClassificationTask(BaseTask):
         Args:
             X_train, y_train, X_test, y_test: Union[np.ndarray, List, pd.DataFrame]
                 A pair of features (X_train) and targets (y_train) used to fit a
-                pipeline. Additionally, a holdout of this paris (X_test, y_test) can
+                pipeline. Additionally, a holdout of this pairs (X_test, y_test) can
                 be provided to track the generalization performance of each stage.
-                Providing X_train, y_train and dataset together is not supported.
             optimize_metric (str): name of the metric that is used to
                 evaluate a pipeline.
             budget_type (Optional[str]):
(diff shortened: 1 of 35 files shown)


autoPyTorch/api/tabular_classification.py
```diff
@@ -86,3 +103,151 @@ def _get_required_dataset_properties(self, dataset: BaseDataset) -> Dict[str, An
...
+        """
+        Search for the best pipeline configuration for the given dataset.
+
+        Fit both optimizes the machine learning models and builds an ensemble out of them.
+        To disable ensembling, set ensemble_size==0.
+        using the optimizer.
+        Args:
+            X_train, y_train, X_test, y_test: Union[np.ndarray, List, pd.DataFrame]
+                A pair of features (X_train) and targets (y_train) used to fit a
+                pipeline. Additionally, a holdout of this paris (X_test, y_test) can
```
ravinkohli, line 136: there's a typo in paris, should be pairs

autoPyTorch/api/tabular_classification.py
```diff
@@ -86,3 +103,151 @@ def _get_required_dataset_properties(self, dataset: BaseDataset) -> Dict[str, An
...
+                be provided to track the generalization performance of each stage.
+                Providing X_train, y_train and dataset together is not supported.
```
ravinkohli, line 138: Does it make sense to have this line as the user can not send the dataset here anyways right?

autoPyTorch/api/tabular_classification.py
```diff
@@ -86,3 +103,151 @@ def _get_required_dataset_properties(self, dataset: BaseDataset) -> Dict[str, An
...
+        )
+
+    def predict(
+            self,
+            X_test: np.ndarray,
+            batch_size: Optional[int] = None,
+            n_jobs: int = 1
+    ) -> np.ndarray:
+        if self.InputValidator is None or not self.InputValidator._is_fitted:
+            raise ValueError("predict() is only supported after calling fit. Kindly call first "
```
ravinkohli, line 230: shouldn't it be only supported after calling search?

autoPyTorch/api/tabular_classification.py
```diff
@@ -86,3 +103,151 @@ def _get_required_dataset_properties(self, dataset: BaseDataset) -> Dict[str, An
...
+
+        # Allow to predict in the original domain -- that is, the user is not interested
+        # in our encoded values
+        return self.InputValidator.target_validator.inverse_transform(predicted_indexes)
+
+    def predict_proba(self,
+                      X_test: Union[np.ndarray, pd.DataFrame, List],
+                      batch_size: Optional[int] = None, n_jobs: int = 1) -> np.ndarray:
+        if self.InputValidator is None or not self.InputValidator._is_fitted:
+            raise ValueError("predict() is only supported after calling fit. Kindly call first "
```
ravinkohli, line 250: same here

autoPyTorch/data/base_feature_validator.py
```diff
@@ -0,0 +1,139 @@
...
+    Attributes:
+        feat_type (List[str]):
+            List of the column types found by this estimator during fit.
+        data_type (str):
+            Class name of the data type provided during fit.
+        encoder (typing.Optional[BaseEstimator])
+            Host a encoder object if the data requires transformation (for example,
+            if provided a categorical column in a pandas DataFrame)
+        enc_columns (typing.List[str])
+            List of columns that where encoded.
```
ravinkohli, line 42: were*

autoPyTorch/api/tabular_classification.py
```diff
@@ -69,9 +80,15 @@ def __init__(
...
+            resampling_strategy=resampling_strategy,
+            resampling_strategy_args=resampling_strategy_args,
+            search_space_updates=search_space_updates,
+            task_type=TASK_TYPES_TO_STRING[TABULAR_CLASSIFICATION],
         )
-        self.task_type = TASK_TYPES_TO_STRING[TABULAR_CLASSIFICATION]
+
+        # Create a validator object to make sure that the data provided by
+        # the user matches the autopytorch requirements
+        self.InputValidator = TabularInputValidator(is_classification=True)
```
ravinkohli, line 91: I think we should also pass the logger here?

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -0,0 +1,377 @@
...
+                        return 1
+                    else:
+                        raise ValueError((cmp1, cmp2))
+                self.feat_type = sorted(
+                    self.feat_type,
+                    key=functools.cmp_to_key(comparator)
+                )
+
+                self.categories = [
+                    # We fit an ordinal encoder, where all categorical
```
ravinkohli, line 93: maybe this comment should be added along with comment on line 73

autoPyTorch/pipeline/base_pipeline.py
```diff
@@ -106,6 +106,52 @@ def __init__(
 
         self._additional_run_info = {}  # type: Dict[str, str]
 
+    def fit(self, X: Dict[str, Any], y: Optional[np.ndarray] = None,
+            **fit_params: Any) -> Pipeline:
+        """Fit the selected algorithm to the training data.
+        Arguments:
+            X (typing.Dict):
+            A fit dictionary so that contains information to fit a pipeline
```
ravinkohli, line 114: maybe fit dictionary that contains, without so

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/encoding/OrdinalEncoder.py
```diff
@@ -19,7 +19,9 @@ def fit(self, X: Dict[str, Any], y: Any = None) -> BaseEncoder:
 
         self.check_requirements(X, y)
 
-        self.preprocessor['categorical'] = OE(categories=X['dataset_properties']['categories'])
+        self.preprocessor['categorical'] = OE(handle_unknown='use_encoded_value',
```
ravinkohli, line 22: maybe we dont need an ordinal encoder in the search space as the data has already gone through this phase. In case for a dataset, OneHotEncoding is not preferred, NoEncoder will have the same effect as doing Ordinal encoding.

autoPyTorch/data/tabular_target_validator.py
```diff
@@ -0,0 +1,265 @@
...
+
+        Arguments:
+            y_train (SUPPORTED_TARGET_TYPES)
+                The labels of the current task. They are going to be encoded in case
+                of classification
+            y_test (typing.Optional[SUPPORTED_TARGET_TYPES])
+                A holdout set of labels
+        """
+        if not self.is_classification or self.type_of_target == 'multilabel-indicator':
+            # Only fit an encoder for classification tasks
```
ravinkohli, line 39: Should we also raise an error if in regression the data type of the target is not integer?

ravinkohli (commented):
This PR looks really good. I have left a few minor comments to be addressed.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/encoding/OrdinalEncoder.py
franchuterivera: Oh no wait, we do not touch the numerical columns. The objective here is to make the data numerical.

So if you have a numerical columns with values like 0.0000000001 and so on, maybe changing that to 1, 2, 3, 4 and so one with an ordinal encoder improves the performance.

autoPyTorch/data/tabular_target_validator.py
franchuterivera: I think #85 can handle this. Now I do not know what you mean exactly, as we have different regression targets. Usually we expected regression to have float type, which is something that the code handles right now.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/encoding/OrdinalEncoder.py
ravinkohli: but in its current form, its just working on the categorical columns. Also, that could be handled with the scaler component. Me and sebastian also discussed today about not having an encoding component. As when we merge the embedding PR which I'll finish soon, PyTorch has this nn.Embedding layer which performs OneHotEncoding and learns the embedding all by itself.

autoPyTorch/data/tabular_target_validator.py
ravinkohli: Oh I meant if somehow someone gives y values as a string. Which does not contain numbers but words, by mistake. I think in this validator module we should take care of this as well

PR #91, opened 2021-02-15
Embedding layer

ravinkohli:
This pull request adds entity embeddings and removes ordinal encoder from the search space.


autoPyTorch/pipeline/components/setup/network_embedding/LearnedEntityEmbedding.py
```diff
@@ -0,0 +1,132 @@
...
+            categorical_feature_slice = x[:, x_pointer: x_pointer + num_in]
+            concat_seq.append(self.ee_layers[layer_pointer](categorical_feature_slice))
+            layer_pointer += 1
+            x_pointer += num_in
+            last_concat = x_pointer
+
+        concat_seq.append(x[:, last_concat:])
+        return torch.cat(concat_seq, dim=1)
+
+    def _create_ee_layers(self) -> nn.ModuleList:
```
bastiscode, line 75: Why not using `nn.Embedding` here?

ravinkohli: It is essentially the same thing here as well, so I decided to stick with what Lucas had in the older autopytorch code

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/TabularColumnTransformer.py
```diff
@@ -57,9 +57,9 @@ def fit(self, X: Dict[str, Any], y: Any = None) -> "TabularColumnTransformer":
         if len(X['dataset_properties']['categorical_columns']):
             categorical_pipeline = make_pipeline(*preprocessors['categorical'])
 
-        self.preprocessor = make_column_transformer(
-            (numerical_pipeline, X['dataset_properties']['numerical_columns']),
-            (categorical_pipeline, X['dataset_properties']['categorical_columns']),
+        self.preprocessor = ColumnTransformer([
+            ('numerical_pipeline', numerical_pipeline, X['dataset_properties']['numerical_columns']),
```
franchuterivera, line 61: If there are only categorical columns, numerical_pipeline is set to drop. Does it make sense?

Can we enhance [this](https://github.com/automl/Auto-PyTorch/blob/8783240e2ef818d4cbb3447ae89a1319e14408c4/test/test_pipeline/components/preprocessing/test_tabular_column_transformer.py#L20) check to make sure all columns are available after the transformer?

test/test_pipeline/test_losses.py
```diff
@@ -37,6 +37,7 @@ def test_get_name_error():
 
 @pytest.mark.parametrize('weighted', [True, False])
 def test_losses(weighted):
+    torch.manual_seed(1)
```
franchuterivera, line 40: why did you have to seed the training? is m wondering if we should do this (if needed) in conftest

autoPyTorch/pipeline/tabular_classification.py
```diff
@@ -216,6 +246,7 @@ def _get_pipeline_steps(self, dataset_properties: Optional[Dict[str, Any]],
             ("feature_preprocessor", FeatureProprocessorChoice(default_dataset_properties)),
             ("tabular_transformer", TabularColumnTransformer()),
             ("preprocessing", EarlyPreprocessing()),
+            ("network_embedding", NetworkEmbeddingChoice(default_dataset_properties)),
```
franchuterivera, line 249: I guess this is also needed in regression?

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/TabularColumnTransformer.py
ravinkohli: I think it makes sense as all the columns would be categorical and they would be passed through the categorical pipeline and drop would not do anything. I'll add the test to make sure all the columns are passed

test/test_pipeline/test_losses.py
ravinkohli: basically to determine the weights for each class the labels generated should contain all 4 classes. But because its random sometimes the labels array does not have a class and this causes weights to be inferred for only three classes and gives an error when we try to use them in cross entropy

autoPyTorch/pipeline/tabular_classification.py
ravinkohli: yeah thanks for pointing out, I think I forgot

franchuterivera:
Hi @ravinkohli , are this conflicts coming from the regression or merge? Also, I have a request. If you see [this](https://github.com/automl/Auto-PyTorch/runs/1927214088), looks like the main test is failing because of an issue with regression. I wonder if cholesterol is too complicated and we should probably *for pytest* move to [this](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.load_boston.html) OR maybe add more time to the test.

ravinkohli:
> Hi @ravinkohli , are this conflicts coming from the regression or merge? Also, I have a request. If you see [this](https://github.com/automl/Auto-PyTorch/runs/1927214088), looks like the main test is failing because of an issue with regression. I wonder if cholesterol is too complicated and we should probably _for pytest_ move to [this](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.load_boston.html) OR maybe add more time to the test.

Hey, yes the merge conflicts were from the regression pipeline, I have resolved them now. I think moving the load_boston is a better option as we only want to test if API is working.

test/test_evaluation/test_evaluation.py
```diff
@@ -380,6 +380,10 @@ def test_silent_exception_in_target_function(self):
                 """'save_targets_ensemble'",)""",
                 """AttributeError("'BackendMock' object has no attribute """
                 """'save_targets_ensemble'")""",
+                """AttributeError("'BackendMock' object has no attribute """
```
franchuterivera, line 383: I cannot spot the difference, which one is it?

test/utils.py
```diff
@@ -0,0 +1,71 @@
+from pathlib import Path
+
+
+class DisplayablePath(object):
```
franchuterivera, line 4: This is no longer needed?

ravinkohli: yeah it is not, I thought I deleted it, thanks for pointing out

test/test_evaluation/test_evaluation.py
ravinkohli: what do you mean? So I realised that we were never setting up the logger in the backend and therefore I added a line [here](https://github.com/ravinkohli/Auto-PyTorch/blob/36ae93c6a8bf83c8ecca8601946e473213a184b4/autoPyTorch/evaluation/abstract_evaluator.py#L334) and also needed to add this to the test as we are not mocking this function

PR #93, opened 2021-02-17
updated sklearn imports in /components

LucFrachon:
The sklearn imports were not working with sklearn 0.24. I updated the failing imports in /components/baselines and /components/metrics, they should now work.


ravinkohli:
Thanks for this PR however, we have moved to v0.1.0 which has a completely different implementation. Therefore, we will not be looking into this PR anytime soon.

PR #88, opened 2021-02-09
Dev shuhei

nabenabe0928:
[New]
- resampling_strategy, resampling_strategy_args -> splitting_type, splitting_params
- put all the splitting functions in resampling_strategy.py in a class as static methods
- refactoring of base_dataset.py
- Removed DEFAULT_RESAMPLING_PARAMETER and added NamedTuple

PR #94, opened 2021-02-17
FIX weighted loss issue

### LINKED ISSUE #92: Weights are never passed for cross entropy in binary classification
Currently, to assign weights to the loss function, we are checking the output type, and if its binary we are making a kwargs dict with 'pos_weight': weights. This assumes that for binary classification the loss function is BCEWithLogitsLoss. However, that is not the case as by default we are using CrossEntropyLoss for binary classification and therefore the weights are never passed to this loss function.

ravinkohli:
Changed tests for losses and how weighted strategy is handled in the base trainer

Fixes #92

### DIFF
--- a/autoPyTorch/pipeline/components/training/losses.py
+++ b/autoPyTorch/pipeline/components/training/losses.py
@@ -11,16 +11,17 @@ from torch.nn.modules.loss import _Loss as Loss
 from autoPyTorch.constants import BINARY, CLASSIFICATION_TASKS, CONTINUOUS, MULTICLASS, REGRESSION_TASKS, \
     STRING_TO_OUTPUT_TYPES, STRING_TO_TASK_TYPES, TASK_TYPES_TO_STRING
 
+
 losses = dict(classification=dict(
     CrossEntropyLoss=dict(
-        module=CrossEntropyLoss, supported_output_type=MULTICLASS),
+        module=CrossEntropyLoss, supported_output_types=(MULTICLASS, BINARY)),
     BCEWithLogitsLoss=dict(
-        module=BCEWithLogitsLoss, supported_output_type=BINARY)),
+        module=BCEWithLogitsLoss, supported_output_types=(BINARY,))),
     regression=dict(
         MSELoss=dict(
-            module=MSELoss, supported_output_type=CONTINUOUS),
+            module=MSELoss, supported_output_types=(CONTINUOUS,)),
         L1Loss=dict(
-            module=L1Loss, supported_output_type=CONTINUOUS)))
+            module=L1Loss, supported_output_types=(CONTINUOUS,))))
 
 default_losses = dict(classification=CrossEntropyLoss, regression=MSELoss)
 
@@ -38,16 +39,16 @@ def get_supported_losses(task: int, output_type: int) -> Dict[str, Type[Loss]]:
     supported_losses = dict()
     if task in CLASSIFICATION_TASKS:
         for key, value in losses['classification'].items():
-            if output_type == value['supported_output_type']:
+            if output_type in value['supported_output_types']:
                 supported_losses[key] = value['module']
     elif task in REGRESSION_TASKS:
         for key, value in losses['regression'].items():
-            if output_type == value['supported_output_type']:
+            if output_type in value['supported_output_types']:
                 supported_losses[key] = value['module']
     return supported_losses
 
 
-def get_loss_instance(dataset_properties: Dict[str, Any], name: Optional[str] = None) -> Type[Loss]:
+def get_loss(dataset_properties: Dict[str, Any], name: Optional[str] = None) -> Type[Loss]:
     assert 'task_type' in dataset_properties, \
         "Expected dataset_properties to have task_type got {}".format(dataset_properties.keys())
     assert 'output_type' in dataset_properties, \
--- a/test/test_pipeline/test_losses.py
+++ b/test/test_pipeline/test_losses.py
@@ -3,8 +3,7 @@ import pytest
 import torch
 from torch import nn
 
-from autoPyTorch.constants import STRING_TO_OUTPUT_TYPES
-from autoPyTorch.pipeline.components.training.losses import get_loss_instance
+from autoPyTorch.pipeline.components.training.losses import get_loss
 from autoPyTorch.utils.implementations import get_loss_weight_strategy
 
 
@@ -14,7 +13,7 @@ from autoPyTorch.utils.implementations import get_loss_weight_strategy
                                          'continuous'])
 def test_get_no_name(output_type):
     dataset_properties = {'task_type': 'tabular_classification', 'output_type': output_type}
-    loss = get_loss_instance(dataset_properties)
+    loss = get_loss(dataset_properties)
     assert isinstance(loss(), nn.Module)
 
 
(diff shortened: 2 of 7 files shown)


autoPyTorch/pipeline/components/training/losses.py
```diff
@@ -11,16 +11,17 @@
...
     STRING_TO_OUTPUT_TYPES, STRING_TO_TASK_TYPES, TASK_TYPES_TO_STRING
 
+
 losses = dict(classification=dict(
     CrossEntropyLoss=dict(
-        module=CrossEntropyLoss, supported_output_type=MULTICLASS),
+        module=CrossEntropyLoss, supported_output_types=(MULTICLASS, BINARY)),
     BCEWithLogitsLoss=dict(
-        module=BCEWithLogitsLoss, supported_output_type=BINARY)),
+        module=BCEWithLogitsLoss, supported_output_types=(BINARY,))),
```
franchuterivera, line 19: Hey, can you add a check in so that we make sure this is a list? Or can you instead of a tuple actually make a list? It is very easy to miss the comma...  Like this can be a purest making sure everything is a list or even an in code check

test/test_pipeline/test_losses.py
```diff
@@ -32,29 +31,24 @@ def test_get_name_error():
...
+    weights = None
+    if bool(weighted) and 'classification' in dataset_properties['task_type']:
+        strategy = get_loss_weight_strategy(loss)
+        weights = strategy(y=labels)
+        weights = torch.from_numpy(weights)
+        weights = weights.type(torch.FloatTensor)
+        kwargs = {'pos_weight': weights} if loss.__name__ == 'BCEWithLogitsLoss' else {'weight': weights}
+    loss = loss() if weights is None else loss(**kwargs)
+    score = loss(predictions, targets)
+    assert isinstance(score, torch.Tensor)
```
franchuterivera, line 54: Just to make sure nothing is broken, can you check the size of this score and you actually are providing a fixed target... I think the value in this case is deterministic?

franchuterivera (changes requested):
Hello, everything looks great, just some question to see if it is posible to improve the testing

autoPyTorch/pipeline/components/training/losses.py
ravinkohli: yeah I'll make them lists, and add a check for that

test/test_pipeline/test_losses.py
ravinkohli: in this case no it is not deterministic, I didn't add the seeding. But do you think it will be better to add the seed for all np, torch at the top of the contest file?

ravinkohli: I have updated the test_losses in a way that we don't need to set the seed for this test as per our discussion today

autoPyTorch/pipeline/components/training/losses.py
```diff
@@ -11,16 +11,17 @@
 from autoPyTorch.constants import BINARY, CLASSIFICATION_TASKS, CONTINUOUS, MULTICLASS, REGRESSION_TASKS, \
     STRING_TO_OUTPUT_TYPES, STRING_TO_TASK_TYPES, TASK_TYPES_TO_STRING
 
+
 losses = dict(classification=dict(
     CrossEntropyLoss=dict(
-        module=CrossEntropyLoss, supported_output_type=MULTICLASS),
```
nabenabe0928, line 16: Can you change those names to "MultiClass" and "Binary"? Plus, it would be better to define typing variable in the file that defines MULTICLASS and BINARY. e.g. SupportedOutputTypes = Union[MultiClass, Binary]

autoPyTorch/pipeline/components/training/losses.py
```diff
@@ -11,16 +11,17 @@
...
+        module=CrossEntropyLoss, supported_output_types=[MULTICLASS, BINARY]),
     BCEWithLogitsLoss=dict(
-        module=BCEWithLogitsLoss, supported_output_type=BINARY)),
+        module=BCEWithLogitsLoss, supported_output_types=[BINARY])),
     regression=dict(
         MSELoss=dict(
-            module=MSELoss, supported_output_type=CONTINUOUS),
+            module=MSELoss, supported_output_types=[CONTINUOUS]),
```
nabenabe0928, line 22: Can you change these dicts to NamedTuple or this [BaseDict](https://github.com/automl/Auto-PyTorch/pull/88/commits/ce1f63851c3fb1cc26db2a356d8990ada2045b28)?
By explicitly define the elements of dicts, the readability will increase and we can avoid typo.
The greatest benefit from those data structures is your working environment can suggest the keys.

autoPyTorch/pipeline/components/training/losses.py
ravinkohli: okay, should I change the supported output types also or you'll include it in your PR as well?

autoPyTorch/pipeline/components/training/losses.py
ravinkohli: I can change it later, as BaseDict is included in your PR. is it fine?

nabenabe0928 (commented):
Ok, let's do that so:)

autoPyTorch/pipeline/components/training/losses.py
nabenabe0928: I think it would be better to do it by yourself because you are more familiar with the related codes

autoPyTorch/pipeline/components/training/losses.py
```diff
@@ -11,16 +11,17 @@
...
         L1Loss=dict(
-            module=L1Loss, supported_output_type=CONTINUOUS)))
+            module=L1Loss, supported_output_types=[CONTINUOUS])))
```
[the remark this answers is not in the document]
ravinkohli: Thats true, but @franchuterivera in his older comment said he prefers them to be a list as people can miss the , that is needed to make a single element tuple i.e, (CONTINUOUS,).

autoPyTorch/pipeline/components/training/losses.py
ravinkohli: okay that makes sense

ravinkohli: I think these supported output types is a property of the particular loss. for example, if we had our own loss functions, this information will go to the get_properties function and hence for me it does not make sense to create supported output types as a typing variable.

autoPyTorch/pipeline/components/training/losses.py
nabenabe0928: But as mentioned, if we input something like supported_output_type=[MULTICLASS], the argument is gonna be updated by this.

autoPyTorch/pipeline/components/training/losses.py
nabenabe0928: > these supported output types are a property of the particular loss
Hum, that is true. But at the same time, we would not know the supported types explicitly from the codes, so let's make sure we can know the supported type explicitly (doc-string or class function such as CrossEntropyLoss.is_output_type_supported() etc...)

autoPyTorch/pipeline/components/training/losses.py
nabenabe0928: Yes, that is fine:)

autoPyTorch/pipeline/components/training/losses.py
ravinkohli: yes I'll add doc string for this file.

nabenabe0928 (changes requested):
Could you add the names of supported losses in the doc-string of get_supported_losses as well?

autoPyTorch/pipeline/components/training/losses.py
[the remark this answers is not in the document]
ravinkohli: I did not understand the two approaches.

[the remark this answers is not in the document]
nabenabe0928: Ah, maybe I misunderstood. Nothing and I will resolve the conversation.

nabenabe0928 (approved):

PR #102, opened 2021-02-19
First push for Mix/cut regularization

franchuterivera:
Enables cutoff and mix regularization recipes to the cocktail!


test/test_pipeline/components/test_training.py
```diff
@@ -226,79 +232,140 @@ def test_epoch_training(self):
...
+        selected_choice = config_dict.pop('__choice__', None)
+        for key, value in config_dict.items():
+            # Remove the selected_choice string from the parameter
+            # so we can query in the object for it
+            key = key.replace(selected_choice + ':', '')
+            assert key in vars(trainer_choice.choice)
+            assert value == trainer_choice.choice.__dict__[key]
+
+
+@pytest.mark.parametrize("cutmix_prob", [1.0, 0.0])
```
ravinkohli, line 314: are you certain this would generate a number between 0 and 1? I tried it multiple times and always got 1. Maybe it should be sample from 0, 1. Also, as in the implementation of mixup [here](https://github.com/franchuterivera/Auto-PyTorch/blob/a423cc98392dbe84cc22ede049ce226dcbac0825/autoPyTorch/pipeline/components/training/trainer/GridCutMixTrainer.py#L35), it may not always change the features of X depending on the r sampled which may cause the test to fail.

test/test_pipeline/components/test_training.py
```diff
@@ -226,79 +232,140 @@ def test_epoch_training(self):
...
+        assert isinstance(loss_func(criterion, y_pred).numpy().item(), int)
+    else:
+        # There has to be a change in the features
+        np.any(np.not_equal(X_new.numpy(), X.numpy()))
+        assert 0 < target_dict['lam'] < 1
+        # There has to be a mixup of loss function
+        # That's why the loss function returns a float
+
+
+@pytest.mark.parametrize("cutout_prob", [1.0, 0.0])
```
ravinkohli, line 344: same here

ravinkohli (changes requested):
The implementations look great. I have left 2 comments about the test which will require changes.

test/test_pipeline/components/test_training.py
franchuterivera: Actually, the if you see this [line](https://github.com/automl/Auto-PyTorch/blob/a423cc98392dbe84cc22ede049ce226dcbac0825/autoPyTorch/pipeline/components/training/trainer/RowCutOutTrainer.py#L32) we have a greater than operator. in the extreme cases this will return a value on `[0, 1)`. So if the fixture is zero, we have `0 > 0` which means that we do return the exact value of the criterion always and in the other case , when the fixture is 1, because is an open interval 1 is never returned so this test always return `1/fraction (criterion) + (1-1/fraction)`.

So the expectation is that when we have 0 as fixture we return an integer always (as the criterion is a count) or we return a float as we return a fraction of the criterion (which in this case is 0.5* 1 +0.5 * 1 = 1.0). The value is not the important thing but  the object type. Does this make sense? 

Or maybe I am missing something -- in this case, can you elaborate?

ravinkohli: okay, I see, then it's fine. but currently, the cutout probability fixture will only be 0 and 1 don't you wanna sample from 0, 1?

franchuterivera: I think the result from sampling is whether you don't do a mixup (probability 0) or you do a mixup (probability 1), so this cases should be covered.

My main motivation not to sample but pass the extreme cutoff probabilities is to make the testcase deterministic, while still testing the possibilities.

ravinkohli: okay that makes sense.

ravinkohli (approved):

PR #103, opened 2021-02-19
[WIP] Initial implementation of adversarial training, categorical batch normalization, categorical weight decay, categorical skip connection, categorical multibranch choice

ArlindKadra:
- [x] FGSM Adversarial Training
- [x] Activation controlled BN
- [x] Activation controlled WD
- [x] Activation controlled skip connections, the multi-branch choice is also a hyperparameter now. 
- [x] Add unit tests

### DIFF
--- a/autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
+++ b/autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
@@ -155,8 +155,8 @@ class ShapedResNetBackbone(ResNetBackbone):
         cs.add_condition(CS.EqualsCondition(mb_choice, use_sc, True))
         cs.add_condition(
             CS.AndConjunction(
-                CS.EqualsCondition(shake_drop_prob, mb_choice, 'shake-drop'),
-                CS.EqualsCondition(shake_drop_prob, use_sc, True)
+                CS.EqualsCondition(shake_drop_prob, use_sc, True),
+                CS.EqualsCondition(shake_drop_prob, mb_choice, "shake-drop"),
             )
         )
 
--- a/autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
+++ b/autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
@@ -154,8 +154,11 @@ class ShapedResNetBackbone(ResNetBackbone):
 
         use_dropout = CategoricalHyperparameter(
             "use_dropout", choices=use_dropout[0], default_value=use_dropout[1])
+        use_batch_normalization = CategoricalHyperparameter(
+            "use_batch_norm", choices=use_batch_norm[0], default_value=use_batch_norm[1])
         max_dropout = UniformFloatHyperparameter("max_dropout", lower=max_dropout[0][0], upper=max_dropout[0][1],
                                                  default_value=max_dropout[1])
+        cs.add_hyperparameters([use_batch_normalization])
         cs.add_hyperparameters([use_dropout])
         cs.add_hyperparameters([max_dropout])
         cs.add_condition(CS.EqualsCondition(max_dropout, use_dropout, True))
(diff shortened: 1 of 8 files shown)


autoPyTorch/pipeline/components/training/trainer/AdversarialTrainer.py
```diff
@@ -0,0 +1,157 @@
...
+            typing.Dict[str, np.ndarray]: arguments to the criterion function.
+        """
+        X_adversarial = self.fgsm_attack(X, y)
+        return (X, X_adversarial), {'y_a': y}
+
+    def criterion_preparation(self, y_a: np.ndarray, y_b: np.ndarray = None, lam: float = 1.0
+                              ) -> typing.Callable:
+        # Initial implementation, consider the adversarial loss and the normal network loss
+        # equally.
+        return lambda criterion, pred, adversarial_pred: 0.5 * criterion(pred, y_a) + \
```
ravinkohli, line 59: does it need to be fixed? maybe we can sample randomly or set lam: 0.5 in data_preparation

ArlindKadra: That is a really good point. I made it fixed since that is how it was in the paper and like that it cares equally about both. 
It makes sense for it to be a hyperparameter too though.

ravinkohli: okay then we can add a hyperparameter later, when we merge this into refactor_development

ArlindKadra: Yep, that was exactly what I was thinking.

nabenabe0928 (changes requested):
Comment for the function def `get_hyperparameter_search_space` in `autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py`.

Tuple[Tuple[bool], bool]?
I think if `use_batch_norm=False` then deactivate it and `use_batch_norm=True` then creating config space looks better.
In any case, there are only six possible choices for this.
```
((True, False), True)
((True, False), False)

no need to define config space for the candidates below, because it is a fixed parameter.
((True), True)
((True), False)
((False), True)
((False), False)
```

Based on this, we have four choices:
if we fix the parameter or not and the default parameter,
so should be `use_batch_norm` (default) should be bool and `search_for_batchnorm` (or any other name) should be another bool variable.

autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
```diff
@@ -153,8 +154,11 @@ def get_hyperparameter_search_space(dataset_properties: Optional[Dict] = None,
 
         use_dropout = CategoricalHyperparameter(
             "use_dropout", choices=use_dropout[0], default_value=use_dropout[1])
+        use_batch_normalization = CategoricalHyperparameter(
+            "use_batch_norm", choices=use_batch_norm[0], default_value=use_batch_norm[1])
```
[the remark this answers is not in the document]
ArlindKadra: I did not use the `use_batch_norm` name, because it shadows the variable from the method.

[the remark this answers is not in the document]
ArlindKadra: > Comment for the function def `get_hyperparameter_search_space` in `autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py`.
> 
> Tuple[Tuple[bool], bool]?
> I think if `use_batch_norm=False` then deactivate it and `use_batch_norm=True` then creating config space looks better.
> In any case, there are only six possible choices for this.
> 
> ```
> ((True, False), True)
> ((True, False), False)
> 
> no need to define config space for the candidates below, because it is a fixed parameter.
> ((True), True)
> ((True), False)
> ((False), True)
> ((False), False)
> ```
> 
> Based on this, we have four choices:
> if we fix the parameter or not and the default parameter,
> so should be `use_batch_norm` (default) should be bool and `search_for_batchnorm` (or any other name) should be another bool variable.

@nabenabe0928 I disagree, because this would be inconsistent with `use_dropout`. It can be improved in the future.

autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
```diff
@@ -82,6 +82,7 @@ def get_properties(dataset_properties: Optional[Dict[str, Any]] = None) -> Dict[
     @staticmethod
     def get_hyperparameter_search_space(dataset_properties: Optional[Dict] = None,  # type: ignore[override]
                                         num_groups: Tuple[Tuple, int] = ((1, 15), 5),
+                                        use_batch_norm: Tuple[Tuple, bool] = ((True, False), True),
```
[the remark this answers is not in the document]
ArlindKadra: Same reasoning for this one.

ravinkohli (approved):

PR #104, opened 2021-02-19
[WiP] Adds SWA, SE and LA optimizer

ravinkohli:
Adds Stochastic weight averaging, snapshot ensembles and look ahead optimizer

- [x] SWA
- [x] SE
- [x] Lookahead
- [x] Tests


autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
```diff
@@ -337,6 +342,15 @@ def _fit(self, X: Dict[str, Any], y: Any = None, **kwargs: Any) -> torch.nn.Modu
 
             torch.cuda.empty_cache()
 
+        if self.choice.use_swa:
+            # update batch norm statistics
+            swa_utils.update_bn(X['train_data_loader'], self.choice.swa_model)
+            # change model
+            update_model_state_dict_from_swa(X['network'], self.choice.swa_model.state_dict())
```
ravinkohli, line 349: Question: Do you think this implementation is fine? Or should I make an attribute 'swa_network' in base network which will directly predict from the averaged model?

ravinkohli: okay, then I'll leave it be. Thanks

autoPyTorch/pipeline/components/training/trainer/utils.py
```diff
@@ -0,0 +1,27 @@
+import re
+from typing import Dict
+
+import torch
+from torch.nn.parameter import Parameter
+
+
+def update_model_state_dict_from_swa(model: torch.nn.Module, swa_state_dict: Dict) -> None:
```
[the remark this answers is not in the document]
ravinkohli: yeah, it cant be done directly as SWA model adds this 'module:' key to all the parameters as well as introduces a parameter 'n_averaged'

autoPyTorch/pipeline/components/setup/network/base_network.py
```diff
@@ -26,6 +26,7 @@ def __init__(
     ) -> None:
         super(NetworkComponent, self).__init__()
         self.network = network
+        self.network_snapshots: List[torch.nn.Module] = []
```
franchuterivera, line 29: Can you add this to the constructor? The thing with scikit learn is that it can use joblib to launch the job in a different process, so when it rebuilds the fitted object it will lose this snapshot attribute, which will be troublesome for debug.

autoPyTorch/pipeline/components/setup/network/base_network.py
```diff
@@ -105,22 +107,35 @@ def predict(self, loader: torch.utils.data.DataLoader) -> torch.Tensor:
         """
         Performs batched prediction given a loader object
         """
-        assert self.network is not None
-        self.network.eval()
-
+        if len(self.network_snapshots) == 0:
+            assert self.network is not None
+            return self._predict(network=self.network.float(), loader=loader).cpu().numpy()
```
franchuterivera, line 112: Why are you casting the network to float?

autoPyTorch/pipeline/components/setup/network/base_network.py
```diff
@@ -105,22 +107,35 @@ def predict(self, loader: torch.utils.data.DataLoader) -> torch.Tensor:
...
+        else:
+            # if there are network snapshots,
+            # take average of predictions of all snapshots
+            Y_snapshot_preds = list()
+
+            for network in self.network_snapshots:
+                Y_snapshot_preds.append(self._predict(network.float(), loader))
+            Y_snapshot_preds = torch.stack(Y_snapshot_preds)
+            assert isinstance(Y_snapshot_preds, torch.Tensor)
```
franchuterivera, line 121: Doing is instance is painful (computationally), do `typing.cast()`

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -240,6 +281,24 @@ def on_epoch_end(self, X: Dict[str, Any], epoch: int) -> bool:
         If returns True, the training is stopped
 
         """
+        if X['is_cyclic_scheduler']:
+            if hasattr(self.scheduler, 'T_cur') and self.scheduler.T_cur == 0 and epoch != 1:
+                if self.use_swa:
+                    self.swa_model.update_parameters(self.model)
+                if self.use_se:
```
franchuterivera, line 292: Can you improve the name of use_se? 
In case there is a new person working, it will find it hard to understand what this is. Also, it is nice is you could also change the name `use_swa` to something more intuitive, it doesn't matter that the name is a bit longer

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
```diff
@@ -337,6 +343,15 @@ def _fit(self, X: Dict[str, Any], y: Any = None, **kwargs: Any) -> torch.nn.Modu
 
             torch.cuda.empty_cache()
 
+        if self.choice.use_swa:
+            # update batch norm statistics
+            swa_utils.update_bn(X['train_data_loader'], self.choice.swa_model.double())
```
franchuterivera, line 348: and why casting it here to double?

autoPyTorch/pipeline/components/training/trainer/utils.py
```diff
@@ -0,0 +1,165 @@
...
+class Lookahead(Optimizer):
+    r"""PyTorch implementation of the lookahead wrapper.
+    Lookahead Optimizer: https://arxiv.org/abs/1907.08610
+    """
+
+    def __init__(self, optimizer: Optimizer, config: Dict[str, Any]) -> None:
+        """optimizer: inner optimizer
+        la_steps (int): number of lookahead steps
+        la_alpha (float): linear interpolation factor. 1.0 recovers the inner optimizer.
+        pullback_momentum (str): change to inner optimizer momentum on interpolation update
```
franchuterivera, line 44: I think we no longer have this option and later below is hardcoded?

test/test_pipeline/test_tabular_classification.py
```diff
@@ -279,3 +282,29 @@ def test_set_range_search_space_updates(self, fit_dictionary):
...
             # head, units_layer does not exist in the configspace
             assert 'fully_connected:units_layer' in e.args[0]
+
+    def test_trainer_cocktails(self, fit_dictionary, mocker):  # noqa F811
+        fit_dictionary['epochs'] = 10
+        pipeline = TabularClassificationPipeline(
+            dataset_properties=fit_dictionary['dataset_properties'],
+            include={'lr_scheduler': ['CosineAnnealingWarmRestarts']})
+        cs = pipeline.get_hyperparameter_search_space()
+        config = cs.get_default_configuration()
```
franchuterivera, line 292: Can you add a check to make sure CosineAnnealingWarmRestarts is in the config below?

test/test_pipeline/test_tabular_classification.py
```diff
@@ -279,3 +282,29 @@ def test_set_range_search_space_updates(self, fit_dictionary):
             # As we are setting num_layers to 1 for fully connected
             # head, units_layer does not exist in the configspace
             assert 'fully_connected:units_layer' in e.args[0]
+
+    def test_trainer_cocktails(self, fit_dictionary, mocker):  # noqa F811
+        fit_dictionary['epochs'] = 10
+        pipeline = TabularClassificationPipeline(
+            dataset_properties=fit_dictionary['dataset_properties'],
+            include={'lr_scheduler': ['CosineAnnealingWarmRestarts']})
+        cs = pipeline.get_hyperparameter_search_space()
```
franchuterivera, line 291: Can you also update the config to have `use_se` and `use_swa` set to true?
something like:
```
config_dict = config.get_dictionary()
--> here set saw and see to true
config = COnfiguration(cs, values=config_dict)
```

autoPyTorch/pipeline/components/setup/network/base_network.py
ravinkohli: okay makes sense

autoPyTorch/pipeline/components/setup/network/base_network.py
ravinkohli: I tried that but it was still giving the same error.

test/test_pipeline/test_tabular_classification.py
```diff
@@ -279,3 +282,29 @@ def test_set_range_search_space_updates(self, fit_dictionary):
...
+        assert 'network_snapshots' in X and \
+               len(X['network_snapshots']) == config.get(f'trainer:{trainer}:se_lastk')
+
+        mocker.patch("autoPyTorch.pipeline.components.setup.network.base_network.NetworkComponent._predict",
+                     return_value=torch.Tensor([1]))
+        # Assert that predict gives no error when swa and se are on
+        assert isinstance(pipeline.predict(fit_dictionary['X_train']), np.ndarray)
+        # As SE is True, _predict should be called 3 times
+        assert pipeline.named_steps['network']._predict.call_count == 3
+
```
franchuterivera, line 309: Then let us check here that we have a desired amount of network snapshots.

Also, do you think we can have a check for `se_lastk`?

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
ravinkohli: okay I'll use the full names then. I also dont like this way of naming :P

test/test_pipeline/test_tabular_classification.py
```diff
@@ -279,3 +282,29 @@ def test_set_range_search_space_updates(self, fit_dictionary):
...
+        assert isinstance(pipeline.named_steps['trainer'].choice.optimizer, Lookahead)
```
franchuterivera, line 310: Can we check that the lookahead points to the correct parameters of the network? You know that the lookahead has the self.optimizer.parameters which should point to the network parameters. Also, can you check that `_la_step` makes sense in the training of the Look ahead?

franchuterivera (changes requested):
Hello, this looks good :)  -- I think we can improve the testing to address some corner case scenarios in my mind.

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
ravinkohli: I think the data in the loader is of double precision and hence I was getting an error because the model was in float

autoPyTorch/pipeline/components/setup/network/base_network.py
ravinkohli: as I converted the network to double earlier, for predict I am changing It back to float

autoPyTorch/pipeline/components/training/trainer/utils.py
ravinkohli: i'll remove it from the doc

test/test_pipeline/test_tabular_classification.py
ravinkohli: sure

test/test_pipeline/test_tabular_classification.py
ravinkohli: sure

test/test_pipeline/test_tabular_classification.py
ravinkohli: I am doing this [here](https://github.com/automl/Auto-PyTorch/blob/5a7e0f67f20e67a8928e11455e34f518dcde2edd/test/test_pipeline/test_tabular_classification.py#L300)

test/test_pipeline/test_tabular_classification.py
ravinkohli: what do you mean by makes sense in the training? Do you mean that for the given epochs if we use certain la_steps, the final value of _la_step should be a specific value?

PR #105, opened 2021-02-19
Refactoring base dataset

nabenabe0928:
No change in the inputs and functionalities, inputs, returns.


autoPyTorch/datasets/base_dataset.py
```diff
@@ -172,8 +162,8 @@ def update_transform(self, transform: Optional[torchvision.transforms.Compose],
     def __getitem__(self, index: int, train: bool = True) -> Tuple[np.ndarray, ...]:
         """
         The base dataset uses a Subset of the data. Nevertheless, the base dataset expect
-        both validation and test data to be present in the same dataset, which motivated the
-        need to dynamically give train/test data with the __getitem__ command.
+        both validation and test data to be present in the same dataset, which is motivated
```
ravinkohli, line 165: the need for dynamically giving train or test data arises because the BaseDataset has both train and val data. So I think the correct way to explain would be: the base dataset expects both validation and test data to be present in the same dataset, which motivates the need to dynamically ...

autoPyTorch/pipeline/components/training/trainer/StandardTrainer.py
```diff
@@ -56,6 +56,8 @@ def get_properties(dataset_properties: typing.Optional[typing.Dict[str, typing.A
     def get_hyperparameter_search_space(dataset_properties: typing.Optional[typing.Dict] = None,
                                         weighted_loss: typing.Tuple[typing.Tuple, bool] = ((True, False), True)
                                         ) -> ConfigurationSpace:
+        weighted_loss = CategoricalHyperparameter("weighted_loss", choices=weighted_loss[0],
```
ravinkohli, line 59: I dont think there is a need to have this line as we are already making the hyperparameter inside the if statement, moreover, on the branch [refactor_development_regularization_cocktails](https://github.com/automl/Auto-PyTorch/tree/refactor_development_regularization_cocktails) we are dramatically updating the code for trainers. So, I'll suggest you to ignore this module for now

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
```diff
@@ -98,11 +98,11 @@ def get_components(self) -> Dict[str, autoPyTorchComponent]:
         return components
 
     def get_hyperparameter_search_space(
-        self,
-        dataset_properties: Optional[Dict[str, str]] = None,
-        default: Optional[str] = None,
-        include: Optional[List[str]] = None,
-        exclude: Optional[List[str]] = None,
+            self,
```
ravinkohli, line 101: this is over indented, through the project we are following a 4 space tab for continuation indents

test/test_pipeline/components/test_tabular_column_transformer.py
```diff
@@ -12,6 +12,11 @@
     TabularColumnTransformer
 )
 
+@pytest.mark.parametrize("fit_dictionary", ['fit_dictionary_numerical_only',
```
ravinkohli, line 15: I dont think this change is correct.

.gitignore
```diff
@@ -138,3 +138,7 @@ dask-worker-space/
 # Test output
 tmp/
 .tmp_evaluation
+
+# private
```
ravinkohli, line 142: I think we dont want to have ignore files specific to a collaborator, so you could either use git to ignore changes made to these files like [this](https://git-scm.com/docs/git-update-index), or be careful when adding files to staging area

ravinkohli (changes requested):
The changes to the Base dataset largely seem okay to me, I have made some comments that will require changes

test/test_pipeline/components/test_tabular_column_transformer.py
nabenabe0928: I did not change, but it was merged from the latest change (so I will get it back to the original)

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
nabenabe0928: It also happened somehow, but I will get them back, thanks.

autoPyTorch/pipeline/components/training/trainer/StandardTrainer.py
nabenabe0928: Maybe I should learn what happened from the merge. This is also not my change.

autoPyTorch/datasets/base_dataset.py
nabenabe0928: Ok, I will adapt to your comment.

.gitignore
nabenabe0928: Thanks, I will do that

autoPyTorch/datasets/base_dataset.py
```diff
@@ -23,25 +23,26 @@
...
     if val_tensors is not None:
-        for i in range(len(val_tensors)):
-            check_valid_data(val_tensors[i])
+        for val_tensor in val_tensors:
+            check_valid_data(val_tensor)
 
 
 class TransformSubset(Subset):
     """
+    I did not understand the meaning of this doc-string.
```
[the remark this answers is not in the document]
ravinkohli: Could you raise an issue for this or better, ask it on chat? I think its not ideal to have such remarks in the code.

ravinkohli (approved):

PR #106, opened 2021-02-19
Refactoring base dataset splitting functions

nabenabe0928:
This PR fork from the PR #105.
- added CrossValFuncs class and HoldOutFuncs class to group the functions.
- added is_stratified() to the CrossValTypes and HoldOutValTypes class.
- added SplittingFunc type instead of CROSS_VAL_FN

### DIFF
--- a/autoPyTorch/datasets/base_dataset.py
+++ b/autoPyTorch/datasets/base_dataset.py
@@ -1,6 +1,6 @@
 from abc import ABCMeta
-from typing import Any, Dict, List, Optional, Sequence, Tuple, Union, cast
 import os
+from typing import Any, Dict, List, Optional, Sequence, Tuple, Union, cast
 import uuid
 
 import numpy as np
@@ -124,7 +124,7 @@ class BaseDataset(Dataset, metaclass=ABCMeta):
         self.resampling_strategy_args = resampling_strategy_args
         self.task_type: Optional[str] = None
         self.issparse: bool = issparse(self.train_tensors[0])
-        self.input_shape: Tuple[int] = self.train_tensors[0].shape[1:] 
+        self.input_shape: Tuple[int] = self.train_tensors[0].shape[1:]
 
         if len(self.train_tensors) == 2 and self.train_tensors[1] is not None:
             self.output_type: str = type_of_target(self.train_tensors[1])
--- a/autoPyTorch/datasets/base_dataset.py
+++ b/autoPyTorch/datasets/base_dataset.py
@@ -1,7 +1,7 @@
-from abc import ABCMeta
 import os
-from typing import Any, Dict, List, Optional, Sequence, Tuple, Union, cast
 import uuid
+from abc import ABCMeta
+from typing import Any, Dict, List, Optional, Sequence, Tuple, Union, cast
 
 import numpy as np
 
--- a/autoPyTorch/datasets/base_dataset.py
+++ b/autoPyTorch/datasets/base_dataset.py
@@ -1,5 +1,7 @@
 from abc import ABCMeta
 from typing import Any, Dict, List, Optional, Sequence, Tuple, Union, cast
+import os
+import uuid
 
 import numpy as np
 
@@ -13,18 +15,18 @@ import torchvision
 
 from autoPyTorch.constants import CLASSIFICATION_OUTPUTS, STRING_TO_OUTPUT_TYPES
 from autoPyTorch.datasets.resampling_strategy import (
-    CROSS_VAL_FN,
+    CrossValFunc,
+    CrossValFuncs,
     CrossValTypes,
     DEFAULT_RESAMPLING_PARAMETERS,
-    HOLDOUT_FN,
-    HoldoutValTypes,
-    get_cross_validators,
-    get_holdout_validators,
-    is_stratified,
+    HoldOutFunc,
+    HoldOutFuncs,
+    HoldoutValTypes
 )
-from autoPyTorch.utils.common import FitRequirement, hash_array_or_matrix
+from autoPyTorch.utils.common import FitRequirement
 
-BaseDatasetType = Union[Tuple[np.ndarray, np.ndarray], Dataset]
+
+BaseDatasetInputType = Union[Tuple[np.ndarray, np.ndarray], Dataset]
 
 
 def check_valid_data(data: Any) -> None:
@@ -33,7 +35,8 @@ def check_valid_data(data: Any) -> None:
             'The specified Data for Dataset must have both __getitem__ and __len__ attribute.')
 
 
-def type_check(train_tensors: BaseDatasetType, val_tensors: Optional[BaseDatasetType] = None) -> None:
+def type_check(train_tensors: BaseDatasetInputType,
+               val_tensors: Optional[BaseDatasetInputType] = None) -> None:
     """To avoid unexpected behavior, we use loops over indices."""
     for i in range(len(train_tensors)):
         check_valid_data(train_tensors[i])
@@ -71,10 +74,10 @@ class TransformSubset(Subset):
 class BaseDataset(Dataset, metaclass=ABCMeta):
     def __init__(
         self,
-        train_tensors: BaseDatasetType,
+        train_tensors: BaseDatasetInputType,
         dataset_name: Optional[str] = None,
-        val_tensors: Optional[BaseDatasetType] = None,
-        test_tensors: Optional[BaseDatasetType] = None,
+        val_tensors: Optional[BaseDatasetInputType] = None,
+        test_tensors: Optional[BaseDatasetInputType] = None,
         resampling_strategy: Union[CrossValTypes, HoldoutValTypes] = HoldoutValTypes.holdout_validation,
         resampling_strategy_args: Optional[Dict[str, Any]] = None,
         shuffle: Optional[bool] = True,
@@ -106,21 +109,23 @@ class BaseDataset(Dataset, metaclass=ABCMeta):
             val_transforms (Optional[torchvision.transforms.Compose]):
                 Additional Transforms to be applied to the validation/test data
         """
-        self.dataset_name = dataset_name if dataset_name is not None \
-            else hash_array_or_matrix(train_tensors[0])
+        self.dataset_name = dataset_name
+
+        if self.dataset_name is not None:
+            self.dataset_name = str(uuid.uuid1(clock_seq=os.getpid()))
 
         if not hasattr(train_tensors[0], 'shape'):
             type_check(train_tensors, val_tensors)
         self.train_tensors, self.val_tensors, self.test_tensors = train_tensors, val_tensors, test_tensors
-        self.cross_validators: Dict[str, CROSS_VAL_FN] = {}
-        self.holdout_validators: Dict[str, HOLDOUT_FN] = {}
+        self.cross_validators: Dict[str, CrossValFunc] = {}
+        self.holdout_validators: Dict[str, HoldOutFunc] = {}
         self.rng = np.random.RandomState(seed=seed)
         self.shuffle = shuffle
         self.resampling_strategy = resampling_strategy
         self.resampling_strategy_args = resampling_strategy_args
         self.task_type: Optional[str] = None
         self.issparse: bool = issparse(self.train_tensors[0])
-        self.input_shape: Tuple[int] = self.train_tensors[0].shape[1:]
+        self.input_shape: Tuple[int] = self.train_tensors[0].shape[1:] 
 
         if len(self.train_tensors) == 2 and self.train_tensors[1] is not None:
             self.output_type: str = type_of_target(self.train_tensors[1])
@@ -134,8 +139,8 @@ class BaseDataset(Dataset, metaclass=ABCMeta):
         self.is_small_preprocess = True
 
         # Make sure cross validation splits are created once
-        self.cross_validators = get_cross_validators(*CrossValTypes)
-        self.holdout_validators = get_holdout_validators(*HoldoutValTypes)
+        self.cross_validators = CrossValFuncs.get_cross_validators(*CrossValTypes)
+        self.holdout_validators = HoldOutFuncs.get_holdout_validators(*HoldoutValTypes)
         self.splits = self.get_splits_from_resampling_strategy()
 
         # We also need to be able to transform the data, be it for pre-processing
@@ -263,7 +268,7 @@ class BaseDataset(Dataset, metaclass=ABCMeta):
         if not isinstance(cross_val_type, CrossValTypes):
             raise NotImplementedError(f'The selected `cross_val_type` "{cross_val_type}" is not implemented.')
         kwargs = {}
-        if is_stratified(cross_val_type):
+        if cross_val_type.is_stratified():
             # we need additional information about the data for stratification
             kwargs["stratify"] = self.train_tensors[-1]
         splits = self.cross_validators[cross_val_type.name](
@@ -298,7 +303,7 @@ class BaseDataset(Dataset, metaclass=ABCMeta):
         if not isinstance(holdout_val_type, HoldoutValTypes):
             raise NotImplementedError(f'The specified `holdout_val_type` "{holdout_val_type}" is not supported.')
         kwargs = {}
-        if is_stratified(holdout_val_type):
+        if holdout_val_type.is_stratified():
             # we need additional information about the data for stratification
             kwargs["stratify"] = self.train_tensors[-1]
         train, val = self.holdout_validators[holdout_val_type.name](val_share, self._get_indices(), **kwargs)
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -138,3 +138,7 @@ dask-worker-space/
 # Test output
 tmp/
 .tmp_evaluation
+
+# private
```
ravinkohli, line 142: yeah I think you should remove these from here

autoPyTorch/datasets/base_dataset.py
```diff
@@ -13,36 +13,36 @@
...
-    get_cross_validators,
-    get_holdout_validators,
-    is_stratified,
+    HoldOutFuncs,
+    HoldOutFunc
 )
 from autoPyTorch.utils.common import FitRequirement, hash_array_or_matrix
 
-BASE_DATASET_INPUT = Union[Tuple[np.ndarray, np.ndarray], Dataset]
+BaseDatasetType = Union[Tuple[np.ndarray, np.ndarray], Dataset]
```
ravinkohli, line 26: maybe we can have this called BaseDatasetInputType

autoPyTorch/datasets/base_dataset.py
```diff
@@ -103,8 +101,8 @@ def __init__(
         if not hasattr(train_tensors[0], 'shape'):
             type_check(train_tensors, val_tensors)
         self.train_tensors, self.val_tensors, self.test_tensors = train_tensors, val_tensors, test_tensors
-        self.cross_validators: Dict[str, CROSS_VAL_FN] = {}
-        self.holdout_validators: Dict[str, HOLDOUT_FN] = {}
```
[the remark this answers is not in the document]
ravinkohli: would you make another commit to this PR?

autoPyTorch/datasets/base_dataset.py
```diff
@@ -186,34 +175,24 @@ def __getitem__(self, index: int, train: bool = True) -> Tuple[np.ndarray, ...]:
             A transformed single point prediction
         """
 
-        if hasattr(self.train_tensors[0], 'loc'):
-            X = self.train_tensors[0].iloc[[index]]
-        else:
-            X = self.train_tensors[0][index]
+        X = self.train_tensors[0].iloc[[index]] if hasattr(self.train_tensors[0], 'loc') \
```
ravinkohli, line 178: is there a reason why you are changing these if else to be in a single line?

autoPyTorch/datasets/resampling_strategy.py
```diff
@@ -67,87 +100,86 @@ class HoldoutValTypes(IntEnum):
...
+        splits = list(cv.split(indices, kwargs["stratify"]))
+        return splits
+
+    @staticmethod
+    def k_fold_cross_validation(num_splits: int, indices: np.ndarray, **kwargs: Any) \
+            -> List[Tuple[np.ndarray, np.ndarray]]:
+        """
+        Standard k fold cross validation.
+
+        :param indices: array of indices to be split
```
ravinkohli, line 153: We are following google's doc string format. For more info, check 3.8 of [this](https://google.github.io/styleguide/pyguide.html)

autoPyTorch/datasets/resampling_strategy.py
```diff
@@ -67,87 +100,86 @@ class HoldoutValTypes(IntEnum):
...
+            -> List[Tuple[np.ndarray, np.ndarray]]:
+        """
+        Returns train and validation indices respecting the temporal ordering of the data.
+        Dummy example: [0, 1, 2, 3] with 3 folds yields
+            [0] [1]
+            [0, 1] [2]
+            [0, 1, 2] [3]
+
+        :param indices: array of indices to be split
+        :param num_splits: number of cross validation splits
```
ravinkohli, line 172: please adjust them here as well

autoPyTorch/pipeline/components/training/trainer/StandardTrainer.py
```diff
@@ -56,6 +56,8 @@ def get_properties(dataset_properties: typing.Optional[typing.Dict[str, typing.A
     def get_hyperparameter_search_space(dataset_properties: typing.Optional[typing.Dict] = None,
                                         weighted_loss: typing.Tuple[typing.Tuple, bool] = ((True, False), True)
                                         ) -> ConfigurationSpace:
+        weighted_loss = CategoricalHyperparameter("weighted_loss", choices=weighted_loss[0],
```
ravinkohli, line 59: again, I dont think you want to do this

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
```diff
@@ -98,11 +98,11 @@ def get_components(self) -> Dict[str, autoPyTorchComponent]:
         return components
 
     def get_hyperparameter_search_space(
-        self,
-        dataset_properties: Optional[Dict[str, str]] = None,
-        default: Optional[str] = None,
-        include: Optional[List[str]] = None,
-        exclude: Optional[List[str]] = None,
+            self,
+            dataset_properties: Optional[Dict[str, str]] = None,
```
ravinkohli, line 102: I think these are not your changes, could you look into them?

test/test_pipeline/components/test_tabular_column_transformer.py
```diff
@@ -12,6 +12,11 @@
     TabularColumnTransformer
 )
 
+@pytest.mark.parametrize("fit_dictionary", ['fit_dictionary_numerical_only',
+                                            'fit_dictionary_categorical_only',
+                                            'fit_dictionary_num_and_categorical'], indirect=True)
+class TestTabularTransformer:
```
ravinkohli, line 18: again here, I dont think these are your changes

ravinkohli (changes requested):
The changes look good. I have added some comments that need to be addressed

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
nabenabe0928: Ok, I will check all the strange changes later (tomorrow). I think this came from the merge to older `refactor_development` branch. Once I updated the branch, I haven't got the changes.

autoPyTorch/datasets/resampling_strategy.py
nabenabe0928: Maybe it's better to tell francisco (based on `git blame`, they are from chico.). My template of the doc-string is google's doc, so when I put new stuff, they're always google's doc. But i will change them this time.

autoPyTorch/datasets/base_dataset.py
nabenabe0928: Ah, ok good to hear. I will change it to this name then.

autoPyTorch/datasets/base_dataset.py
nabenabe0928: That is just because it is shorter and I use inline if statement whenever I can shorten lines. I use them. In other words, this is my convention. For me, It does not make sense to make lines longer unless the error statements become complicated.

ravinkohli: okay

autoPyTorch/datasets/base_dataset.py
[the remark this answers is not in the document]
nabenabe0928: Sorry, this is the comment for the older commit. It has been already solved in this version

ravinkohli (changes requested):
hey, could you rebase? I think after #105 there are some related changes

nabenabe0928:
> hey, could you rebase? I think after #105 there are some related changes
According to the following command and the output, it looks ok.

```
(base) shuhei@shuhei-UX430UNR:~/research/Auto-PyTorch$ git pull --rebase upstream refactor_development 
From github.com:automl/Auto-PyTorch
 * branch            refactor_development -> FETCH_HEAD
Current branch refactoring-base-dataset_splitting-functions is up to date.
```

ravinkohli:
> > hey, could you rebase? I think after #105 there are some related changes
> 
> According to the following command and the output, it looks ok.
> 
> ```
> (base) shuhei@shuhei-UX430UNR:~/research/Auto-PyTorch$ git pull --rebase upstream refactor_development 
> From github.com:automl/Auto-PyTorch
>  * branch            refactor_development -> FETCH_HEAD
> Current branch refactoring-base-dataset_splitting-functions is up to date.
> ```

Hey, okay then, this can be merged, has @franchuterivera seen it?

ravinkohli (approved):

autoPyTorch/datasets/base_dataset.py
```diff
@@ -13,36 +13,36 @@
...
+    for train_tensor in train_tensors:
+        check_valid_data(train_tensor)
     if val_tensors is not None:
         for i in range(len(val_tensors)):
             check_valid_data(val_tensors[i])
 
 
 class TransformSubset(Subset):
     """
+    I did not understand the meaning of this doc-string.
```
franchuterivera, line 45: Let us change the statement here. Maybe something like this (please change as you see fit)?

```
This class extends the PyTorchs' Subset class with the ability to select among train and test data. 
Fundamentally, the main dataset class (BaseDataset)  requires selecting which transformations are to be applied to the data, when accessing an item within the dataset.
 The TransformSubset allows an extra boolean while indexing the main dataset towards this goal.
```

autoPyTorch/datasets/base_dataset.py
```diff
@@ -97,18 +97,15 @@ def __init__(
             val_transforms (Optional[torchvision.transforms.Compose]):
                 Additional Transforms to be applied to the validation/test data
         """
-        if dataset_name is not None:
-            self.dataset_name = dataset_name
-        else:
-            self.dataset_name = hash_array_or_matrix(train_tensors[0])
+        self.dataset_name = dataset_name if dataset_name is not None \
```
franchuterivera, line 100: I think is better to prevent the use of `\` when possible. 

Also, this `hash_array_or_matrix` function is something we do not user anymore... maybe try using [this](https://github.com/automl/Auto-PyTorch/blob/b3c301c66c230a5a4040ca1fc1a20858d2e2d9dc/autoPyTorch/api/base_task.py#L914).

autoPyTorch/datasets/base_dataset.py
franchuterivera: I think it makes sense, but only if you do not have a backslash. There is even a PEP rule for that which we should probably enable...

autoPyTorch/datasets/base_dataset.py
```diff
@@ -13,36 +13,36 @@
 
 from autoPyTorch.constants import CLASSIFICATION_OUTPUTS, STRING_TO_OUTPUT_TYPES
 from autoPyTorch.datasets.resampling_strategy import (
-    CROSS_VAL_FN,
+    CrossValFunc,
+    CrossValFuncs,
```
franchuterivera, line 17: Do you think we can do something to improve the code so that we do not have 2 elements that only differ with an 's'?

What I mean is that, is there something we can further do to get rid of either `CrossValFunc` or `CrossValFuncs`?

autoPyTorch/datasets/resampling_strategy.py
```diff
@@ -25,26 +29,59 @@ def __call__(self,
...
 
+    def is_stratified(self) -> bool:
+        stratified = [self.stratified_holdout_validation]
+        return getattr(self, self.name) in stratified
+
 
+"""TODO: deprecate soon"""
 RESAMPLING_STRATEGIES = [CrossValTypes, HoldoutValTypes]
 
+"""TODO: deprecate soon"""
```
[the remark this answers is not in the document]
franchuterivera: I think that when possible, it is better to stick with python native solutions unless there is a strong need to do so...

Now, we do not know how long it will take for this new PR to be added, so can you try not adding `deprecate soon`? It gives the intuition that something is broken, which I do not think is the case right now.

autoPyTorch/datasets/resampling_strategy.py
```diff
@@ -67,87 +100,96 @@ class HoldoutValTypes(IntEnum):
...
-    splits = list(cv.split(indices))
-    return splits
+class HoldOutFuncs():
+    @staticmethod
+    def holdout_validation(val_share: float, indices: np.ndarray, **kwargs: Any) -> Tuple[np.ndarray, np.ndarray]:
+        train, val = train_test_split(indices, test_size=val_share, shuffle=False)
+        return train, val
+
+    @staticmethod
+    def stratified_holdout_validation(val_share: float, indices: np.ndarray, **kwargs: Any) \
```
franchuterivera, line 110: Sorry for the trouble, but since you are modifying this, can you remove the backslash here?

franchuterivera (changes requested):
Hello, thanks for the changes, this adds further documentation and clarity to the resampling part of the code. Can you take a look into the comments I left to further make it more clear?

autoPyTorch/datasets/base_dataset.py
nabenabe0928: I changed the doc-string in the `refactor_development` [here](https://github.com/automl/Auto-PyTorch/blob/refactor_development/autoPyTorch/datasets/base_dataset.py#L45).
Could you compare and make it better?

autoPyTorch/datasets/base_dataset.py
nabenabe0928: Ok, so 
`self.dataset_name = str(uuid.uuid1(clock_seq=os.getpid()))`?

autoPyTorch/datasets/base_dataset.py
nabenabe0928: Yeah, I did it in the successive PR (#108)!

autoPyTorch/datasets/resampling_strategy.py
nabenabe0928: What kind of change do you want?
Could you write it down here?

autoPyTorch/datasets/base_dataset.py
```diff
@@ -106,14 +108,16 @@ def __init__(
             val_transforms (Optional[torchvision.transforms.Compose]):
                 Additional Transforms to be applied to the validation/test data
         """
-        self.dataset_name = dataset_name if dataset_name is not None \
-            else hash_array_or_matrix(train_tensors[0])
+        self.dataset_name = ""
+
+        if self.dataset_name is not None:
```
franchuterivera, line 113: I think self.dataset_name is always going to be None due to line 111?

Can we make dataset_name not Optional to simplify the code?

franchuterivera (commented):
thanks for the code, the changes look more readable now. Do you think we can add unit testing to make sure we don't have any misses? For example, add unit testing for resampling strategy so that we make sure that the functions are working as expected.

autoPyTorch/datasets/base_dataset.py
```diff
@@ -71,10 +73,10 @@ def __getitem__(self, idx: int) -> np.ndarray:
 class BaseDataset(Dataset, metaclass=ABCMeta):
     def __init__(
         self,
-        train_tensors: BaseDatasetType,
-        dataset_name: Optional[str] = None,
-        val_tensors: Optional[BaseDatasetType] = None,
-        test_tensors: Optional[BaseDatasetType] = None,
+        train_tensors: BaseDatasetInputType,
+        dataset_name: str = "",
```
franchuterivera, line 77: Sorry for the trouble, what I mean is to have:
```
dataset_name: str,
```
So that the dataset is a required name? If you think it should not be a required name then `Optional[str] =None`  makes sense. For looking at the code, the main code should provide it?

autoPyTorch/datasets/base_dataset.py
```diff
@@ -106,14 +108,16 @@ def __init__(
             val_transforms (Optional[torchvision.transforms.Compose]):
                 Additional Transforms to be applied to the validation/test data
         """
-        self.dataset_name = dataset_name if dataset_name is not None \
-            else hash_array_or_matrix(train_tensors[0])
+        self.dataset_name = dataset_name
+
+        if self.dataset_name == "":
+            self.dataset_name = str(uuid.uuid1(clock_seq=os.getpid()))
```
franchuterivera, line 114: Then if you think that it should not be required, then maybe do it as:
```
self.dataset_name = dataset_name if dataset_name is not None else str(
    uuid.uuid1(clock_seq=os.getpid())
)
```

What do you think?

autoPyTorch/datasets/tabular_dataset.py
```diff
@@ -73,7 +73,7 @@ def __init__(self,
                  seed: Optional[int] = 42,
                  train_transforms: Optional[torchvision.transforms.Compose] = None,
                  val_transforms: Optional[torchvision.transforms.Compose] = None,
-                 dataset_name: Optional[str] = None,
```
franchuterivera, line 76: Maybe `None` here is ok. My concern is that we where defining the name both in the dataset and in [here](https://github.com/automl/Auto-PyTorch/blob/988a3a23a356fd9492331792f25cf37935a921eb/autoPyTorch/datasets/base_dataset.py#L114)

Which one do you think makes more sense?

autoPyTorch/datasets/base_dataset.py
nabenabe0928: Yeah, actually I am also thinking about it this way.
Probably, it will be better.

nabenabe0928: Ah, wait. But when I think about the case where we would like to use totally new datasets which do not have any name, probably we would like to choose our own name. In this sense, it is better to get back to `Optional`.

Issue #97, opened 2021-02-18
Create calculate_loss support


franchuterivera:
We want to make sure we have a minimization problem in all of our optimization contexts. We should follow autosklearn, in particular [here](https://github.com/automl/auto-sklearn/blob/598d33d0bc3438cdf3e7c1952f93afb6999943a0/autosklearn/metrics/__init__.py#L416).

This has to apply to individual models loss calculation (train evaluator, test evaluator) but also in the ensemble builder. Also calculate_score has to be update to return the real function value, not the result of the scorer, which is always a maximization problem.

PR #113, opened 2021-02-23
move to a minimization problem

franchuterivera:
Addresses #97

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -29,7 +29,7 @@ jobs:
     - name: Run tests
       run: |
         if [ ${{ matrix.code-cov }} ]; then codecov='--cov=autoPyTorch --cov-report=xml'; fi
-        python -m pytest --durations=20 --timeout=300 --timeout-method=thread -v $codecov test
+        python -m pytest --durations=20 --timeout=600 --timeout-method=signal -v $codecov test
     - name: Check for files left behind by test
       if: ${{ always() }}
       run: |
--- a/autoPyTorch/ensemble/ensemble_builder.py
+++ b/autoPyTorch/ensemble/ensemble_builder.py
@@ -1413,7 +1413,8 @@ class EnsembleBuilder(object):
 
     def _get_list_of_sorted_preds(self) -> List[Tuple[str, float, int]]:
         """
-            Returns a list of sorted predictions in descending order
+            Returns a list of sorted predictions in descending performance order.
+            (We are solving a minimization problem)
             Losses are taken from self.read_losses.
 
             Parameters
@@ -1422,7 +1423,10 @@ class EnsembleBuilder(object):
 
             Return
             ------
-            sorted_keys: list
+            sorted_keys:
+                given a sequence of pairs of (loss[i], num_run[i]) = (l[i], n[i]),
+                we will sort s.t. l[0] <= l[1] <= ... <= l[N] and for any pairs of
+                i, j (i < j, l[i] = l[j]), the resulting sequence satisfies n[i] <= n[j]
         """
         # Sort by loss - smaller is better!
         sorted_keys = list(sorted(
--- a/autoPyTorch/ensemble/ensemble_builder.py
+++ b/autoPyTorch/ensemble/ensemble_builder.py
@@ -33,7 +33,7 @@ from autoPyTorch.constants import BINARY
 from autoPyTorch.ensemble.abstract_ensemble import AbstractEnsemble
 from autoPyTorch.ensemble.ensemble_selection import EnsembleSelection
 from autoPyTorch.pipeline.components.training.metrics.base import autoPyTorchMetric
-from autoPyTorch.pipeline.components.training.metrics.utils import calculate_score
+from autoPyTorch.pipeline.components.training.metrics.utils import calculate_loss, calculate_score
 from autoPyTorch.utils.backend import Backend
 from autoPyTorch.utils.logging_ import get_named_client_logger
 
@@ -512,13 +512,13 @@ class EnsembleBuilder(object):
         # objects as attributes. For this reason, we dump to disk the stage of the past
         # ensemble iterations to kick-start the ensembling process
         # {"file name": {
-        #    "ens_score": float
+        #    "ens_loss": float
         #    "mtime_ens": str,
         #    "mtime_test": str,
         #    "seed": int,
         #    "num_run": int,
         # }}
-        self.read_scores = {}
+        self.read_losses = {}
         # {"file_name": {
         #    Y_ENSEMBLE: np.ndarray
         #    Y_TEST: np.ndarray
@@ -528,7 +528,7 @@ class EnsembleBuilder(object):
 
         # Depending on the dataset dimensions,
         # regenerating every iteration, the predictions
-        # scores for self.read_preds
+        # losses for self.read_preds
         # is too computationally expensive
         # As the ensemble builder is stateless
         # (every time the ensemble builder gets resources
@@ -551,17 +551,17 @@ class EnsembleBuilder(object):
                         traceback.format_exc(),
                     )
                 )
-        self.ensemble_score_file = os.path.join(
+        self.ensemble_loss_file = os.path.join(
             self.backend.internals_directory,
-            'ensemble_read_scores.pkl'
+            'ensemble_read_losses.pkl'
         )
-        if os.path.exists(self.ensemble_score_file):
+        if os.path.exists(self.ensemble_loss_file):
             try:
-                with (open(self.ensemble_score_file, "rb")) as memory:
-                    self.read_scores = pickle.load(memory)
+                with (open(self.ensemble_loss_file, "rb")) as memory:
+                    self.read_losses = pickle.load(memory)
             except Exception as e:
                 self.logger.warning(
-                    "Could not load the previous iterations of ensemble_builder scores."
+                    "Could not load the previous iterations of ensemble_builder losses."
                     "This might impact the quality of the run. Exception={} {}".format(
                         e,
                         traceback.format_exc(),
@@ -770,8 +770,8 @@ class EnsembleBuilder(object):
             time_left - used_time,
         )
 
-        # populates self.read_preds and self.read_scores
-        if not self.score_ensemble_preds():
+        # populates self.read_preds and self.read_losses
+        if not self.compute_loss_per_model():
             if return_predictions:
                 return self.ensemble_history, self.ensemble_nbest, train_pred, test_pred
             else:
@@ -817,9 +817,9 @@ class EnsembleBuilder(object):
         if self.max_resident_models is not None:
             self._delete_excess_models(selected_keys=candidate_models)
 
-        # Save the read scores status for the next iteration
-        with open(self.ensemble_score_file, "wb") as memory:
-            pickle.dump(self.read_scores, memory)
+        # Save the read losses status for the next iteration
+        with open(self.ensemble_loss_file, "wb") as memory:
+            pickle.dump(self.read_losses, memory)
 
         if ensemble is not None:
             train_pred = self.predict(set_="train",
@@ -873,10 +873,10 @@ class EnsembleBuilder(object):
         # get the megabytes
         return round(this_model_cost / math.pow(1024, 2), 2)
 
-    def score_ensemble_preds(self) -> bool:
+    def compute_loss_per_model(self) -> bool:
         """
-            score predictions on ensemble building data set;
-            populates self.read_preds and self.read_scores
+            Compute the loss of the predictions on ensemble building data set;
+            populates self.read_preds and self.read_losses
         """
 
         self.logger.debug("Read ensemble data set predictions")
@@ -933,9 +933,9 @@ class EnsembleBuilder(object):
                 self.logger.info('Error loading file (not .npy or .npy.gz): %s', y_ens_fn)
                 continue
 
-            if not self.read_scores.get(y_ens_fn):
-                self.read_scores[y_ens_fn] = {
-                    "ens_score": -np.inf,
+            if not self.read_losses.get(y_ens_fn):
+                self.read_losses[y_ens_fn] = {
+                    "ens_loss": np.inf,
                     "mtime_ens": 0,
                     "mtime_test": 0,
                     "seed": _seed,
@@ -955,45 +955,38 @@ class EnsembleBuilder(object):
                     Y_TEST: None,
                 }
 
-            if self.read_scores[y_ens_fn]["mtime_ens"] == os.path.getmtime(y_ens_fn):
+            if self.read_losses[y_ens_fn]["mtime_ens"] == os.path.getmtime(y_ens_fn):
                 # same time stamp; nothing changed;
                 continue
 
-            # actually read the predictions and score them
-            y_ensemble = self._read_np_fn(y_ens_fn)
-            scores = calculate_score(
-                metrics=self.metrics,
-                target=self.y_true_ensemble,
-                prediction=y_ensemble,
-                task_type=self.task_type,
-            )
+            # actually read the predictions and compute their respective loss
             try:
                 y_ensemble = self._read_np_fn(y_ens_fn)
-                scores = calculate_score(
+                losses = calculate_loss(
                     metrics=self.metrics,
                     target=self.y_true_ensemble,
                     prediction=y_ensemble,
                     task_type=self.task_type,
                 )
 
-                if np.isfinite(self.read_scores[y_ens_fn]["ens_score"]):
+                if np.isfinite(self.read_losses[y_ens_fn]["ens_loss"]):
                     self.logger.debug(
-                        'Changing ensemble score for file %s from %f to %f '
+                        'Changing ensemble loss for file %s from %f to %f '
                         'because file modification time changed? %f - %f',
                         y_ens_fn,
-                        self.read_scores[y_ens_fn]["ens_score"],
-                        scores[self.opt_metric],
-                        self.read_scores[y_ens_fn]["mtime_ens"],
+                        self.read_losses[y_ens_fn]["ens_loss"],
+                        losses[self.opt_metric],
+                        self.read_losses[y_ens_fn]["mtime_ens"],
                         os.path.getmtime(y_ens_fn),
                     )
 
-                self.read_scores[y_ens_fn]["ens_score"] = scores[self.opt_metric]
+                self.read_losses[y_ens_fn]["ens_loss"] = losses[self.opt_metric]
 
                 # It is not needed to create the object here
-                # To save memory, we just score the object.
-                self.read_scores[y_ens_fn]["mtime_ens"] = os.path.getmtime(y_ens_fn)
-                self.read_scores[y_ens_fn]["loaded"] = 2
-                self.read_scores[y_ens_fn]["disc_space_cost_mb"] = self.get_disk_consumption(
+                # To save memory, we just compute the loss.
+                self.read_losses[y_ens_fn]["mtime_ens"] = os.path.getmtime(y_ens_fn)
+                self.read_losses[y_ens_fn]["loaded"] = 2
+                self.read_losses[y_ens_fn]["disc_space_cost_mb"] = self.get_disk_consumption(
                     y_ens_fn
                 )
 
@@ -1005,21 +998,22 @@ class EnsembleBuilder(object):
                     y_ens_fn,
                     traceback.format_exc(),
                 )
-                self.read_scores[y_ens_fn]["ens_score"] = -np.inf
+                self.read_losses[y_ens_fn]["ens_loss"] = np.inf
 
         self.logger.debug(
             'Done reading %d new prediction files. Loaded %d predictions in '
             'total.',
             n_read_files,
-            np.sum([pred["loaded"] > 0 for pred in self.read_scores.values()])
+            np.sum([pred["loaded"] > 0 for pred in self.read_losses.values()])
         )
         return True
 
     def get_n_best_preds(self) -> List[str]:
         """
-            get best n predictions (i.e., keys of self.read_scores)
-            according to score on "ensemble set"
+            get best n predictions (i.e., keys of self.read_losses)
+            according to the loss on the "ensemble set"
             n: self.ensemble_nbest
+
             Side effects:
                 ->Define the n-best models to use in ensemble
                 ->Only the best models are loaded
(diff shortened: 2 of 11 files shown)


.github/workflows/pytest.yml
```diff
@@ -29,7 +29,7 @@ jobs:
     - name: Run tests
       run: |
         if [ ${{ matrix.code-cov }} ]; then codecov='--cov=autoPyTorch --cov-report=xml'; fi
-        python -m pytest --durations=20 --timeout=300 --timeout-method=thread -v $codecov test
+        python -m pytest --durations=20 --timeout=600 --timeout-method=signal -v $codecov test
```
ravinkohli, line 32: just a general question, why are we moving to timeout method signal?

autoPyTorch/ensemble/ensemble_builder.py
```diff
@@ -1033,20 +1027,19 @@ def get_n_best_preds(self) -> List[str]:
         num_keys = len(sorted_keys)
         # remove all that are at most as good as random
         # note: dummy model must have run_id=1 (there is no run_id=0)
-        dummy_scores = list(filter(lambda x: x[2] == 1, sorted_keys))
-
+        dummy_losses = list(filter(lambda x: x[2] == 1, sorted_keys))
         # Leave this here for when we enable dummy classifier/scorer
-        if dummy_scores:
+        if dummy_losses:
```
ravinkohli, line 1032: Could we explicitly test if the list has dummy loss, i.e, if len(dummy_losses) > 0 or something. I think it will make the code more readable.

autoPyTorch/ensemble/ensemble_builder.py
```diff
@@ -1033,20 +1027,19 @@ def get_n_best_preds(self) -> List[str]:
...
-            self.logger.debug("Use %f as dummy score" % dummy_score[1])
-            sorted_keys = list(filter(lambda x: x[1] > dummy_score[1], sorted_keys))
+            num_dummy = len(dummy_losses)
+            dummy_loss = dummy_losses[0]
+            self.logger.debug("Use %f as dummy loss" % dummy_loss[1])
+            sorted_keys = list(filter(lambda x: x[1] < dummy_loss[1], sorted_keys))
 
             # remove Dummy Classifier
             sorted_keys = list(filter(lambda x: x[2] > 1, sorted_keys))
             if not sorted_keys:
```
ravinkohli, line 1041: same here

autoPyTorch/ensemble/ensemble_builder.py
```diff
@@ -1419,32 +1414,33 @@ def _add_ensemble_trajectory(self, train_pred: np.ndarray, test_pred: np.ndarray
...
+            key=lambda x: (x[1], x[2]),
+        ))
         return sorted_keys
 
     def _delete_excess_models(self, selected_keys: List[str]) -> None:
         """
             Deletes models excess models on disc. self.max_models_on_disc
             defines the upper limit on how many models to keep.
-            Any additional model with a worst score than the top
+            Any additional model with a worst loss than the top
```
ravinkohli, line 1443: I think it should be a worse loss than the top

ravinkohli (commented):
This PR looks great. I have added minor comments and then it can be merged

.github/workflows/pytest.yml
ravinkohli: got the answer while reading the group chat. Thanks :P

autoPyTorch/ensemble/ensemble_builder.py
franchuterivera: Thanks,  I just pushed a new commit with this.

franchuterivera: Thanks,  I just pushed a new commit with this.

autoPyTorch/ensemble/ensemble_builder.py
franchuterivera: Thanks,  I just pushed a new commit with this.

autoPyTorch/ensemble/ensemble_builder.py
franchuterivera: Thanks,  I just pushed a new commit with this.

.github/workflows/pytest.yml
franchuterivera: According to the documentation [here](https://pypi.org/project/pytest-timeout/):

> The benefit of this method is that the pytest process is not terminated and the test run can complete normally.

So it sounds like the test will still fail on a timeout, BUT all other test will complete. So this is great to see if everything is working except a particular test. Today, without this, pytest just dies.

ravinkohli: ohh okay. Thanks for the explanation.

ravinkohli (approved):
I think this can be merged once the tests pass.

autoPyTorch/ensemble/ensemble_selection.py
```diff
@@ -71,60 +71,47 @@ def _fit(
...
-                # Memory-efficient averaging!
-                fant_ensemble_prediction.fill(0.0)
+                # fant_ensemble_prediction is the prediction of the current ensemble
+                # and should be ([predictions[selected_prev_iterations] + predictions[j])/(s+1)
+                # We overwrite the contents of fant_ensemble_prediction
+                # directly with weighted_ensemble_prediction + new_prediction and then scale for avg
                 np.add(
-                    fant_ensemble_prediction,
                     weighted_ensemble_prediction,
+                    pred,
```
nabenabe0928, line 94: Can't we just do like `weighted_ensemble_prediction = np.sum(predictions, axis=(<our preference>))`?

autoPyTorch/ensemble/ensemble_selection.py
```diff
@@ -71,60 +71,47 @@ def _fit(
...
                     out=fant_ensemble_prediction
                 )
-                np.add(
+                np.multiply(
                     fant_ensemble_prediction,
-                    (1. / float(s + 1)) * pred,
+                    (1. / float(s + 1)),
```
nabenabe0928, line 99: For me, it is just a redundant code. `fant_ensemble_prediction / float(s + 1)` looks enough

autoPyTorch/ensemble/ensemble_selection.py
```diff
@@ -71,60 +71,47 @@ def _fit(
...
+            if s > 0:
+                np.add(
                     weighted_ensemble_prediction,
-                    (s / float(s + 1)),
+                    ensemble[-1],
                     out=weighted_ensemble_prediction,
                 )
 
+            # Memory-efficient averaging!
             for j, pred in enumerate(predictions):
```
nabenabe0928, line 87: I would like to check here. (When I complete, I will put another comment.)

autoPyTorch/ensemble/ensemble_selection.py
```diff
@@ -71,60 +71,47 @@ def _fit(
...
-                    out=weighted_ensemble_prediction,
-                )
-                np.multiply(
+            if s > 0:
+                np.add(
                     weighted_ensemble_prediction,
-                    (s / float(s + 1)),
+                    ensemble[-1],
                     out=weighted_ensemble_prediction,
                 )
```
nabenabe0928, line 84: I checked this part (inside the IF statement) and the behavior seems correct.
But, are there any reasons why we use `np.add` instead of `np.ndarray += np.ndarray`?

autoPyTorch/ensemble/ensemble_selection.py
```diff
@@ -71,60 +71,47 @@ def _fit(
...
-                    (1. / float(s + 1)) * pred,
+                    (1. / float(s + 1)),
                     out=fant_ensemble_prediction
                 )
 
-                # Calculate score is versatile and can return a dict of score
-                # when all_scoring_functions=False, we know it will be a float
-                score = calculate_score(
+                # Calculate loss is versatile and can return a dict of slosses
+                losses[j] = calculate_loss(
```
nabenabe0928, line 104: For me, the following code is better because we can eliminate the explicit for loop.

```
kwargs = {
    "metrics": [self.metric],
    "target": labels,
    "task_type": self.task_type
}

fant_ensemble_predictions = [
    (weighted_ensemble_prediction + pred) / (s + 1)
    for pred in predictions
]
losses = np.array([
    calculate_loss(prediction=fant_ensemble_prediction, **kwargs)
    ] for fant_ensemble_prediction in fant_ensemble_predictions)
```

nabenabe0928:
```
The code will look like this.

def _fit(...):
    ... some processes

    n_preds = len(predictions)
    kwargs = {
        "metrics": [self.metric],
        "target": labels,
        "task_type": self.task_type
    }

    for i in range(ensemble_size):
        n_selected = len(ensemble)

        if n_selected > 0:
            weighted_ensemble_prediction += ensemble[-1]

        fant_ensemble_predictions = [
            (weighted_ensemble_prediction + pred) / (n_selected + 1)
            for pred in predictions
        ]
        losses = np.array([
            calculate_loss(prediction=fant_ensemble_prediction, **kwargs)
        ] for fant_ensemble_prediction in fant_ensemble_predictions)
        ... some processes
```

nabenabe0928 (changes requested):

autoPyTorch/ensemble/ensemble_builder.py
```diff
@@ -1419,32 +1414,33 @@ def _add_ensemble_trajectory(self, train_pred: np.ndarray, test_pred: np.ndarray
     def _get_list_of_sorted_preds(self) -> List[Tuple[str, float, int]]:
         """
             Returns a list of sorted predictions in descending order
```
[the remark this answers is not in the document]
franchuterivera: Thanks, I will clarify that this is in descending performance order.

autoPyTorch/ensemble/ensemble_builder.py
```diff
@@ -1419,32 +1414,33 @@ def _add_ensemble_trajectory(self, train_pred: np.ndarray, test_pred: np.ndarray
...
-            Scores are taken from self.read_scores.
+            Losses are taken from self.read_losses.
+
             Parameters
             ----------
             None
+
             Return
             ------
             sorted_keys: list
```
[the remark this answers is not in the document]
franchuterivera: Thanks, I will add this as part of the comment.

autoPyTorch/ensemble/ensemble_selection.py
```diff
@@ -71,60 +71,47 @@ def _fit(
             dtype=np.float64,
```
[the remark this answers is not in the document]
franchuterivera: I think np.zeros make here more sense as we are interested in the `'len(array)` more than `np.shape(array)`

autoPyTorch/ensemble/ensemble_selection.py
franchuterivera: Yes, I have coded this so that is super-memory efficient. `+=` internally duplicates the array.

autoPyTorch/ensemble/ensemble_selection.py
franchuterivera: Please do note that the PR here does not changes this logic.

This logic is done like this so that is memory efficient. This can all be coded in almost a single line, but, internally numpy will make a copy of several arrays which will impede ensemble selection on big datasets like `Albert`. For example, every addition without specifying the `out=` directly will likely create a copy of the original array.

nabenabe0928 (changes requested):

autoPyTorch/ensemble/ensemble_selection.py
nabenabe0928: I tested by the following code.
```
import numpy as np
from memory_profiler import profile
import time


# Larger than Albert (https://www.openml.org/d/41147)
A_shape = (500000, 10)


@profile()
def test_original():
    rng = np.random.RandomState(0)
    A = rng.random(A_shape)
    B = np.zeros(A[0].shape, dtype=np.float64)
    C = np.zeros(A[0].shape, dtype=np.float64)
    D = np.zeros(len(A), dtype=np.float64)

    start = time.time()
    for i, a in enumerate(A):
        np.add(B, a, out=C)
        np.multiply(C, 1 / 2, out=C)
        # linear order processing
        D[i] = C.sum()

    print("Original : {:.2f}[s]".format(time.time() - start))


@profile()
def test_mycode():
    rng = np.random.RandomState(0)
    A = rng.random(A_shape)
    B = np.zeros(A[0].shape, dtype=np.float64)

    start = time.time()
    C = [(B + a) / 2 for a in A]
    # linear order processing
    D = [c.sum() for c in C]

    print("My Code : {:.2f}[s]".format(time.time() - start))


test_original()
test_mycode()
```

The results were the followings:
```
Original : 132.43[s]
Filename: stress_check.py

Line #    Mem usage    Increment  Occurences   Line Contents
============================================================
    10     62.4 MiB     62.4 MiB           1   @profile()
    11                                         def test_original():
    12     62.5 MiB      0.2 MiB           1       rng = np.random.RandomState(0)
    13    100.7 MiB     38.2 MiB           1       A = rng.random(A_shape)
    14    100.7 MiB      0.0 MiB           1       B = np.zeros(A[0].shape, dtype=np.float64)
    15    100.7 MiB      0.0 MiB           1       C = np.zeros(A[0].shape, dtype=np.float64)
    16    100.7 MiB      0.0 MiB           1       D = np.zeros(len(A), dtype=np.float64)
    17                                         
    18    100.7 MiB      0.0 MiB           1       start = time.time()
    19    104.6 MiB      0.0 MiB      500001       for i, a in enumerate(A):
    20    104.6 MiB      0.0 MiB      500000           np.add(B, a, out=C)
    21    104.6 MiB      0.0 MiB      500000           np.multiply(C, 1 / 2, out=C)
    22                                                 # linear order processing
    23    104.6 MiB      3.8 MiB      500000           D[i] = C.sum()
    24                                         
    25    104.6 MiB      0.0 MiB           1       print("Original : {:.2f}[s]".format(time.time() - start))
    26    104.6 MiB      0.0 MiB           1       return D


My Code : 69.62[s]
Filename: stress_check.py

Line #    Mem usage    Increment  Occurences   Line Contents
============================================================
    29     62.6 MiB     62.6 MiB           1   @profile()
    30                                         def test_mycode():
    31     62.6 MiB      0.0 MiB           1       rng = np.random.RandomState(0)
    32    100.6 MiB     38.0 MiB           1       A = rng.random(A_shape)
    33    100.6 MiB      0.0 MiB           1       B = np.zeros(A[0].shape, dtype=np.float64)
    34                                         
    35    100.6 MiB      0.0 MiB           1       start = time.time()
    36    205.3 MiB    104.7 MiB      500003       C = [(B + a) / 2 for a in A]
    37                                             # linear order processing
    38    224.6 MiB     19.3 MiB      500003       D = [c.sum() for c in C]
    39                                         
    40    224.6 MiB      0.0 MiB           1       print("My Code : {:.2f}[s]".format(time.time() - start))
    41    224.6 MiB      0.0 MiB           1       return D
```

As you mentioned, the memory usage is huge in my case, but do you think the benefits of memory reduction surpass the benefits of computational time?
If you process the size of Albert dataset, it will take `100 sec` (twice slower) for 1 iteration and you need to do the same process for the number of learners in the ensemble, right?

In current computational resources, `100MB` difference is not a problem, but `50 sec` differences per call is quite large difference, isn't it?

nabenabe0928 (changes requested):

autoPyTorch/ensemble/ensemble_selection.py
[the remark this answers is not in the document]
nabenabe0928: I did not get your point. Which line shows `len(array)`?

autoPyTorch/ensemble/ensemble_selection.py
nabenabe0928: I improved the speed more.
With 20% more memory, 10000 times faster in the case of Albert size.

```
import numpy as np
from memory_profiler import profile
import time


# Larger than Albert (https://www.openml.org/d/41147)
A_shape = (500000, 5)


@profile()
def test_original():
    rng = np.random.RandomState(0)
    A = rng.random(A_shape)
    B = np.zeros(A[0].shape, dtype=np.float64)
    C = np.zeros(A[0].shape, dtype=np.float64)
    D = np.zeros(len(A), dtype=np.float64)

    start = time.time()
    for i, a in enumerate(A):
        np.add(B, a, out=C)
        np.multiply(C, 1 / 2, out=C)
        # linear order processing
        D[i] = C.sum()

    print("Original : {:.2f}[s]".format(time.time() - start))
    return D


@profile()
def test_mycode():
    rng = np.random.RandomState(0)
    A = rng.random(A_shape)
    B = np.zeros(A[0].shape, dtype=np.float64)

    start = time.time()
    C = (A + B) / 2
    # linear order processing
    D = C.sum(axis=1)

    print("My Code : {:.2f}[s]".format(time.time() - start))
    return D


original = test_original()
mine = test_mycode()
cnt = np.count_nonzero(np.abs(original - mine) > 1e-5)
print(f"# of wrong answers: {cnt}")

```

The results are the followings:
```
Original : 140.89[s]
Filename: stress_check.py

Line #    Mem usage    Increment  Occurences   Line Contents
============================================================
    10     62.3 MiB     62.3 MiB           1   @profile()
    11                                         def test_original():
    12     62.5 MiB      0.2 MiB           1       rng = np.random.RandomState(0)
    13     81.4 MiB     18.9 MiB           1       A = rng.random(A_shape)
    14     81.4 MiB      0.0 MiB           1       B = np.zeros(A[0].shape, dtype=np.float64)
    15     81.4 MiB      0.0 MiB           1       C = np.zeros(A[0].shape, dtype=np.float64)
    16     81.4 MiB      0.0 MiB           1       D = np.zeros(len(A), dtype=np.float64)
    17                                         
    18     81.4 MiB      0.0 MiB           1       start = time.time()
    19     85.3 MiB      0.0 MiB      500001       for i, a in enumerate(A):
    20     85.3 MiB      0.0 MiB      500000           np.add(B, a, out=C)
    21     85.3 MiB      0.0 MiB      500000           np.multiply(C, 1 / 2, out=C)
    22                                                 # linear order processing
    23     85.3 MiB      3.8 MiB      500000           D[i] = C.sum()
    24                                         
    25     85.3 MiB      0.0 MiB           1       print("Original : {:.2f}[s]".format(time.time() - start))
    26     85.3 MiB      0.0 MiB           1       return D


My Code : 0.02[s]
Filename: stress_check.py

Line #    Mem usage    Increment  Occurences   Line Contents
============================================================
    29     66.4 MiB     66.4 MiB           1   @profile()
    30                                         def test_mycode():
    31     66.4 MiB      0.0 MiB           1       rng = np.random.RandomState(0)
    32     85.5 MiB     19.0 MiB           1       A = rng.random(A_shape)
    33     85.5 MiB      0.0 MiB           1       B = np.zeros(A[0].shape, dtype=np.float64)
    34                                         
    35     85.5 MiB      0.0 MiB           1       start = time.time()
    36    104.5 MiB     19.0 MiB           1       C = (A + B) / 2
    37                                             # linear order processing
    38    108.7 MiB      4.2 MiB           1       D = C.sum(axis=1)
    39                                         
    40    108.7 MiB      0.0 MiB           1       print("My Code : {:.2f}[s]".format(time.time() - start))
    41    108.7 MiB      0.0 MiB           1       return D


# of wrong answers: 0
```

franchuterivera: I do not think you can simply use memory profiler and assess the implication o this change. For example, see [this](https://stackoverflow.com/questions/6018986/memory-profiler-for-numpy). Also, depending on the multiprocessing strategy, you will have to account for the parent-to-child memory overhead, which cannot be easily measured in isolation.

I think the objective of this PR is other, which is to move to a minimization problem. There is no change or proposed change to the for loop strategy of ensemble selection.

Nevertheless, if you can demonstrate a speed improvement that does not crashes due to real virtual memory as measured by [resource module](https://docs.python.org/3/library/resource.html) that pynysher uses, please propose a new PR for this change. I will recommend to actually launch a run with albert with and without this change, and measure the speed performance during execution. You are likely going to see that the bottleneck is somewhere else in ensemble selection, which is during score calculation in scikit learn.

autoPyTorch/ensemble/ensemble_builder.py
```diff
@@ -1419,32 +1414,33 @@ def _add_ensemble_trajectory(self, train_pred: np.ndarray, test_pred: np.ndarray
...
-                (k, v["ens_score"], v["num_run"])
-                for k, v in self.read_scores.items()
+                (k, v["ens_loss"], v["num_run"])
+                for k, v in self.read_losses.items()
             ],
-            key=lambda x: x[2],
-        )))  # type: List[Tuple[str, float, int]]
-        # Then by score
-        sorted_keys = list(reversed(sorted(sorted_keys, key=lambda x: x[1])))
+            # Sort by loss as priority 1 and then by num_run on a ascending order
```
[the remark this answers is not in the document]
nabenabe0928: I think the statement is not clear especially `priority 1`. Please clarify more. I would suggest the previous comment as it is mathematically clear and no one has doubts.

autoPyTorch/ensemble/ensemble_selection.py
nabenabe0928: Ok, i got you. I will send a new PR later.

autoPyTorch/ensemble/ensemble_selection.py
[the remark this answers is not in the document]
nabenabe0928: I will add small changes later then.

autoPyTorch/ensemble/ensemble_selection.py
```diff
@@ -71,60 +71,47 @@ def _fit(
...
         )
         for i in range(ensemble_size):
-            scores = np.zeros(
+            losses = np.zeros(
```
nabenabe0928, line 74: Could you add the doc-string in this function `_fit()`?

nabenabe0928 (changes requested):

autoPyTorch/ensemble/ensemble_selection.py
```diff
@@ -231,13 +231,15 @@ def get_models_with_weights(
...
-        Arguments:
+        Args:
             models (List[Tuple[float, BasePipeline]]):
                 A dictionary that maps a model's name to it's actual python object.
 
         Returns:
-            List[Tuple[float, BasePipeline]]:
-                each model with the related weight, sorted by performance
+            output (List[Tuple[float, BasePipeline]]):
+                each model with the related weight, sorted by ascending
```
nabenabe0928, line 240: sorted by the ascending order with respect to the performance

nabenabe0928: Ok, but errors in grammar are not good sometimes, because it can cause some misunderstandings. I addressed this point, so basically I could understand this statement (just a grammar issue).

PR #118, opened 2021-02-25
handle nans in categorical columns

ravinkohli:
This PR fixes the error we get when there are nans in categorical columns

### DIFF
--- a/autoPyTorch/data/tabular_feature_validator.py
+++ b/autoPyTorch/data/tabular_feature_validator.py
@@ -54,7 +54,8 @@ class TabularFeatureValidator(BaseFeatureValidator):
                     if X[column].isna().all():
                         X[column] = pd.to_numeric(X[column])
                         # Also note this change in self.dtypes
-                        self.dtypes[list(X.columns).index(column)] = X[column].dtype
+                        if len(self.dtypes) != 0:
+                            self.dtypes[list(X.columns).index(column)] = X[column].dtype
 
             self.enc_columns, self.feat_type = self._get_columns_to_encode(X)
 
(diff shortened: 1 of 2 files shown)


test/test_data/test_feature_validator.py
```diff
@@ -231,10 +231,14 @@ def test_featurevalidator_unsupported_numpy(input_data_featuretest):
...
+def test_featurevalidator_categorical_nan(input_data_featuretest):
     validator = TabularFeatureValidator()
-    with pytest.raises(ValueError, match=r"Categorical features in a dataframe.*missing/NaN"):
-        validator.fit(input_data_featuretest)
+    validator.fit(input_data_featuretest)
+    transformed_X = validator.transform(input_data_featuretest)
+    assert np.shape(input_data_featuretest) == np.shape(transformed_X)
+    assert np.issubdtype(transformed_X.dtype, np.number)
+    assert validator._is_fitted
+    assert isinstance(transformed_X, np.ndarray)
```
franchuterivera, line 241: Can you add a check so that we see that a nan was encoded?

LIke, can you add a check that:

- makes sure that TabularFeatureValidator encoder categories consider the -1 and that input_data_featuretest has a nan?

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -53,10 +53,25 @@ def _fit(
                 for column in X.columns:
                     if X[column].isna().all():
                         X[column] = pd.to_numeric(X[column])
+                        # Also note this change in self.dtypes
+                        self.dtypes[list(X.columns).index(column)] = X[column].dtype
```
franchuterivera: This is a great catch!

franchuterivera (commented):
Just to make sure the test is working, can you add tune up the check to make sure that -1 is in the encoder categories, and that the input has a nan?

PR #121, opened 2021-03-02
Run history traditional

### LINKED ISSUE #96: Create a runhistory for the traditional pipeline
Smac creates a run history with the performance of the models.

Traditional pipeline is not included here. 

Can we add a similar runhistory for the traditional pipeline to know how the performance of this estimator was?

ravinkohli:
Closes #96


autoPyTorch/api/base_task.py
```diff
@@ -600,6 +602,14 @@ def _do_traditional_prediction(self, num_run: int, time_for_traditional: int) ->
             status, cost, runtime, additional_info = future.result()
```
franchuterivera, line 602: Can you encapsulate this in a try/except block. In dask, when you get the result of a configuration, if an exception happen, it will be raised when collecting the result. If 1 configuration failed, we can show the error to the user but 'continue' as other configurations maybe work.

autoPyTorch/evaluation/abstract_evaluator.py
```diff
@@ -374,7 +375,7 @@ def _loss(self, y_true: np.ndarray, y_hat: np.ndarray) -> Dict[str, float]:
                 For logloss for example: optimum(0) - (-1 * actual score)
         """
 
-        if not isinstance(self.configuration, Configuration):
+        if isinstance(self.configuration, int):
             return {self.metric.name: 1.0}
```
franchuterivera, line 379: I am sorry, I know this is not related to this PR, but can you update this to be self.metric-> worst possible result instead of 1.0?

franchuterivera: Well actually a loss....

autoPyTorch/api/base_task.py
```diff
@@ -616,7 +626,8 @@ def _do_traditional_prediction(self, num_run: int, time_for_traditional: int) ->
                         "Traditional prediction for %s failed with run state %s and additional output: %s.",
                         classifier, str(status), str(additional_info),
                     )
-        return num_run
+        self._logger.debug("Run history traditional: {}".format(run_history))
```
franchuterivera, line 629: Can you write this to a json file in `.autopytorch` directory so we have this for debug?

autoPyTorch/api/base_task.py
ravinkohli: As far as I understand, if we have a failed configuration, getting results from dask will work fine. By 'if an exception happen' do you mean an exception that happens outside of the TAE block?

autoPyTorch/evaluation/abstract_evaluator.py
ravinkohli: Definitely, I can do this

autoPyTorch/api/base_task.py
ravinkohli: sure

autoPyTorch/api/base_task.py
```diff
@@ -551,16 +551,18 @@ def _do_dummy_prediction(self, num_run: int) -> None:
                     % (str(status), str(additional_info))
                 )
 
-    def _do_traditional_prediction(self, num_run: int, time_for_traditional: int) -> int:
+    def _do_traditional_prediction(self, num_run: int, time_for_traditional: int) -> Tuple[int, RunHistory]:
```
nabenabe0928, line 554: Could you add a doc-string for the function?

autoPyTorch/api/base_task.py
```diff
@@ -888,6 +909,11 @@ def _search(
                 pd.DataFrame(self.ensemble_performance_history).to_json(
                     os.path.join(self._backend.internals_directory, 'ensemble_history.json'))
 
+        # Add run history of traditional algorithms to the full run history
+        if hasattr(self, 'run_history') and self.run_history is not None and traditional_run_history is not None:
```
nabenabe0928, line 913: `None not in [self.run_history, traditional_run_history]` is a bit shorter:-)

franchuterivera (changes requested):
There is something broken with the deleted files...

This file  should not be empty:
```
ls -l test/test_ensemble/.autoPyTorch/runs/0_3_100.0/predictions_valid_0_3_100.0.npy 
-rw-r--r-- 1 chico chico 160 Feb 18 18:14 test/test_ensemble/.autoPyTorch/runs/0_3_100.0/predictions_valid_0_3_100.0.npy
```

Can you remove this change from your PR?

autoPyTorch/evaluation/abstract_evaluator.py
```diff
@@ -85,8 +85,9 @@ def predict(self, X: Union[np.ndarray, pd.DataFrame],
     def estimator_supports_iterative_fit(self) -> bool:  # pylint: disable=R0201
         return False
 
-    def get_additional_run_info(self) -> None:  # pylint: disable=R0201
-        return None
+    def get_additional_run_info(self) -> Dict[str, Any]:  # pylint: disable=R0201
+        return {'pipeline_configuration': self.configuration,
+                'trainer_configuration': self.pipeline.named_steps['model_trainer'].choice.model.get_config()}
```
nabenabe0928, line 90: It is not your fault, but I could not know what the `named_steps['model_trainer']` is.
If I understand codes correctly, it should be some models, right?
Could you put the reference to the file path (e.g. base class, typical examples of models) in the doc-string?

test/test_api/test_api.py
```diff
@@ -102,19 +104,24 @@ def test_tabular_classification(openml_id, resampling_strategy, backend):
     ) if 'SUCCESS' in str(run_value.status)]
     assert len(succesful_runs) > 1, [(k, v) for k, v in estimator.run_history.data.items()]
 
+    # Delay the rest of the test as it can take time to store model files
```
nabenabe0928, line 107: What do you mean? Could you explain a bit more? Preferably in both the comment and the next change.

test/test_api/test_api.py
```diff
@@ -272,17 +279,20 @@ def test_tabular_regression(openml_name, resampling_strategy, backend):
 
     # Search for an existing run key in disc. A individual model might have
     # a timeout and hence was not written to disc
+    successful_num_run = None
+    SUCCESS = False
     for i, (run_key, value) in enumerate(estimator.run_history.data.items()):
-        if 'SUCCESS' not in str(value.status):
```
nabenabe0928, line 276: Could you unite this process?  We have the two identical processings. (Later should be fine.)

nabenabe0928 (changes requested):

franchuterivera:
Also, let us think about the following scenario. Things go wrong, and only the traditional pipeline was fitted.

If you this [line](https://github.com/automl/Auto-PyTorch/blob/2711aed0749c7792ee4371c69958a4d9dccae05a/autoPyTorch/api/base_task.py#L176) Run history is zero and because of [this](https://github.com/automl/Auto-PyTorch/blob/2711aed0749c7792ee4371c69958a4d9dccae05a/autoPyTorch/api/base_task.py#L913)  we will not update this with the traditional pipeline run history and the singlebest ensemble will not kick in.

What do you think about making `self.run_history = {}` since the beginning and we update that dict accordingly?

ravinkohli:
> There is something broken with the deleted files...
> 
> This file should not be empty:
> 
> ```
> ls -l test/test_ensemble/.autoPyTorch/runs/0_3_100.0/predictions_valid_0_3_100.0.npy 
> -rw-r--r-- 1 chico chico 160 Feb 18 18:14 test/test_ensemble/.autoPyTorch/runs/0_3_100.0/predictions_valid_0_3_100.0.npy
> ```
> 
> Can you remove this change from your PR?

Sure

ravinkohli:
> Also, let us think about the following scenario. Things go wrong, and only the traditional pipeline was fitted.
> 
> If you this [line](https://github.com/automl/Auto-PyTorch/blob/2711aed0749c7792ee4371c69958a4d9dccae05a/autoPyTorch/api/base_task.py#L176) Run history is zero and because of [this](https://github.com/automl/Auto-PyTorch/blob/2711aed0749c7792ee4371c69958a4d9dccae05a/autoPyTorch/api/base_task.py#L913) we will not update this with the traditional pipeline run history and the singlebest ensemble will not kick in.
> 
> What do you think about making `self.run_history = {}` since the beginning and we update that dict accordingly?

Hey, I had not thought of this scenario. Yes, I think it makes sense to have self.run_history maintain the data property of the runhistories as and when we get it either for traditional or smac. I'll do this

test/test_api/test_api.py
ravinkohli: do you mean the process in test_tabular_classification and test_tabular_regression? I'm sorry I did not understand which two processes to unite

test/test_api/test_api.py
ravinkohli: actually I'll remove this. It is not important

autoPyTorch/evaluation/abstract_evaluator.py
ravinkohli: sure

autoPyTorch/api/base_task.py
ravinkohli: Actually according to @franchuterivera requested changes, this line is no longer needed

autoPyTorch/evaluation/abstract_evaluator.py
franchuterivera: BTW, you can read about named steps from scikit learn [here](https://scikit-learn.org/stable/modules/generated/sklearn.pipeline.Pipeline.html).

nabenabe0928: @franchuterivera Hum, but it does not help me...

test/test_api/test_api.py
nabenabe0928: Yes, yes those tests. It has the same process, right? (but since it is a test code, if you think it is not necessary, it is also ok to ignore this comment)

ravinkohli: yeah there are subtle differences in 1 or 2 lines so I think its fine to not unite them

autoPyTorch/api/base_task.py
```diff
@@ -173,7 +173,7 @@ def __init__(
         self._dataset_requirements: Optional[List[FitRequirement]] = None
         self._metric: Optional[autoPyTorchMetric] = None
         self._logger: Optional[PicklableClientLogger] = None
-        self.run_history: Optional[RunHistory] = None
+        self.run_history: Dict = {}
```
franchuterivera, line 176: Do you think it is possible to make this a:
```
Optional[Runhistory] = None
```
There is rich information in the runhistory, like the configuration to id, among other things. If we only store the dictionary, we lose them.

autoPyTorch/api/base_task.py
```diff
@@ -677,6 +688,10 @@ def _do_traditional_prediction(self, num_run: int, time_left: int, func_eval_tim
                                      "Please consider increasing the run time to further improve performance.")
                 break
 
+        self._logger.debug("Run history traditional: {}".format(run_history))
+        # add run history of traditional to api run history
+        self.run_history.update(run_history.data)
```
franchuterivera, line 693: And then here we can create a wrapper method or something like this that calls `runhistory.add()` to update the runhistory as needed

autoPyTorch/api/base_task.py
```diff
@@ -947,8 +962,9 @@ def _search(
                 search_space_updates=self.search_space_updates
             )
             try:
-                self.run_history, self.trajectory, budget_type = \
+                run_history, self.trajectory, budget_type = \
                     _proc_smac.run_smbo()
+                self.run_history.update(run_history.data)
```
franchuterivera, line 967: same here

test/test_api/test_api.py
```diff
@@ -98,23 +100,26 @@ def test_tabular_classification(openml_id, resampling_strategy, backend):
         assert os.path.exists(os.path.join(tmp_dir, expected_file)), expected_file
 
     # Check that smac was able to find proper models
-    succesful_runs = [run_value.status for run_value in estimator.run_history.data.values(
+    succesful_runs = [run_value.status for run_value in estimator.run_history.values(
     ) if 'SUCCESS' in str(run_value.status)]
-    assert len(succesful_runs) > 1, [(k, v) for k, v in estimator.run_history.data.items()]
+    assert len(succesful_runs) > 1, [(k, v) for k, v in estimator.run_history.items()]
```
franchuterivera, line 105: Then we can stil luse the `.data` here

franchuterivera (changes requested):
Some minor changes but everything else looks great

autoPyTorch/api/base_task.py
```diff
@@ -651,6 +655,13 @@ def _do_traditional_prediction(self, num_run: int, time_left: int, func_eval_tim
                     if status == StatusType.SUCCESS:
                         self._logger.info(
                             f"Fitting {cls} took {runtime}s, performance:{cost}/{additional_info}")
+                        configuration = additional_info['pipeline_configuration']
+                        origin = additional_info['configuration_origin']
+                        del additional_info['configuration_origin']
```
nabenabe0928, line 660: I think you know it, but this `del` does not change the memory usage unless you re-define `configuration` or `origin` at some points.
```
In [1]: d = {}

In [2]: d[0] = [1, 2, 3, 4]

In [3]: a = d[0]

In [4]: id(a) == id(d[0])
Out[4]: True

In [5]: a[2] = 6

In [6]: id(a) == id(d[0])
Out[6]: True

In [7]: a = []

In [8]: id(a) == id(d[0])
Out[8]: False
```

ravinkohli: actually, I did not know this. The idea was also not to reduce memory usage. But I think I can remove these lines as the intended purpose is no longer valid

PR #127, opened 2021-03-08
[Doc] First push of the developer documentation

franchuterivera:
First Push for the developer documentation.

PR #141, opened 2021-03-16
[Feat] Better traditional pipeline cutoff time

franchuterivera:
This PR addresses #130 . It does so by making sure we have enough time to fit a traditional pipeline.

Also, it guarantees that at least 1 traditional pipeline and 1 neural network are trained.

### DIFF
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -650,8 +650,8 @@ class BaseTask:
 
             # Exit if no more time is available for a new classifier
             if time_left < func_eval_time_limit:
-                self.logger.warning("Not enough time to fit all traditional machine learning models."
-                                    "Please consider increasing the run time to further improve performance.")
+                self._logger.warning("Not enough time to fit all traditional machine learning models."
+                                     "Please consider increasing the run time to further improve performance.")
                 break
 
         return num_run
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -573,7 +573,7 @@ class BaseTask:
         Returns:
             num_run: (int)
                 The incremented identifier index. This depends on how many machine learning
-                models where fitted.
+                models were fitted.
         """
 
         # Mypy Checkings -- Traditional prediction is only called for search
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -625,6 +625,14 @@ class BaseTask:
                 # Increment the launched job index
                 num_run = n_r
 
+            # When managing time, we need to take into account the allocated time resources,
+            # which are dependent on the number of cores. 'dask_futures' is a proxy to the number
+            # of workers /n_jobs that we have, in that if there are 4 cores allocated, we can run at most
+            # 4 task in parallel. Every 'cutoff' seconds, we generate up to 4 tasks.
+            # If we only have 4 workers and there are 4 futures in dask_futures, it means that every
+            # worker has a task. We would not like to launch another job until a worker is available. To this
+            # end, the following if-statement queries the number of active jobs, and forces to wait for a job
+            # completion via future.result(), so that a new worker is available for the next iteration.
             if len(dask_futures) >= self.n_jobs:
 
                 # How many workers to wait before starting fitting the next iteration
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -551,7 +551,30 @@ class BaseTask:
                     % (str(status), str(additional_info))
                 )
 
-    def _do_traditional_prediction(self, num_run: int, time_for_traditional: int) -> int:
+    def _do_traditional_prediction(self, num_run: int, time_left: int, func_eval_time_limit: int
+                                   ) -> int:
+        """
+        Fits traditional machine learning algorithms to the provided dataset, while
+        complying with time resource allocation.
+
+        This method currently only supports classification.
+
+        Args:
+            num_run: (int)
+                An identifier to indicate the current machine learning algorithm
+                being processed
+            time_left: (int)
+                Hard limit on how many machine learning algorithms can be fit. Depending on how
+                fast a traditional machine learning algorithm trains, it will allow multiple
+                models to be fitted.
+            func_eval_time_limit: (int)
+                Maximum training time each algorithm is allowed to take, during training
+
+        Returns:
+            num_run: (int)
+                The incremented identifier index. This depends on how many machine learning
+                models where fitted.
+        """
 
         # Mypy Checkings -- Traditional prediction is only called for search
         # where the following objects are created
@@ -559,18 +582,18 @@ class BaseTask:
         assert self._logger is not None
         assert self._dask_client is not None
 
-        self._logger.info("Starting to create dummy predictions.")
+        self._logger.info("Starting to create traditional pipeline predictions.")
 
         memory_limit = self._memory_limit
         if memory_limit is not None:
             memory_limit = int(math.ceil(memory_limit))
         available_classifiers = get_available_classifiers()
-        dask_futures = list()
-        time_for_traditional_classifier_sec = int(time_for_traditional / len(available_classifiers))
+        dask_futures = []
+
         for n_r, classifier in enumerate(available_classifiers, start=num_run):
             start_time = time.time()
             scenario_mock = unittest.mock.Mock()
-            scenario_mock.wallclock_limit = time_for_traditional_classifier_sec
+            scenario_mock.wallclock_limit = time_left
             # This stats object is a hack - maybe the SMAC stats object should
             # already be generated here!
             stats = Stats(scenario_mock)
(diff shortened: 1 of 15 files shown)


autoPyTorch/api/base_task.py
```diff
@@ -551,89 +551,142 @@ def _do_dummy_prediction(self, num_run: int) -> None:
...
+                    self._dask_client.submit(
+                        ta.run, config=classifier,
+                        cutoff=func_eval_time_limit,
+                    )
+                ])
+
+                # Increment the launched job index
+                num_run = n_r
+
+            if len(dask_futures) >= self.n_jobs:
```
ravinkohli, line 628: Could you explain how this portion is waiting for resources? I did not quite understand this part of the code

franchuterivera: thanks for the feedback, I will add a comment explaining this. Let me elaborate here before the push. Will something like this works?
```
When managing time, we need to take into account the allocated time resources,
which are dependent on the number of cores. 'dask_futures' is a proxy to the number 
of workers /n_jobs that we have, in that if there are 4 cores allocated, we can run at most
4 task in parallel. Every 'cutoff' seconds, we generate up to 4 tasks.
If we only have 4 workers and there are 4 futures in dask_futures, it means that every
worker has a task. We would not like to launch another job until a worker is available. To this
end, the following if-statement queries the number of active jobs, and forces to wait for a job
completion via future.result(), so that a new worker is available for the next iteration. 
```

ravinkohli: Yes, this works. Thanks for the explanation.

autoPyTorch/api/tabular_regression.py
```diff
@@ -103,26 +103,27 @@ def _get_required_dataset_properties(self, dataset: BaseDataset) -> Dict[str, An
...
+        optimize_metric: str,
+        X_train: Optional[Union[List, pd.DataFrame, np.ndarray]] = None,
+        y_train: Optional[Union[List, pd.DataFrame, np.ndarray]] = None,
+        X_test: Optional[Union[List, pd.DataFrame, np.ndarray]] = None,
+        y_test: Optional[Union[List, pd.DataFrame, np.ndarray]] = None,
+        dataset_name: Optional[str] = None,
+        budget_type: Optional[str] = None,
+        budget: Optional[float] = None,
+        total_walltime_limit: int = 100,
+        func_eval_time_limit: Optional[int] = None,
```
nabenabe0928, line 117: As mentioned below, I think `np.inf` is better.

autoPyTorch/api/tabular_classification.py
```diff
@@ -156,16 +156,20 @@ def search(
...
                 this value high enough so that typical machine
                 learning algorithms can be fit on the training
                 data.
-            traditional_per_total_budget (float), (default=0.1):
-                Percent of total walltime to be allocated for
-                running traditional classifiers.
+            enable_traditional_pipeline (bool), (default=True):
+                We fit traditional machine learning algorithms
+                (LightGBM, CatBoost, RandomForest, ExtraTrees, KNN, SVM)
+                prior building PyTorch Neural Networks. You can disable this
```
nabenabe0928, line 169: prior to? (before?)

autoPyTorch/api/base_task.py
```diff
@@ -764,16 +851,22 @@ def _search(
 
         # ============> Run traditional ml
 
-        traditional_task_name = 'runTraditional'
-        self._stopwatch.start_task(traditional_task_name)
-        elapsed_time = self._stopwatch.wall_elapsed(self.dataset_name)
-        time_for_traditional = int(traditional_per_total_budget * max(0, (self._time_for_task - elapsed_time)))
-        if time_for_traditional <= 0:
-            if traditional_per_total_budget > 0:
```
nabenabe0928, line 772: Is it okay to delete this ValueError?

autoPyTorch/api/base_task.py
```diff
@@ -551,89 +551,150 @@ def _do_dummy_prediction(self, num_run: int) -> None:
                     % (str(status), str(additional_info))
                 )
 
-    def _do_traditional_prediction(self, num_run: int, time_for_traditional: int) -> int:
+    def _do_traditional_prediction(self, num_run: int, time_left: int, func_eval_time_limit: int
```
nabenabe0928, line 554: I haven't checked this file (`base_tasks.py`), but checked all the others.

autoPyTorch/api/base_task.py
franchuterivera: I have changed how the logic get's implemented, so it did not made sense to me to have it here. Now, if there is not enough time to fit a set of traditional pipelines, you will end up [here](https://github.com/automl/Auto-PyTorch/blob/0b641b0615c74e5d78d3b633caaee0b1d5bc509b/autoPyTorch/api/base_task.py#L668).

I can add an additional message if you think we need it, but fundamentally if the user provides a negative time for example, it means he wants to disable this feature.

autoPyTorch/api/tabular_classification.py
franchuterivera: thanks, I have changed this.

autoPyTorch/pipeline/components/setup/traditional_ml/classifier_models/__init__.py
```diff
@@ -13,12 +13,13 @@
     SVMModel)
 
 _classifiers = {
+    # Sort by fit importance
```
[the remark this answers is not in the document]
franchuterivera: I added a clarification in the code for this. Let me know fi this doesn't make sense.

autoPyTorch/api/tabular_regression.py
```diff
@@ -147,16 +148,16 @@ def search(self,
                 in seconds for the search of appropriate models.
                 By increasing this value, autopytorch has a higher
                 chance of finding better models.
-            func_eval_time_limit (int), (default=60): Time limit
+            func_eval_time_limit (int), (default=None): Time limit
```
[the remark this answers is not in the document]
franchuterivera: I have changed the variable name to contain secs, as an indication that it should be seconds.

np.inf means no time constraint, which is different than having a `None`. When we have a None, this variable time get's internally calculated. I have added this in the doc string. Please let me know what you think.

ravinkohli (approved):

autoPyTorch/api/base_task.py
nabenabe0928: Ok, I just wanted to confirm

autoPyTorch/pipeline/components/setup/traditional_ml/classifier_models/__init__.py
```diff
@@ -13,7 +13,16 @@
     SVMModel)
 
 _classifiers = {
-    # Sort by fit importance
+    # Sort by more robust models
```
nabenabe0928, line 16: Thanks:)

autoPyTorch/api/tabular_regression.py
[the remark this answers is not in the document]
nabenabe0928: Ok, I checked the doc-string and it makes much more sense to me.
Btw, `secs` sounds... not good (I mean the pronunciation), right?
In physics, we use sec, so `sec` should be fine

autoPyTorch/api/base_task.py
```diff
@@ -551,89 +551,150 @@ def _do_dummy_prediction(self, num_run: int) -> None:
...
+        """
+        Fits traditional machine learning algorithms to the provided dataset, while
+        complying with time resource allocation.
+
+        This method currently only supports classification.
+
+        Args:
+            num_run: (int)
+                An identifier to indicate the current machine learning algorithm
+                being processed
```
nabenabe0928, line 565: Does `num_run` mean like the job id of a given ML algorithm?
I did not get this variable `num_run`.

autoPyTorch/api/base_task.py
```diff
@@ -551,89 +551,150 @@ def _do_dummy_prediction(self, num_run: int) -> None:
...
+                Hard limit on how many machine learning algorithms can be fit. Depending on how
+                fast a traditional machine learning algorithm trains, it will allow multiple
+                models to be fitted.
+            func_eval_time_limit_secs: (int)
+                Maximum training time each algorithm is allowed to take, during training
+
+        Returns:
+            num_run: (int)
+                The incremented identifier index. This depends on how many machine learning
+                models where fitted.
```
nabenabe0928, line 576: where -> were

autoPyTorch/api/base_task.py
```diff
@@ -551,89 +551,150 @@ def _do_dummy_prediction(self, num_run: int) -> None:
...
-
         memory_limit = self._memory_limit
         if memory_limit is not None:
             memory_limit = int(math.ceil(memory_limit))
         available_classifiers = get_available_classifiers()
-        dask_futures = list()
-        time_for_traditional_classifier_sec = int(time_for_traditional / len(available_classifiers))
+        dask_futures = []
+
+        total_number_classifiers = len(available_classifiers) + num_run
```
nabenabe0928, line 591: check the definition later (relate to `num_run`)

autoPyTorch/api/base_task.py
```diff
@@ -551,89 +551,150 @@ def _do_dummy_prediction(self, num_run: int) -> None:
...
+                stats = Stats(scenario_mock)
+                stats.start_timing()
+                ta = ExecuteTaFuncWithQueue(
+                    backend=self._backend,
+                    seed=self.seed,
+                    metric=self._metric,
+                    logger_port=self._logger_port,
+                    cost_for_crash=get_cost_of_crash(self._metric),
+                    abort_on_first_run_crash=False,
+                    initial_num_run=n_r,
```
nabenabe0928, line 611: It looks the definitions of `num_run` and `n_r` changed. `num_run` and `n_r` are identifiers of algorithms, right?

autoPyTorch/api/base_task.py
```diff
@@ -551,89 +551,150 @@ def _do_dummy_prediction(self, num_run: int) -> None:
...
+            # 4 task in parallel. Every 'cutoff' seconds, we generate up to 4 tasks.
+            # If we only have 4 workers and there are 4 futures in dask_futures, it means that every
+            # worker has a task. We would not like to launch another job until a worker is available. To this
+            # end, the following if-statement queries the number of active jobs, and forces to wait for a job
+            # completion via future.result(), so that a new worker is available for the next iteration.
+            if len(dask_futures) >= self.n_jobs:
+
+                # How many workers to wait before starting fitting the next iteration
+                workers_to_wait = 1
+                if n_r >= total_number_classifiers - 1 or time_left <= func_eval_time_limit_secs:
```
nabenabe0928, line 640: I haven't checked inside this if statement. Otherwise, I checked all the codes in this PR at least once.

autoPyTorch/api/base_task.py
franchuterivera: Yes exactly, it is just an ordinal number to track how many machine learning algorithms we have launched, so this is the expectation:
- Num_run 0 is the dummy prediction, in case all run fails, to be at least as good as random prediction.
- Num_run 1-7 (depends on the number of traditional pipelines) should be traditional pipelines
- then all remaining runs are neural networks.

I will add this to #127

autoPyTorch/api/base_task.py
franchuterivera: yes exactly, and they changed because there was a bug in that two machine learning algorithms previously had the same number. So two algorithm where written to the same files in disk,

nabenabe0928 (approved):

Issue #143, opened 2021-03-19
[minor issue] mypy error in base_task.py 


nabenabe0928:
From what I saw, this issue is caused by the type of `self.dataset_name`.
Currently, the type of `self.dataset_name` is `Optional[str]`, but since each function defines the argument as `str`, it causes the problem.
It is urgent issue, so I will send a PR and I will ask you to approve this PR quickly.

```
autoPyTorch/api/base_task.py:788: error: Argument 1 to "_get_logger" of "BaseTask" has incompatible type "Optional[str]"; expected "str"  [arg-type]
autoPyTorch/api/base_task.py:864: error: Argument 1 to "wall_elapsed" of "StopWatch" has incompatible type "Optional[str]"; expected "str"  [arg-type]
autoPyTorch/api/base_task.py:876: error: Argument 1 to "wall_elapsed" of "StopWatch" has incompatible type "Optional[str]"; expected "str"  [arg-type]
autoPyTorch/api/base_task.py:896: error: Argument "dataset_name" to "EnsembleBuilderManager" has incompatible type "Optional[str]"; expected "str"  [arg-type]
autoPyTorch/api/base_task.py:927: error: Argument "dataset_name" to "AutoMLSMBO" has incompatible type "Optional[str]"; expected "str"  [arg-type]
```

ravinkohli:
Hey @nabenabe0928 could you tell me the steps to reproduce this error. I tried running mypy check on the refactor_development but did not get it.

nabenabe0928:
1. Checkout `refactor_development` branch
2. Add the code below to the end of `base_task.py`
3. Run `mypy autoPytorch/api/base_task.py`

```
if __name__=='__main__':
    import numpy as np

    from autoPyTorch.api.base_task import BaseTask
    from autoPyTorch.datasets.base_dataset import BaseDataset

    bd = BaseDataset(train_tensors=np.arange(100))
    bt = BaseTask()
    bt._search(optimize_metric='foo', dataset=bd)
```

PR #144, opened 2021-03-19
[fix] Fix the mypy issue in base_task.py

nabenabe0928:
I addressed the [issue#143](https://github.com/automl/Auto-PyTorch/issues/143).
Since it is urgent, please approve it as soon as possible.

### DIFF
--- a/autoPyTorch/datasets/base_dataset.py
+++ b/autoPyTorch/datasets/base_dataset.py
@@ -106,8 +106,12 @@ class BaseDataset(Dataset, metaclass=ABCMeta):
             val_transforms (Optional[torchvision.transforms.Compose]):
                 Additional Transforms to be applied to the validation/test data
         """
-        self.dataset_name = dataset_name if dataset_name is not None \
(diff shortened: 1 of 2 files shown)


autoPyTorch/datasets/base_dataset.py
```diff
@@ -106,8 +106,14 @@ def __init__(
             val_transforms (Optional[torchvision.transforms.Compose]):
                 Additional Transforms to be applied to the validation/test data
         """
-        self.dataset_name = dataset_name if dataset_name is not None \
-            else hash_array_or_matrix(train_tensors[0])
+
+        # Explicit definition for mypy compatibility
+        self.dataset_name: str = ""
```
ArlindKadra, line 111: Is there no other way to pass the mypy check here? 
Overall the variable declaration is not used at all and the variable is always reassigned. (Not sure how much it can be applied, but there can also be inline mypy check skips.)

ArlindKadra (approved):

PR #146, opened 2021-03-23
Hyperparameter Search Space updates now with constant and include ability

### LINKED ISSUE #115: Ability to make hyperparameters constant
For some experiments, we may want to set certain hyperparameters to be constant. This currently can be done if we want to include components for ex, ` include={'trainer': ['StandardTrainer']}` or for categorical hyperparameters by simply sending a single value into the value_range parameter of an update, for ex, ```HyperparameterSearchSpaceUpdate(node_name="network_backbone",
                   hyperparameter="ShapedResNetBackbone:resnet_shape",
                   value_range=['brick'],
                   default_value='brick')```. 

However, for integer or float hyperparameter, ConfigSpace gives an error if we want to set the lower and upper to the same value. This can be done either via another parameter such as 'make_constant' in the HyperparameterSearchSpaceUpdate or as decided earlier we can encapsulate ConfigSpace by introducing a search_space module that can handle such an update elegently

### LINKED ISSUE #125: The enhancement in the arguments of HyperparameterSearchSpaceUpdates
Currently, when we create configuration space, we often take `var_name=((lower_bound, upper_bound), default_value)` as an argument. However, since `var_name` is already used as a value of the variable, this is not `Tuple[Tuple[type, type], type]`.
For example, while we use `lr` as a float in some functions, we take `lr` as a tuple in the function to create configuration space.
Meta-programming is not intuitive and not easy to understand. We should unite the name for consistency.

ravinkohli:
Closes #125 - Introduces a new class HyperparameterSearchSpace 
Closes #115  
HyperparameterSearchSpaceUpdates can now also update component search space in autoPyTorchChoice modules by passing __choice__ hyperparameters in the update.


autoPyTorch/pipeline/components/base_choice.py
```diff
@@ -247,35 +249,37 @@ def _check_dataset_properties(self, dataset_properties: Dict[str, Any]) -> None:
         """
         assert isinstance(dataset_properties, dict), "dataset_properties must be a dictionary"
 
-    def _apply_search_space_update(self, name: str, new_value_range: Union[List, Tuple],
-                                   default_value: Union[int, float, str], log: bool = False) -> None:
+    def _apply_search_space_update(self, hyperparameter_search_space_update: HyperparameterSearchSpaceUpdate) -> None:
         """Allows the user to update a hyperparameter
```
franchuterivera, line 253: I think this doc-string can be improved.. It really updates the hyperparameter search space I think.

autoPyTorch/pipeline/components/base_choice.py
```diff
@@ -247,35 +249,37 @@ def _check_dataset_properties(self, dataset_properties: Dict[str, Any]) -> None:
...
 
-        Arguments:
+        Args:
             name {string} -- name of hyperparameter
```
franchuterivera, line 256: And this could use a parenthesis rather than curly brackets

autoPyTorch/pipeline/components/base_choice.py
```diff
@@ -247,35 +249,37 @@ def _check_dataset_properties(self, dataset_properties: Dict[str, Any]) -> None:
...
-        self._cs_updates[name] = tuple([new_value_range, default_value, log])
+        self._cs_updates[hyperparameter_search_space_update.hyperparameter] = hyperparameter_search_space_update
 
-    def _get_search_space_updates(self, prefix: Optional[str] = None) -> Dict[str, Tuple]:
+    def _get_search_space_updates(self, prefix: Optional[str] = None) -> Dict[str, HyperparameterSearchSpace]:
         """Get the search space updates with the given prefix
 
-        Keyword Arguments:
+        Args:
             prefix {str} -- Only return search space updates with given prefix (default: {None})
```
franchuterivera, line 267: same here, update to parenthesis

autoPyTorch/pipeline/components/base_choice.py
```diff
@@ -247,35 +249,37 @@ def _check_dataset_properties(self, dataset_properties: Dict[str, Any]) -> None:
...
 
         Returns:
             dict -- Mapping of search space updates. Keys don't contain the prefix.
         """
+        RETURN_ALL = False
         if prefix is None:
-            return self._cs_updates
-        result: Dict[str, Tuple] = dict()
+            RETURN_ALL = True
```
franchuterivera, line 274: Why not directly ask for prefix is None here rather than having RETURN_ALL?

autoPyTorch/pipeline/components/base_choice.py
```diff
@@ -247,35 +249,37 @@ def _check_dataset_properties(self, dataset_properties: Dict[str, Any]) -> None:
...
+        result: Dict[str, HyperparameterSearchSpace] = dict()
 
-        # iterate over all search space updates of this node and filter the ones out, that have the given prefix
+        # iterate over all search space updates of this node and keep the ones that have the given prefix
         for key in self._cs_updates.keys():
-            if key.startswith(prefix):
-                result[key[len(prefix) + 1:]] = self._cs_updates[key]
+            if RETURN_ALL:
+                result[key] = self._cs_updates[key].get_search_space()
+            elif re.search(f'^{prefix}', key) is not None:
```
franchuterivera, line 282: Actually, I think you would not need the assert if you do the change of `if prefix is None` in this for loop

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/feature_preprocessing/RandomKitchenSinks.py
```diff
@@ -35,35 +36,38 @@ def fit(self, X: Dict[str, Any], y: Any = None) -> BaseEstimator:
...
-        gamma: Tuple[Tuple, float, bool] = ((3.0517578125e-05, 8), 1.0, True),
-        degree: Tuple[Tuple, int] = ((2, 5), 3),
-        coef0: Tuple[Tuple, float] = ((-1, 1), 0)
+        n_components: HyperparameterSearchSpace = HyperparameterSearchSpace(hyperparameter='n_components',
+                                                                            value_range=(0.5, 0.9),
+                                                                            default_value=0.5,
+                                                                            ),
+        gamma: HyperparameterSearchSpace = HyperparameterSearchSpace(hyperparameter='gamma',
+                                                                     value_range=(3.0517578125e-05, 8),
+                                                                     default_value=0.01,
```
franchuterivera, line 45: Shouldn't the default be 1.0?

test/test_pipeline/components/setup/test_setup_networks.py
```diff
@@ -47,8 +47,6 @@ def test_pipeline_fit(self, fit_dictionary_tabular, embedding, backbone, head):
         assert backbone == config.get('network_backbone:__choice__', None)
         assert head == config.get('network_head:__choice__', None)
         pipeline.set_hyperparameters(config)
-        # Need more epochs to make sure validation performance is met
-        fit_dictionary_tabular['epochs'] = 100
```
franchuterivera, line 51: Why do you want to change this?

autoPyTorch/pipeline/tabular_classification.py
```diff
@@ -196,9 +196,8 @@ def _get_hyperparameter_search_space(self,
             embeddings = cs.get_hyperparameter('network_embedding:__choice__').choices
             if 'LearnedEntityEmbedding' in embeddings:
                 encoders = cs.get_hyperparameter('encoder:__choice__').choices
-                default = cs.get_hyperparameter('network_embedding:__choice__').default_value
                 possible_default_embeddings = copy.copy(list(embeddings))
-                del possible_default_embeddings[possible_default_embeddings.index(default)]
+                del possible_default_embeddings[possible_default_embeddings.index('LearnedEntityEmbedding')]
```
franchuterivera, line 200: Can you change the comment before that says:
`Here we add custom code, like this with this`

autoPyTorch/pipeline/components/setup/network_head/fully_convolutional.py
```diff
@@ -70,32 +71,41 @@ def get_properties(dataset_properties: Optional[Dict[str, Any]] = None) -> Dict[
...
 
-        pooling_method_hp = CategoricalHyperparameter("pooling_method",
-                                                      choices=["average", "max"])
+        add_hyperparameter(cs, pooling_method, CategoricalHyperparameter)
 
-        activation_hp = CategoricalHyperparameter('activation',
-                                                  choices=list(_activations.keys()))
+        activation_hp = get_hyperparameter(activation, CategoricalHyperparameter)
 
-        cs.add_hyperparameters([num_layers_hp, pooling_method_hp, activation_hp])
```
franchuterivera, line 90: this pooling_method_hp was extra?

test/test_pipeline/test_tabular_classification.py
```diff
@@ -330,4 +330,85 @@ def test_set_range_search_space_updates(self, fit_dictionary_tabular):
...
+                                                 search_space_updates=updates)
+        self._assert_pipeline_search_space(pipeline, updates)
+
+
+@pytest.mark.parametrize("fit_dictionary_tabular", ['iris'], indirect=True)
+def test_constant_pipeline_iris(fit_dictionary_tabular):
+    search_space_updates = HyperparameterSearchSpaceUpdates()
+    search_space_updates.append(node_name='network_backbone',
+                                hyperparameter='__choice__',
+                                value_range=['MLPBackbone'],
```
franchuterivera, line 364: Can you also add shaped mlp backbone to make sure a list works?

test/test_pipeline/test_tabular_classification.py
```diff
@@ -330,4 +330,85 @@ def test_set_range_search_space_updates(self, fit_dictionary_tabular):
...
+            updates.append(node_name=name[0], hyperparameter=hyperparameter_name,
+                           value_range=value_range, default_value=default_value)
+        pipeline = TabularClassificationPipeline(dataset_properties=dataset_properties,
+                                                 search_space_updates=updates)
+        self._assert_pipeline_search_space(pipeline, updates)
+
+
+@pytest.mark.parametrize("fit_dictionary_tabular", ['iris'], indirect=True)
+def test_constant_pipeline_iris(fit_dictionary_tabular):
+    search_space_updates = HyperparameterSearchSpaceUpdates()
```
franchuterivera, line 361: can you also fix the preprocessing to a certain value? for example TruncatedSVD and standard scaler?

test/test_pipeline/test_tabular_classification.py
```diff
@@ -330,4 +330,85 @@ def test_set_range_search_space_updates(self, fit_dictionary_tabular):
...
+    assert run_summary.total_parameter_count > 0
+    assert 'accuracy' in run_summary.performance_tracker['train_metrics'][1]
+
+    # Make sure default pipeline achieves a good score for dummy datasets
+    epoch2loss = run_summary.performance_tracker['val_loss']
+    best_loss = min(list(epoch2loss.values()))
+    epoch_where_best = list(epoch2loss.keys())[list(epoch2loss.values()).index(best_loss)]
+    score = run_summary.performance_tracker['val_metrics'][epoch_where_best]['accuracy']
+
+    assert score >= 0.9, run_summary.performance_tracker['val_metrics']
```
franchuterivera, line 414: Can you also check the train performance to be better than 0.9? Just to make sure that train is also recorded to track overfitting

test/test_pipeline/test_tabular_classification.py
```diff
@@ -330,4 +330,85 @@ def test_set_range_search_space_updates(self, fit_dictionary_tabular):
...
+                                default_value='NoScheduler')
+    search_space_updates.append(node_name='optimizer',
+                                hyperparameter='__choice__',
+                                value_range=['AdamOptimizer'],
+                                default_value='AdamOptimizer')
+    search_space_updates.append(node_name='optimizer',
+                                hyperparameter='AdamOptimizer:lr',
+                                value_range=[1e-2],
+                                default_value=1e-2)
+    pipeline = TabularClassificationPipeline(dataset_properties=fit_dictionary_tabular['dataset_properties'],
```
franchuterivera, line 390: Can you train the performance metric to be something different, to make sure we do not hardcode accuracy somewhere? Could simply be balanced accuracy

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/scaling/base_scaler_choice.py
```diff
@@ -70,20 +70,36 @@ def get_hyperparameter_search_space(self,
...
+        if '__choice__' in updates.keys():
+            choice_hyperparameter = updates['__choice__']
+            if not set(choice_hyperparameter.value_range).issubset(available_scalers):
+                raise ValueError("Expected given update for {} to have "
+                                 "choices in {} got {}".format(self.__class__.__name__,
+                                                               available_scalers,
+                                                               choice_hyperparameter.value_range))
+            if len(dataset_properties['numerical_columns']) == 0:
+                assert len(choice_hyperparameter.value_range) == 1
+                assert 'NoScaler' in choice_hyperparameter.value_range, \
```
franchuterivera, line 83: maybe is better to change this to be a valueerror

autoPyTorch/pipeline/components/base_choice.py
ravinkohli: yeah I dont know why it is curly :P I'll fix this

autoPyTorch/pipeline/components/base_choice.py
ravinkohli: same

autoPyTorch/pipeline/components/base_choice.py
ravinkohli: I dont know but I'll do it directly now

test/test_pipeline/test_tabular_classification.py
```diff
@@ -330,4 +330,85 @@ def test_set_range_search_space_updates(self, fit_dictionary_tabular):
...
+                                             search_space_updates=search_space_updates)
+
+    try:
+        pipeline.fit(fit_dictionary_tabular)
+    except Exception as e:
+        pytest.fail(f"Failed due to {e}")
+
```
franchuterivera, line 397: Can you also make sure that the pipeline.config` matches the above expectation? like adam should be the optimizer and so on...

test/test_pipeline/components/setup/test_setup_networks.py
ravinkohli: because now the fit dictionary already has 100 for the epochs value. See this https://github.com/ravinkohli/Auto-PyTorch/blob/ad43a947a5447c75b4ccd6bd44335fb96aa226d8/test/conftest.py#L290

autoPyTorch/pipeline/components/setup/network_embedding/base_network_embedding_choice.py
```diff
@@ -146,26 +146,42 @@ def get_hyperparameter_search_space(
...
+        if '__choice__' in updates.keys():
+            choice_hyperparameter = updates['__choice__']
+            if not set(choice_hyperparameter.value_range).issubset(available_embedding):
+                raise ValueError("Expected given update for {} to have "
+                                 "choices in {} got {}".format(self.__class__.__name__,
+                                                               available_embedding,
+                                                               choice_hyperparameter.value_range))
+            if len(dataset_properties['categorical_columns']) == 0:
+                assert len(choice_hyperparameter.value_range) == 1
+                assert 'NoEmbedding' in choice_hyperparameter.value_range, \
```
franchuterivera, line 166: Here as well, maybe a valueerror is better

autoPyTorch/pipeline/components/setup/network_head/fully_convolutional.py
ravinkohli: no it has been added in line 94. For hyperparameters that have a condition on them, we can't directly add them to the config space, we need the variable for that hyperparameter and that's why num_layers_hp and activation_hp are explicitly added

autoPyTorch/utils/common.py
```diff
@@ -35,6 +46,25 @@ def __str__(self) -> str:
             self.name, self.supported_types, self.user_defined, self.dataset_property)
 
 
+class HyperparameterSearchSpace(NamedTuple):
+    """
+    A class that holds the search space for an individual hyperparameter.
```
franchuterivera, line 51: should it be search space updates?

test/test_pipeline/test_tabular_classification.py
ravinkohli: yes, sure

autoPyTorch/utils/common.py
```diff
@@ -35,6 +46,25 @@ def __str__(self) -> str:
...
+    Attributes:
+        hyperparameter (str):
+            name of the hyperparameter
+        value_range (Sequence[HyperparameterValueType]):
+            range of the hyperparameter
```
franchuterivera, line 56: Maybe here clarify that it can be a rangle between max-min or a range of possible choices

test/test_pipeline/test_tabular_classification.py
ravinkohli: yes sure.

test/test_pipeline/test_tabular_classification.py
ravinkohli: okay

test/test_pipeline/test_tabular_classification.py
ravinkohli: yes, i'll do it for balanced then. Good catch

test/test_pipeline/test_tabular_classification.py
ravinkohli: yes sure

autoPyTorch/utils/common.py
```diff
@@ -156,3 +186,67 @@ def subsampler(data: Union[np.ndarray, pd.DataFrame, scipy.sparse.csr_matrix],
                x: Union[np.ndarray, List[int]]
                ) -> Union[np.ndarray, pd.DataFrame, scipy.sparse.csr_matrix]:
     return data[x] if isinstance(data, (np.ndarray, scipy.sparse.csr_matrix)) else data.iloc[x]
+
+
+def get_hyperparameter(hyperparameter: HyperparameterSearchSpace,
```
franchuterivera, line 191: Can we add a very simple unit testing that has a configuration space like this:
```
X = Integer, from -5, 5
Y = Float from 0 to 1
X = categoritcal ['a', 'b', 1] 
```

And make sure that this work?

franchuterivera (changes requested):
Thanks for the changes, I think they will help people customize autopytorch as needed, for different configuration spaces. 

I just left some few comments, and I have one additional request, that is to create a unit testing for the hyperparameter search space update apart on an appropriated `test/test_utils` directory with a dummy configuration space to make sure that code is robust and with good coverage.

autoPyTorch/utils/common.py
ravinkohli: I didnt use the word update as this is used as a default also. Now all hyperparameter search spaces need to be defined this way and when we pass an update, the update class creates this search space and passes it to the get_hyperparameter_search_space method. see here https://github.com/ravinkohli/Auto-PyTorch/blob/ad43a947a5447c75b4ccd6bd44335fb96aa226d8/autoPyTorch/utils/hyperparameter_search_space_update.py#L57. What I am confused about is that, as we are calling get_hyperparameter_search_space for the whole hyperparameter search space of the component maybe it can be confusing to use HyperparameterSearchSpace also for individual hyperparameters. What do you think?

autoPyTorch/pipeline/components/base_choice.py
ravinkohli: True

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/feature_preprocessing/RandomKitchenSinks.py
ravinkohli: yes, thanks for catching this. I must have messed up while changing

test/test_pipeline/test_tabular_classification.py
ravinkohli: I am adding PolynomialFeatures as TruncatedSVD got 84% accuracy

franchuterivera (approved):

PR #147, opened 2021-03-24
[Bug] Fix random halt problems on traditional pipelines

franchuterivera:
When saving a traditional classification pipeline, the fit dictionary is being saved alongside the pipeline. This means that the X_train and y_train are saved also, causing a memory pick that halts the run. This PR aims to fix this problem


ravinkohli (approved):

ArlindKadra (approved):
So, we were basically giving the wrong argument initially to `init_params`  in case `configuration` was not a Configuration object.
On a related note, what would `init_params` constitute of?

franchuterivera:
I have added a new push that adds documentation to answer the question of what is `init_params`.

And the biggest problem was that every pipeline was being saved with the fit_dictionary. The fit dictionary includes a copy of the dataset, which was a memory overkill.

PR #155, opened 2021-03-26
Adding the implementation for SGD with decoupled weight decay

ArlindKadra:
I was a bit surprised, but I actually did not see anywhere the implementation for SGD with decoupled weight decay. If it was, you can just reject this PR.


ravinkohli:
I think this PR is worth relooking. @ArlindKadra would you like to update it so we can merge it with development? I can also take ownership.

Issue #140, opened 2021-03-16
Many Dicts in the codes are not understandable.


nabenabe0928:
As in the title, many dicts in the codes are not understandable.

For example, what does `named_steps` in base_pipeline.py mean?

To address this problem, I would like to promote the usage of either [BaseNamedTuple](https://github.com/automl/Auto-PyTorch/pull/139) or [BaseDict](https://github.com/automl/Auto-PyTorch/pull/138).

I felt the following problems regarding the current usage of dicts:
1. When we use a dict in various locations, it is not clear what variables should be included in the dict (or tuple).
2. The definition of each variable is not clear.
3. It is likely to cause a lot of typo, because we cannot have the benefit of the autocompletion functionality from editors (e.g. vscode).

The new rules for this usage are the followings:
1. Define BaseDict or BaseNamedTuple in the head of each file if we need to use ad-hoc dicts
2. Write the locations (which files) of doc-string for each dict or NamedTuple for better understandings
3. Use attribute and avoid string keys (i.e. use `d.foo` instead of `d["foo"]`)

About the last statement (use attributes), that is because it is easier to prevent typo by the autocompletion.

For these reasons, I have introduced `BaseDict` and `BaseNamedTuple`.
Basically, when we would like to have variables with immutable state, we use `BaseNamedTuple` and otherwise we use `BaseDict`.

Both modules can be used by `from autoPytorch.common import create_dictlike_namedtuple, BaseDict`.

The declaration for each module is the following:
```
class menu(NamedTuple):
     fish: int = 5
     meat: int = 10
     drink: int = 7

today = create_dictlike_namedtuple(menu, fish=3, meat=13)
for key, value in today.items():
     print(key, value)
print(today)

[out]
fish 3
meat 13
drink 7
WrapedNamedTuple(fish=3, meat=13, drink=7)
```

```
class NewDict(BaseDict):
     a: int = 3
     b: float = 2.0
     c: str

new_dict = NewDict(a=1, d=5)
print(new_dict)
print(new_dict.a, new_dict.b, new_dict.c, new_dict.d)

[out]        
{'a': 1, 'd': 5, 'b': 2.0, 'c': None}
1 2.0 None 5
```

nabenabe0928:
I found out [dataclasses](https://docs.python.org/3/library/dataclasses.html) are the good solution for this.

I will replace most `NamedTuple` and `dict` with `dataclasses` in the near future.

ravinkohli:
[...]

PR #164, opened 2021-04-09
[refactor] Getting dataset properties from the dataset object

### LINKED ISSUE #152: [refactor] Getting the dataset properties directly from the dataset class
Currently we have 2 functions `get_dataset_requirements` and then `dataset.get_required_properties` (see [here](https://github.com/automl/Auto-PyTorch/blob/db70a436cc8044a7bc03ad8846012bb3651eaca5/autoPyTorch/api/base_task.py#L781)). We can unify them into a single function that will directly return all the required properties from the dataset. This would be especially useful in the [abstract_evaluator](https://github.com/automl/Auto-PyTorch/blob/db70a436cc8044a7bc03ad8846012bb3651eaca5/autoPyTorch/evaluation/abstract_evaluator.py#L302) as it can reduce the clutter [here](https://github.com/automl/Auto-PyTorch/blob/db70a436cc8044a7bc03ad8846012bb3651eaca5/autoPyTorch/evaluation/abstract_evaluator.py#L298)

ravinkohli:
Currently, we have multiple copies of the required dataset info. For example check in the [API](https://github.com/automl/Auto-PyTorch/blob/68fc77f6f1c63d963ae2198e6a95ea8503da09fa/autoPyTorch/api/tabular_classification.py#L100) in the [BaseDataset](https://github.com/automl/Auto-PyTorch/blob/68fc77f6f1c63d963ae2198e6a95ea8503da09fa/autoPyTorch/datasets/base_dataset.py#L369) and in the [AbstractEvaluator](https://github.com/automl/Auto-PyTorch/blob/68fc77f6f1c63d963ae2198e6a95ea8503da09fa/autoPyTorch/evaluation/abstract_evaluator.py#L347).

In this PR, only the required dataset info from the BaseDataset is used, replacing all the other hard coded dictionaries defined. 

It also fixes a bug in AbstractEvaluator, where we were creating a DummyClassificationPipeline instead of a DummyRegressionPipeline. [here](https://github.com/automl/Auto-PyTorch/blob/68fc77f6f1c63d963ae2198e6a95ea8503da09fa/autoPyTorch/evaluation/abstract_evaluator.py#L319)

It also fixes a bug where the pipeline updates were not honored while getting the dataset requirements. For example in BaseTask [here](https://github.com/automl/Auto-PyTorch/blob/fae72a4de79b747918076292b0540a8e4602cb78/autoPyTorch/api/base_task.py#L268) and [here](https://github.com/automl/Auto-PyTorch/blob/fae72a4de79b747918076292b0540a8e4602cb78/autoPyTorch/api/base_task.py#L785). In AbstractEvaluator [here](https://github.com/automl/Auto-PyTorch/blob/fae72a4de79b747918076292b0540a8e4602cb78/autoPyTorch/evaluation/abstract_evaluator.py#L363) and finally the method in utils/pipeline [here](https://github.com/automl/Auto-PyTorch/blob/fae72a4de79b747918076292b0540a8e4602cb78/autoPyTorch/utils/pipeline.py#L25)

Closes #152

### DIFF
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -197,14 +197,6 @@ class BaseTask:
                                  " HyperparameterSearchSpaceUpdates got {}".format(type(self.search_space_updates)))
 
     @abstractmethod
-    def _get_required_dataset_properties(self, dataset: BaseDataset) -> Dict[str, Any]:
-        """
-        given a pipeline type, this function returns the
-        dataset properties required by the dataset object
-        """
-        raise NotImplementedError
-
-    @abstractmethod
     def build_pipeline(self, dataset_properties: Dict[str, Any]) -> BasePipeline:
         """
         Build pipeline according to current task
@@ -267,7 +259,7 @@ class BaseTask:
             return self.search_space
         elif dataset is not None:
             dataset_requirements = get_dataset_requirements(
-                info=self._get_required_dataset_properties(dataset))
+                info=dataset.get_required_dataset_info())
             return get_configuration_space(info=dataset.get_dataset_properties(dataset_requirements),
                                            include=self.include_components,
                                            exclude=self.exclude_components,
@@ -785,7 +777,7 @@ class BaseTask:
         # Initialise information needed for the experiment
         experiment_task_name: str = 'runSearch'
         dataset_requirements = get_dataset_requirements(
-            info=self._get_required_dataset_properties(dataset))
+            info=dataset.get_required_dataset_info())
         self._dataset_requirements = dataset_requirements
         dataset_properties = dataset.get_dataset_properties(dataset_requirements)
         self._stopwatch.start_task(experiment_task_name)
@@ -1049,7 +1041,7 @@ class BaseTask:
             self._logger = self._get_logger(str(self.dataset_name))
 
         dataset_requirements = get_dataset_requirements(
-            info=self._get_required_dataset_properties(dataset))
+            info=dataset.get_required_dataset_info())
         dataset_properties = dataset.get_dataset_properties(dataset_requirements)
         self._backend.save_datamanager(dataset)
 
@@ -1119,7 +1111,7 @@ class BaseTask:
 
         # get dataset properties
         dataset_requirements = get_dataset_requirements(
-            info=self._get_required_dataset_properties(dataset))
+            info=dataset.get_required_dataset_info())
         dataset_properties = dataset.get_dataset_properties(dataset_requirements)
         self._backend.save_datamanager(dataset)
 
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -259,7 +259,10 @@ class BaseTask:
             return self.search_space
         elif dataset is not None:
             dataset_requirements = get_dataset_requirements(
-                info=dataset.get_required_dataset_info())
+                info=dataset.get_required_dataset_info(),
+                include=self.include_components,
+                exclude=self.exclude_components,
+                search_space_updates=self.search_space_updates)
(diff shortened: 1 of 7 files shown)


franchuterivera (approved):

autoPyTorch/api/base_task.py
```diff
@@ -196,14 +196,6 @@ def __init__(
                 raise ValueError("Expected search space updates to be of instance"
                                  " HyperparameterSearchSpaceUpdates got {}".format(type(self.search_space_updates)))
 
-    @abstractmethod
-    def _get_required_dataset_properties(self, dataset: BaseDataset) -> Dict[str, Any]:
```
nabenabe0928, line 200: it would be nice if we have `DatasetInfo` object to refer to the information we should include.
Btw, which name would you prefer, either `dataset_properties` or `required_dataset_info`?

ravinkohli: I know it would be useful to have these objects but I don't see them as a priority. As discussed in the meeting on Wednesday, we should aim to get a new release by next week. Therefore, for the old code that's already been put in, I feel we can start refactoring these into `AttrDict` objects once we are done with the reproducibility results. We can make sure any new code we put, is made with such `AttrDict` that makes the code easier to understand but changing old code should not be our priority right now. I suggest you make an issue for them or as we already have this issue  #140, you can add any problem we have as a comment there.

nabenabe0928: Sorry, I forgot responding you for such a long time!!!
I did not say you have to put AttrDict, but can you put NamedTuple, right?
Anyways, the contents are not clear to me especially the exact key names.

nabenabe0928 (changes requested):

autoPyTorch/api/base_task.py
ravinkohli: I may have mentioned only AttrDict but my comment was for both. Although I would like to clarify what exactly is the issue, is it about the dataset properties that are passed into the fit dictionary? Or the data set info needed to instantiate pipeline. In terms of knowing the exact key names, this is something the user or even we do not have to directly interact with or change the key names. The issue I have with making a AttrDict or namedTuple, is we can not really define what would be needed in a pipeline. For different pipelines, there would be different dataset properties based on the components that are active in it. I can enhance the documentation to explain these terms better, but in its current form, we need it to be a mutable object.

nabenabe0928 (approved):

autoPyTorch/utils/pipeline.py
```diff
@@ -23,75 +22,61 @@
 
 
 def get_dataset_requirements(info: Dict[str, Any],
-                             include_estimators: Optional[List[str]] = None,
-                             exclude_estimators: Optional[List[str]] = None,
-                             include_preprocessors: Optional[List[str]] = None,
-                             exclude_preprocessors: Optional[List[str]] = None
+                             include: Optional[Dict] = None,
```
nabenabe0928, line 25: What kind of change is it?
Could you describe what changed from the previous parameters?

ravinkohli: The previous parameters are not compatible or needed for autoPyTorch, I think we left it unchanged from when we used the code from autosklearn. In our case, we need to pass 4 things to instantiate the pipeline
1. Dataset properties
2. include
3. exclude
4. search space updates.

nabenabe0928: Is it possible to change the name include or exclude to include_something and exclude_something?

Plus I did not get from what we will include or exclude the preprocessor in the previous code. So it will be helpful if you could write the description for them.

ravinkohli: I'll add a doc string but include and exclude is also how its called in our pipelines, So to change that I suggest you to raise an issue and we can address in a different PR.

Issue #148, opened 2021-03-25
[refactor] `_search` function in autoPytorch/api/base_task.py


nabenabe0928:
Currently, comments complement the information what we are doing inside the code, but we can split these sections into functions.
In fact, too long function typically violates flake8 C901.
Also you can check why it is not good to write a long function [here](https://en.wikipedia.org/wiki/Cyclomatic_complexity).

Currently, the code looks like this:
```
def _search(self, ...):
    # Initialise information needed for the experiment
    process 1

    # Print debug information to log
    process 2

    # Handle time resource allocation
    process 3

    # Make sure that at least 2 models are created for the ensemble process
    process 4

    # ============> Run dummy predictions
    process 5

    # ============> Run traditional ml
    process 6

    # ============> Starting ensemble
    process 7

    # ==> Run SMAC
    process 8

    post process
    
    return results
```

However, we can split like this (I do not think we need to split everything):
```
def _init_experiment_info(self, ...):
    process 1
    return results

def _run_dummy_predictions(self, ...):
    process 5
    return results

def _run_traditional_ml(self, ...):
    process 6
    return results

def _start_ensemble(self, ...):
    process 7
    return results

def _run_smac(self, ...):
    process 8
    return results

def _post_processing(self, ...):
    post processing 
    return results

def _search(self, ...):
    self._init_experiment_info(...):

    # Print debug information to log
    process 2

    # Handle time resource allocation
    process 3

    # Make sure that at least 2 models are created for the ensemble process
    process 4

    self._run_dummy_predictions(...)
    self._run_traditional_ml(...)
    self._start_ensemble(...)
    self._run_smac(...)
    self._post_processing(...)

```

nabenabe0928:
Suggetions
1. make each function shorter than 40 lines excluding doc-string
2. prohibit the following indent level more than 3

```
^
| 40 lines are visible in one screen
|
class foo():
    def func(self):
        indent level 0 (normal)
            indent level 1 (normal)
                indent level 2 (can happen)
                    indent level 3 (much -> think about splitting the function)
                        indent level 4 (too much -> should be prohibited)
|
|
v  
```

ravinkohli:
would you like to resolve this issue @nabenabe0928

PR #165, opened 2021-04-12
[refactor] Address the Issue#148 and clean _search() function

nabenabe0928:
For more details, see #148.


autoPyTorch/api/base_task.py
```diff
@@ -684,6 +685,269 @@ def _do_traditional_prediction(self, time_left: int, func_eval_time_limit_secs:
...
         return
 
+    def _run_dummy_predictions(self) -> None:
+        dummy_task_name = 'runDummy'
+        self._stopwatch.start_task(dummy_task_name)
+        self._do_dummy_prediction()
+        self._stopwatch.stop_task(dummy_task_name)
+
+    def _run_traditional_ml(self) -> None:
+        """We would like to obtain training time for at least 1 Neural network in SMAC"""
```
ravinkohli, line 695: could you move this comment to line 706?

autoPyTorch/api/base_task.py
```diff
@@ -684,6 +685,269 @@ def _do_traditional_prediction(self, time_left: int, func_eval_time_limit_secs:
...
+        budget_config: Dict[str, Union[float, str]] = {}
+        if budget_type is not None and budget is not None:
+            budget_config['budget_type'] = budget_type
+            budget_config[budget_type] = budget
+        elif budget_type is not None or budget is not None:
+            raise ValueError("budget type was not specified in budget_config")
+
+        return budget_config
+
+    def _start_smac(self, proc_smac: AutoMLSMBO) -> None:
```
ravinkohli, line 767: I think its fine if we have this function inside _run_smac. Having two functions makes it a bit confusing.

autoPyTorch/api/base_task.py
```diff
@@ -170,13 +170,13 @@ def __init__(
             os.path.join(os.path.dirname(__file__), '../configs/default_pipeline_options.json'))))
 
         self.search_space: Optional[ConfigurationSpace] = None
-        self._dataset_requirements: Optional[List[FitRequirement]] = None
         self._metric: Optional[autoPyTorchMetric] = None
         self._logger: Optional[PicklableClientLogger] = None
         self.run_history: RunHistory = RunHistory()
         self.trajectory: Optional[List] = None
-        self.dataset_name: Optional[str] = None
+        self.dataset_name: str = ""
```
ravinkohli, line 177: why do you want to have it as an empty string?

ravinkohli (changes requested):
Hey, Thanks for the PR. While I agree with dividing _search into smaller functions. Do you think too many functions have been created now? I feel if you could merge the functions that are doing one thing for example  _do_traditional_predictions can be merged into _run_traditional_ml and same for dummy, and smac it would make it much easier to understand whats going on in the code. And while you are making these functions, it would be great to also add documentation for each one. as well as keeping the comments that were there before each process for example `=====> Run dummy`, etc. Also, I think its better if we keep the _search_settings a part of the search function.

autoPyTorch/api/base_task.py
ravinkohli: Okay, I see that it was suggested by Francisco in a previous PR. I suggest you to not make the same changes in different PR. Let merge #106 first and then we can rebase to include this change. It makes it confusing to review the same changes in different places.

PR #166, opened 2021-04-12
[refactor] Separate processing inside BaseTask._search() function

nabenabe0928:
For more details, see #148.


nabenabe0928 (commented):
This post is a mistake

nabenabe0928:
I do not know why but there are double submissions, so I close this PR.

PR #170, opened 2021-04-14
[Feature] Get incumbent config and results

### LINKED ISSUE #150: Return the best configuration found by the optimizer
the [`get_incumbent_config`](https://github.com/automl/Auto-PyTorch/blob/db70a436cc8044a7bc03ad8846012bb3651eaca5/autoPyTorch/api/base_task.py#L1240) and `get_incumbent_results` is currently not doing anything. 

Ideally `get_incumbent_config` should search through the run history object and find the best configurations from there and return it. We could also have a flag to enable searching for results from the traditional classifiers. Also, returning the fitted pipeline as an object but this is optional.

ravinkohli:
Fixes #150 

Changes made-
1. removed `get_incumbent_config` from base task as it would be redundant to have two separate functions for getting configuration and the results
2. `get_incumbent_results` goes through the run history and gets the config that has the minimum cost on the validation set.
3. Also returns the additional info which includes optimisation loss on the other metrics specified  and the train loss.
4. Adds test to check if the function is working properly

### DIFF
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -13,7 +13,7 @@ import unittest.mock
 import uuid
 import warnings
 from abc import abstractmethod
-from typing import Any, Callable, Dict, List, Optional, Union, cast
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union, cast
 
 from ConfigSpace.configuration_space import Configuration, ConfigurationSpace
 
@@ -217,9 +217,7 @@ class BaseTask:
         """
         raise NotImplementedError
 
-    def set_pipeline_config(
-            self,
-            **pipeline_config_kwargs: Any) -> None:
+    def set_pipeline_config(self, **pipeline_config_kwargs: Any) -> None:
         """
         Check whether arguments are valid and
         then sets them to the current pipeline
@@ -253,12 +251,6 @@ class BaseTask:
         """
         return self.pipeline_options
 
-    # def set_search_space(self, search_space: ConfigurationSpace) -> None:
-    #     """
-    #     Update the search space.
-    #     """
-    #     raise NotImplementedError
-    #
     def get_search_space(self, dataset: BaseDataset = None) -> ConfigurationSpace:
         """
         Returns the current search space as ConfigurationSpace object.
@@ -396,9 +388,9 @@ class BaseTask:
             None
         """
         if (
-                hasattr(self, '_is_dask_client_internally_created')
-                and self._is_dask_client_internally_created
-                and self._dask_client
+            hasattr(self, '_is_dask_client_internally_created')
+            and self._is_dask_client_internally_created
+            and self._dask_client
         ):
             self._dask_client.shutdown()
             self._dask_client.close()
@@ -657,9 +649,11 @@ class BaseTask:
                             f"Fitting {cls} took {runtime}s, performance:{cost}/{additional_info}")
                         configuration = additional_info['pipeline_configuration']
                         origin = additional_info['configuration_origin']
+                        additional_info.pop('pipeline_configuration')
+                        additional_info.pop('configuration_origin')
                         run_history.add(config=configuration, cost=cost,
                                         time=runtime, status=status, seed=self.seed,
-                                        origin=origin)
+                                        origin=origin, additional_info=additional_info)
                     else:
                         if additional_info.get('exitcode') == -6:
                             self._logger.error(
@@ -1016,10 +1010,10 @@ class BaseTask:
         return self
 
     def refit(
-            self,
-            dataset: BaseDataset,
-            budget_config: Dict[str, Union[int, str]] = {},
-            split_id: int = 0
+        self,
+        dataset: BaseDataset,
+        budget_config: Dict[str, Union[int, str]] = {},
+        split_id: int = 0
     ) -> "BaseTask":
         """
         Refit all models found with fit to new data.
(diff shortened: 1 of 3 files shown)


autoPyTorch/api/base_task.py
```diff
@@ -1245,17 +1239,32 @@ def __del__(self) -> None:
...
+        Returns:
+
+        """
+        assert self.run_history is not None, "No Run History found, search has not been called."
+        assert not self.run_history.empty(), "Run History is empty."
+
+        run_history_data = self.run_history.data
+        if not include_traditional:
+            run_history_data = dict(
+                filter(lambda elem: elem[1].additional_info is not None and 'trainer_configuration' not in elem[1].
```
ArlindKadra, line 1260: Hmm, what does having the key `trainer_configuration `mean in this scenario ? Since when include_traditional is activated, we do not filter the `run_history_data`.

autoPyTorch/api/base_task.py
```diff
@@ -1245,17 +1239,32 @@ def __del__(self) -> None:
         # until the estimator is deleted
         self._backend.context.delete_directories(force=False)
 
-    @typing.no_type_check
     def get_incumbent_results(
-            self
-    ):
-        pass
```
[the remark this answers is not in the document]
ravinkohli: they were not. We could get them by taking the last config in incumbent_trajectory or by manually going through the run history of the estimator

autoPyTorch/api/base_task.py
ravinkohli: so this is an ad-hoc way of checking if the run history belongs to the traditional pipeline or an autopytorch pipeline. I'll add a comment before this to make it clearer

ArlindKadra: Aha, oki, perfect.

autoPyTorch/api/base_task.py
```diff
@@ -657,9 +649,11 @@ def _do_traditional_prediction(self, num_run: int, time_left: int, func_eval_tim
                             f"Fitting {cls} took {runtime}s, performance:{cost}/{additional_info}")
                         configuration = additional_info['pipeline_configuration']
                         origin = additional_info['configuration_origin']
+                        additional_info.pop('pipeline_configuration')
```
franchuterivera, line 652: Sorry, why are you doing pop here?

autoPyTorch/api/base_task.py
```diff
@@ -1245,17 +1239,32 @@ def __del__(self) -> None:
...
-    @typing.no_type_check
-    def get_incumbent_config(
-            self
-    ):
-        pass
+        Returns:
+
+        """
+        assert self.run_history is not None, "No Run History found, search has not been called."
+        assert not self.run_history.empty(), "Run History is empty."
```
franchuterivera, line 1255: I think this makes more sense as a ValueError? so that users can catch it if they want. Probably with a message saying:
```
Something went wrong, smac was not able to fit any model?
```

autoPyTorch/api/base_task.py
```diff
@@ -1245,17 +1239,32 @@ def __del__(self) -> None:
...
         self._backend.context.delete_directories(force=False)
 
-    @typing.no_type_check
     def get_incumbent_results(
-            self
-    ):
-        pass
+        self,
+        include_traditional: bool = False
+    ) -> Tuple[Configuration, Dict]:
```
franchuterivera, line 1245: This should be a Dict[str, Union[str, int, float]]?

test/test_api/test_api.py
```diff
@@ -177,6 +179,13 @@ def test_tabular_classification(openml_id, resampling_strategy, backend):
     score = estimator.score(y_pred, y_test)
     assert 'accuracy' in score
 
+    # check incumbent config and results
+    incumbent_config, incumbent_results = estimator.get_incumbent_results()
+    assert isinstance(incumbent_config, Configuration)
```
franchuterivera, line 184: Do you think you can create a unit test, like a separate one to test the `include_traditional`?

What we can do is mock.patch the evaluator.run() method to always return a desired value and quickly compare that everything is working as expected?

For example, the tae evaluator run could return:

- time.time() as cost, so we would expect that the best configuration is for sure a lgbm which is the first thing that is ran
- Then without the include, it should be a pytorch mode

What do you think?

autoPyTorch/api/base_task.py
ravinkohli: so we dont have pipeline configuration in the run info as well as additional info

autoPyTorch/api/base_task.py
ravinkohli: yeah agreed

autoPyTorch/api/base_task.py
ravinkohli: if we want to be more specific, then yes. I'll do it

test/test_api/test_api.py
ravinkohli: Can you tell me how to just set cost to time.time(). As the TAE.run returns status, cost, runtime, additional_run_info, how can I set cost as time and maintain the other info?

autoPyTorch/api/base_task.py
franchuterivera: Thanks, it is not only being specific but if not specified, it is converted to `Dict[Any, Any]` which basically checks almost nothing :)

test/test_api/test_api.py
franchuterivera: Todo a patch please follow [this](https://docs.python.org/3/library/unittest.mock.html)

So what we do is that we set as return value a tuple like:
```
eval_patch.return_value = (StatusType.SUCCESS, time.time(), {})
```

We can also provide a tae function [here](https://github.com/automl/Auto-PyTorch/blob/a4e08e297dc298b53d5962f0f4b4e55b5e9ded8c/autoPyTorch/optimizer/smbo.py#L254)
 I think... maybe it is easier?

PR #173, opened 2021-04-15
Fit_pipeline honoring api constraints

### LINKED ISSUE #149: Improve fitting a pipeline to the dataset
The [fit](https://github.com/automl/Auto-PyTorch/blob/db70a436cc8044a7bc03ad8846012bb3651eaca5/autoPyTorch/api/base_task.py#L1075) method currently does not honour the user constraints set while initialising or running the search. These include:
1. Memory constraint
2. Function time limit
3. Resampling strategy
4. include and exclude components and search space updates.

Also in its current format, it can not train the pipeline using cross validation. 

We already have the infrastructure to do so, i.e, the TAE similar to how dummy and traditional predictions are done. 

For an idea, we can also use [auto-sklearn's](https://github.com/automl/auto-sklearn/pull/1096) implementation for this.

ravinkohli:
This PR fixes #149. 

The changes are as follows:_ 
1. the name of api.fit to api.fit_pipeline to avoid ambiguity. 
2. The pipeline is now fit using the TAE which honors the memory and time limit constraints using pynisher.
3. fixes a bug in `build_pipeline`, where it was not taking the include, exclude and search space updates into account. [here](https://github.com/automl/Auto-PyTorch/blob/a4e08e297dc298b53d5962f0f4b4e55b5e9ded8c/autoPyTorch/api/tabular_classification.py#L111)
4. Adds tests for ensuring we can always fit a pipeline and also use it to predict and score on the data. 
5. Aims to reduce the ambiguity regarding disable_file_output a various places in the API

### DIFF
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -131,6 +131,8 @@ class BaseTask:
             required for the chosen resampling strategy. If None, uses
             the default values provided in DEFAULT_RESAMPLING_PARAMETERS
             in ```datasets/resampling_strategy.py```.
+        task_type (str): The task of the experiment as a string. Currently, supported
+            tasks are 'tabular_classification' and 'tabular_regression'
     """
 
     def __init__(
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -239,7 +239,7 @@ class BaseTask:
                     y_train: Union[List, pd.DataFrame, np.ndarray],
                     X_test: Union[List, pd.DataFrame, np.ndarray],
                     y_test: Union[List, pd.DataFrame, np.ndarray],
-                    resampling_strategy: Union[CrossValTypes, HoldoutValTypes] = HoldoutValTypes.holdout_validation,
+                    resampling_strategy: Optional[Union[CrossValTypes, HoldoutValTypes]] = None,
                     resampling_strategy_args: Optional[Dict[str, Any]] = None,
                     dataset_name: Optional[str] = None,
                     return_only: Optional[bool] = False
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -13,7 +13,7 @@ import unittest.mock
 import uuid
 import warnings
 from abc import abstractmethod
-from typing import Any, Callable, Dict, List, Optional, Union, cast
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union, cast
 
 from ConfigSpace.configuration_space import Configuration, ConfigurationSpace
 
@@ -25,7 +25,7 @@ import numpy as np
 
 import pandas as pd
 
-from smac.runhistory.runhistory import DataOrigin, RunHistory
+from smac.runhistory.runhistory import DataOrigin, RunHistory, RunInfo, RunValue
 from smac.stats.stats import Stats
 from smac.tae import StatusType
 
@@ -122,6 +122,15 @@ class BaseTask:
         exclude_components (Optional[Dict]): If None, all possible components are used.
             Otherwise specifies set of components not to use. Incompatible with include
             components
+        search_space_updates (Optional[HyperparameterSearchSpaceUpdates]): updates to be made
+            to the hyperparameter search space of the pipeline
+        resampling_strategy (Union[CrossValTypes, HoldoutValTypes]),
+                (default=HoldoutValTypes.holdout_validation):
+                strategy to split the training data.
+        resampling_strategy_args (Optional[Dict[str, Any]]): arguments
+            required for the chosen resampling strategy. If None, uses
+            the default values provided in DEFAULT_RESAMPLING_PARAMETERS
+            in ```datasets/resampling_strategy.py```.
     """
 
     def __init__(
@@ -144,6 +153,26 @@ class BaseTask:
         search_space_updates: Optional[HyperparameterSearchSpaceUpdates] = None,
         task_type: Optional[str] = None
     ) -> None:
+        """
+
+        Args:
+            seed:
+            n_jobs:
+            logging_config:
+            ensemble_size:
+            ensemble_nbest:
+            max_models_on_disc:
+            temporary_directory:
+            output_directory:
+            delete_tmp_folder_after_terminate:
+            delete_output_folder_after_terminate:
+            include_components:
+            exclude_components:
+            backend:
+
+            :
+            task_type:
+        """
         self.seed = seed
         self.n_jobs = n_jobs
         self.ensemble_size = ensemble_size
@@ -205,7 +234,11 @@ class BaseTask:
         raise NotImplementedError
 
     @abstractmethod
-    def build_pipeline(self, dataset_properties: Dict[str, Any]) -> BasePipeline:
+    def build_pipeline(self, dataset_properties: Dict[str, Any],
+                       include_components: Optional[Dict] = None,
+                       exclude_components: Optional[Dict] = None,
+                       search_space_updates: Optional[HyperparameterSearchSpaceUpdates] = None
+                       ) -> BasePipeline:
         """
         Build pipeline according to current task
         and for the passed dataset properties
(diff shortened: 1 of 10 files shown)


autoPyTorch/api/tabular_classification.py
```diff
@@ -108,16 +108,60 @@ def _get_required_dataset_properties(self, dataset: BaseDataset) -> Dict[str, An
                 'numerical_columns': dataset.numerical_columns,
                 'categorical_columns': dataset.categorical_columns}
 
-    def build_pipeline(self, dataset_properties: Dict[str, Any]) -> TabularClassificationPipeline:
+    def build_pipeline(self, dataset_properties: Dict[str, Any],
+                       include_components: Optional[Dict] = None,
+                       exclude_components: Optional[Dict] = None,
+                       search_space_updates: Optional[HyperparameterSearchSpaceUpdates] = None
+                       ) -> TabularClassificationPipeline:
         return TabularClassificationPipeline(dataset_properties=dataset_properties)
```
ArlindKadra, line 116: Should the include_components, exclude_components and search space updates also be present in line 116 ? in the return. They are for tabular regression.

ArlindKadra:
@ravinkohli As an additional note, please add this to the `regularization_cocktails `branch first. We need it there asap.

autoPyTorch/api/base_task.py
```diff
@@ -144,6 +153,26 @@ def __init__(
         search_space_updates: Optional[HyperparameterSearchSpaceUpdates] = None,
         task_type: Optional[str] = None
     ) -> None:
+        """
+
+        Args:
```
franchuterivera, line 158: I think this docstring is WIP :)?

autoPyTorch/api/base_task.py
```diff
@@ -215,7 +248,23 @@ def build_pipeline(self, dataset_properties: Dict[str, Any]) -> BasePipeline:
...
+    @abstractmethod
+    def _create_dataset(self,
+                        X_train: Union[List, pd.DataFrame, np.ndarray],
+                        y_train: Union[List, pd.DataFrame, np.ndarray],
+                        X_test: Union[List, pd.DataFrame, np.ndarray],
+                        y_test: Union[List, pd.DataFrame, np.ndarray],
+                        resampling_strategy: Union[CrossValTypes, HoldoutValTypes] = HoldoutValTypes.holdout_validation,
+                        resampling_strategy_args: Optional[Dict[str, Any]] = None,
+                        dataset_name: Optional[str] = None,
+                        return_only: Optional[bool] = False
```
franchuterivera, line 264: Instead of having a return_only, maybe create it once? 

So with return only, every time you call this method you will have to create and save a new pipeline. Do you think it is better to:


1. The first thing this function does is check if load_datamanger from the backend is able to load a dataset, and if it is the case, it loads it rather than creating a new one. 
2. Else a dataset is created

What do you think?

autoPyTorch/api/base_task.py
```diff
@@ -1079,73 +1135,223 @@ def refit(
...
-        if self.dataset_name is None:
-            self.dataset_name = str(uuid.uuid1(clock_seq=os.getpid()))
 
-        if self._logger is None:
-            self._logger = self._get_logger(self.dataset_name)
+        resampling_strategy = resampling_strategy if resampling_strategy is not None else self.resampling_strategy
+        resampling_strategy_args = resampling_strategy_args if resampling_strategy_args is not None else \
+            self.resampling_strategy_args
+
+        dataset = self._create_dataset(X_train=X_train,
```
franchuterivera, line 1236: If we enhance the create_dataset, we can call fit_pipeline multiple times without the need for a new dataset each time. Plus the training split will not change. Also, the name then probably should be `get_dataset`

autoPyTorch/api/base_task.py
```diff
@@ -1079,73 +1135,223 @@ def refit(
...
+                                       dataset_name=dataset_name,
+                                       return_only=True)
 
         # get dataset properties
         dataset_requirements = get_dataset_requirements(
             info=self._get_required_dataset_properties(dataset))
         dataset_properties = dataset.get_dataset_properties(dataset_requirements)
         self._backend.save_datamanager(dataset)
 
+        self._backend._make_internals_directory()
```
franchuterivera, line 1251: If this is troublesome to have it here, then let us move it to the constructor

autoPyTorch/api/base_task.py
```diff
@@ -1079,73 +1135,223 @@ def refit(
...
+            exclude_components = self.exclude_components
+        if search_space_updates is None:
+            search_space_updates = self.search_space_updates
+
+        pipeline = self.build_pipeline(dataset_properties=dataset_properties,
+                                       include_components=include_components,
+                                       exclude_components=exclude_components,
+                                       search_space_updates=search_space_updates)
+        if configuration is None:
+            configuration = pipeline.get_hyperparameter_search_space().get_default_configuration()
```
franchuterivera, line 1269: Do you think it makes sense to make this more flexible? Why the default and not a random. Maybe we make configuration a required argument? Shouldn't this be passed to the build_pipeline?

autoPyTorch/api/base_task.py
```diff
@@ -1079,73 +1135,223 @@ def refit(
...
-        if pipeline_config is not None:
-            pipeline.set_hyperparameters(pipeline_config)
+        if include_components is None:
+            include_components = self.include_components
+        if exclude_components is None:
+            exclude_components = self.exclude_components
+        if search_space_updates is None:
+            search_space_updates = self.search_space_updates
+
+        pipeline = self.build_pipeline(dataset_properties=dataset_properties,
```
franchuterivera, line 1264: I wonder if we need this build_pipeline? Can we only rely on the `ExecuteTaFuncWithQueue`? This way configuration can be a int or a string, and then we get a traditional pipeline for example?

test/test_api/test_api.py
```diff
@@ -449,3 +451,92 @@ def test_do_dummy_prediction(dask_client, fit_dictionary_tabular):
     estimator._clean_logger()
 
     del estimator
+
+
+@pytest.mark.parametrize("disable_file_output", [True, False])
+@pytest.mark.parametrize('openml_id', (40981,))
```
franchuterivera, line 457: Can we test another configuration :), I think we always test Australian... Maybe add:

- https://www.openml.org/d/21
- https://www.openml.org/d/40984
- https://www.openml.org/d/1066

Why are very fast?

test/test_api/test_api.py
```diff
@@ -449,3 +451,92 @@ def test_do_dummy_prediction(dask_client, fit_dictionary_tabular):
...
+@pytest.mark.parametrize('resampling_strategy', (HoldoutValTypes.holdout_validation,
+                                                 CrossValTypes.k_fold_cross_validation,
+                                                 ))
+def test_pipeline_fit(openml_id, resampling_strategy, backend, disable_file_output):
+    # Get the data and check that contents of data-manager make sense
+    X, y = sklearn.datasets.fetch_openml(
+        data_id=int(openml_id),
+        return_X_y=True, as_frame=True
+    )
+    X_train, X_test, y_train, y_test = sklearn.model_selection.train_test_split(
```
franchuterivera, line 467: Let us make this function faster with a 20% for train and 80% for testing

franchuterivera (changes requested):
Thank you very much, I believe this function is critical for debug especially.  I left some few comments to further make it more useful.

autoPyTorch/api/tabular_classification.py
ravinkohli: oops, I seem to have missed this. Thanks for pointing out

test/test_api/test_api.py
ravinkohli: okay

autoPyTorch/api/base_task.py
ravinkohli: oh its not important. We already have the documentation for BaseTask. I'll remove this

autoPyTorch/api/base_task.py
ravinkohli: that makes sense

autoPyTorch/api/base_task.py
ravinkohli: So I think I'll remove this part, and make configuration a required argument. This will then be passed to the TAE and that will take care of the rest.

test/test_api/test_api.py
ravinkohli: What do you mean by why are very fast? Yes, I'll put one of these configuration

autoPyTorch/api/base_task.py
ravinkohli: I think it is already in the backend constructor so I'll remove it

autoPyTorch/api/base_task.py
ravinkohli: The reason I had it create and always save a dataset, is that in case we want to have a different dataset to fit a pipeline, i.e, once we have made a search, we can try and fit the best incumbent configuration to a larger dataset or a different related dataset. return_only allows us to use the same function to create a dataset, whenever we get new data, i.e, in search, fit and refit. As we are also returning the dataset after the fit_pipeline method, the same dataset can be used to predict and score using the pipeline. 

Do you still think it is better to load a datamanager from the backend and use it? If so, do you know of a way to achieve them both.

autoPyTorch/api/base_task.py
ArlindKadra: Looking more into it, please also add `task_type `in the docs for `base_task`

ArlindKadra (approved):
Great, looks good. You can merge it as soon as the unit tests pass.

PR #174, opened 2021-04-15
Make it possible to disable imputation

dvzubarev:


ravinkohli:
Thanks for this PR however, we have moved to v0.1.0 which has a completely different implementation. Therefore, we will not be looking into this PR anytime soon.

PR #175, opened 2021-04-19
[ADD] Get incumbent config

### LINKED ISSUE #150: Return the best configuration found by the optimizer

ravinkohli:
Fixes #150 

Changes made-

1.  removed get_incumbent_config from base task as it would be redundant to have two separate functions for getting configuration and the results
2.  get_incumbent_results goes through the run history and gets the config that has the minimum cost on the validation set.
3.  Also returns the additional info which includes optimisation loss on the other metrics specified and the train loss.
4.  Adds test to check if the function is working properly


autoPyTorch/api/base_task.py
```diff
@@ -954,8 +953,11 @@ def _search(
             )
             try:
                 run_history, self.trajectory, budget_type = \
-                    _proc_smac.run_smbo()
+                    _proc_smac.run_smbo(func=tae_func)
                 self.run_history.update(run_history, DataOrigin.INTERNAL)
+                # SMAC only saves pipeline run history, also save final run history to disk
```
franchuterivera, line 958: sorry, what do you mean by this? Is this overwriting the original run history?

autoPyTorch/api/base_task.py
```diff
@@ -1237,17 +1239,35 @@ def __del__(self) -> None:
...
+        assert self.run_history is not None, "No Run History found, search has not been called."
+        if self.run_history.empty():
+            raise ValueError("Run History is empty. Something went wrong, "
+                             "smac was not able to fit any model?")
+
+        run_history_data = self.run_history.data
+        if not include_traditional:
+            # traditional classifiers have trainer_configuration in their additional info
+            run_history_data = dict(
+                filter(lambda elem: elem[1].additional_info is not None and 'trainer_configuration' not in elem[1].
```
franchuterivera, line 1263: I am wondering if we can make this more robust. Can you add an attribute to `additional_info` so that we know for sure that this is a traditional configuration?

test/test_api/test_api.py
```diff
@@ -15,22 +18,105 @@
...
 from autoPyTorch.datasets.resampling_strategy import (
     CrossValTypes,
     HoldoutValTypes,
 )
+from autoPyTorch.datasets.tabular_dataset import TabularDataset
+from autoPyTorch.evaluation.train_evaluator import TrainEvaluator
 from autoPyTorch.optimizer.smbo import AutoMLSMBO
 from autoPyTorch.pipeline.components.training.metrics.metrics import accuracy
 
+from api_utils import dummy_traditional_classification  # noqa (E402: module level import not   at top of file)
```
franchuterivera, line 37: Rather than having the noqa, isn't it better to put it in the correct location?

test/test_api/test_api.py
```diff
@@ -449,3 +551,59 @@ def test_do_dummy_prediction(dask_client, fit_dictionary_tabular):
...
+            estimator.search(
+                X_train=X_train, y_train=y_train,
+                X_test=X_test, y_test=y_test,
+                optimize_metric='accuracy',
+                total_walltime_limit=150,
+                func_eval_time_limit_secs=50,
+                enable_traditional_pipeline=True,
+                load_models=False,
+            )
+    config, results = estimator.get_incumbent_results(include_traditional=include_traditional)
```
franchuterivera, line 604: is it possible to check that the returned config has the lowest cost?

franchuterivera (changes requested):
Hello, it looks really nice. Some minor comments to see If we can further improve this pr

autoPyTorch/api/base_task.py
ravinkohli: no I'm creating it in a separate folder. I wanted to save both the traditional and smac run history combined in one plaace

autoPyTorch/api/base_task.py
ravinkohli: makes sense. Or I can use the data origin external flag maybe

test/test_api/test_api.py
ravinkohli: I thought its a test so didnt care, but surely I can do that

test/test_api/test_api.py
ravinkohli: sure

autoPyTorch/api/base_task.py
```diff
@@ -1244,17 +1242,37 @@ def __del__(self) -> None:
...
-    @typing.no_type_check
-    def get_incumbent_config(
-            self
-    ):
-        pass
+        Returns:
+
+        """
+        assert self.run_history is not None, "No Run History found, search has not been called."
+        if self.run_history.empty():
```
franchuterivera, line 1258: maybe you can add the above assert here, like:
```
if self.run_history is none or self.run_history.empty():
```

autoPyTorch/evaluation/train_evaluator.py
```diff
@@ -172,10 +172,8 @@ def fit_predict_and_loss(self) -> None:
 
             status = StatusType.SUCCESS
 
-            self.logger.debug("In train evaluator fit_predict_and_loss, num_run: {} loss:{}".format(
-                self.num_run,
-                loss
-            ))
+            self.logger.debug("In train evaluator fit_predict_and_loss, num_run: {} loss:{},"
```
franchuterivera, line 175: can you also add the status here? This is useful when you have to wait for a long time to get the runhistory.

test/test_api/test_api.py
```diff
@@ -464,3 +483,70 @@ def test_do_dummy_prediction(dask_client, fit_dictionary_tabular):
...
+    assert isinstance(config, Configuration)
+    assert isinstance(results, dict)
+
+    run_history_data = estimator.run_history.data
+    costs = [run_value.cost for run_key, run_value in run_history_data.items() if run_value.additional_info is not None
+             and (run_value.additional_info['configuration_origin'] != 'traditional' or include_traditional)]
+    assert results['opt_loss']['accuracy'] == min(costs)
+
+    if not include_traditional:
+        assert results['configuration_origin'] != 'traditional'
```
franchuterivera, line 552: can you check the else statement here also?

ravinkohli: what do I check there? It could be both traditional or our pipeline in case include_traditional is true

autoPyTorch/api/base_task.py
ravinkohli: doing so causes mypy to not understand that we checked for run_history is not None. I get `"None" has no attribute "<attr>"` for various statements in this function

franchuterivera: Oh, it was because of mypy, ok ok makes sense

test/test_api/test_api.py
franchuterivera: yes, exactly that.... it is just to prevent some future edits breaking the code and removing the configuration origin

test/test_pipeline/components/training/base.py
```diff
@@ -22,7 +22,7 @@ def prepare_trainer(self,
                         n_samples: int,
                         trainer: BaseTrainerComponent,
                         task_type: int,
-                        epochs=50):
+                        epochs=100):
```
franchuterivera, line 25: I am also adding changes to my PR for this one! I see some random failures.

Issue #186, opened 2021-04-26
[test] Too much runtime for pytest


nabenabe0928:
Currently, pytest takes 2000 seconds in total and we need to shorten the test time.

The current longest calls are the followings:
```
377.47s call     test/test_pipeline/test_tabular_classification.py::TestTabularClassification::test_pipeline_transform[classification_numerical_only]
168.82s call     test/test_api/test_api.py::test_tabular_classification[CrossValTypes.k_fold_cross_validation-40981]
167.68s call     test/test_api/test_api.py::test_tabular_classification[HoldoutValTypes.holdout_validation-40981]
138.26s call     test/test_pipeline/components/training/test_training.py::BaseTrainerComponentTest::test_evaluate
76.01s call     test/test_pipeline/test_tabular_classification.py::TestTabularClassification::test_pipeline_fit[classification_categorical_only]
61.01s call     test/test_api/test_api.py::test_tabular_regression[HoldoutValTypes.holdout_validation-boston]
60.88s call     test/test_api/test_api.py::test_tabular_regression[CrossValTypes.k_fold_cross_validation-boston]
48.10s call     test/test_pipeline/test_tabular_classification.py::TestTabularClassification::test_pipeline_fit[classification_numerical_and_categorical]
45.40s call     test/test_pipeline/test_tabular_classification.py::TestTabularClassification::test_pipeline_predict[classification_numerical_only]
23.33s call     test/test_pipeline/test_tabular_classification.py::TestTabularClassification::test_pipeline_transform[classification_numerical_and_categorical]
20.08s call     test/test_pipeline/components/setup/test_setup_networks.py::TestNetworks::test_pipeline_fit[NoEmbedding-ResNetBackbone-fully_connected-classification_numerical_only]
19.99s call     test/test_pipeline/components/setup/test_setup_networks.py::TestNetworks::test_pipeline_fit[LearnedEntityEmbedding-ShapedResNetBackbone-fully_connected-classification_categorical_only]
19.89s call     test/test_pipeline/components/setup/test_setup_networks.py::TestNetworks::test_pipeline_fit[NoEmbedding-ShapedResNetBackbone-fully_connected-classification_categorical_only]
19.89s call     test/test_pipeline/components/setup/test_setup_networks.py::TestNetworks::test_pipeline_fit[NoEmbedding-ShapedResNetBackbone-fully_connected-classification_numerical_only]
[...]

PR #187, opened 2021-04-26
[test] Reduce the number of sample size in unittests for shorter tests

nabenabe0928:
Since the unittests take quite a long time, I set the number of samples for each test smaller.


ravinkohli:
Addresses #186

PR #190, opened 2021-04-30
[Feature] Introduces no resampling strategy to enable fitting on the whole dataset

### LINKED ISSUE #191: Refit on the whole training set not supported
The `fit_pipeline` method does not support refitting on the training set with 0 in the resampling strategy arguments for the validation set.
This should be fixed since it is a common scenario for the library.

ravinkohli:
This PR includes the following:

1. Added NoResamplingStrategyTypes to resampling_strategy.py with an option for shuffling and not shuffling the data
2. Added FitEvaluator which handles fit and predict a pipeline in case the resampling strategy is in NoResamplingStrategyTypes
3. Fixes bug in BaseDataLoader where test loader was not being created
4. Adds capability to base trainer choice to use val_data_loader or test_data_loader depending on availability.
5. Fixes bug in trainer initialization which was causing weighted strategy to be used in regression
6. Fixes bug in getting updates for base trainer choice and its components. More specifically, it was causing Lookahead to be repeated two times in the name of the hyperparameter, for example `'la_alpha' was Lookahead:Lookahead:la_alpha` should be `Lookahead:la_alpha`

Fixes #191 and #192 
Partially addresses #186. 
P.S. if you have any suggestions for a better name than FitEvaluator please let me know.


ArlindKadra:
Addresses #191

autoPyTorch/evaluation/fit_evaluator.py
```diff
@@ -0,0 +1,317 @@
...
+        return Y_train_pred, Y_valid_pred, Y_test_pred
+
+    def _predict(self, pipeline: BaseEstimator,
+                 train_indices: Union[np.ndarray, List]
+                 ) -> Tuple[np.ndarray, Optional[np.ndarray], Optional[np.ndarray]]:
+
+        train_pred = self.predict_function(subsampler(self.X_train, train_indices), pipeline,
+                                           self.y_train[train_indices])
+
+        if self.X_valid is not None:
```
ArlindKadra, line 153: Would we ever have a value for `x_valid `since this is a `fit_evaluator` only for the `no_resampling_strategy`

autoPyTorch/evaluation/fit_evaluator.py
```diff
@@ -0,0 +1,317 @@
...
+        if file_output:
+            loss_, additional_run_info_ = self.file_output(
+                None, valid_pred, test_pred,
+            )
+        else:
+            loss_ = None
+            additional_run_info_ = {}
+
+        validation_loss, test_loss = self.calculate_auxiliary_losses(
+            valid_pred, test_pred
```
ArlindKadra, line 191: same here for `valid_pred`

autoPyTorch/evaluation/fit_evaluator.py
```diff
@@ -0,0 +1,317 @@
...
+        additional_run_info = (
+            {} if additional_run_info is None else additional_run_info
+        )
+        for metric_name, value in loss.items():
+            additional_run_info[metric_name] = value
+        additional_run_info['duration'] = self.duration
+        additional_run_info['num_run'] = self.num_run
+        if train_loss is not None:
+            additional_run_info['train_loss'] = train_loss
+        if validation_loss is not None:
```
ArlindKadra, line 208: same for `validation_loss`

autoPyTorch/evaluation/fit_evaluator.py
```diff
@@ -0,0 +1,317 @@
...
+                     'additional_run_info': additional_run_info,
+                     'status': status}
+
+        self.queue.put(rval_dict)
+        return None
+
+    def file_output(
+        self,
+        Y_optimization_pred: np.ndarray,
+        Y_valid_pred: np.ndarray,
```
ArlindKadra, line 223: same for `Y_valid_pred`

autoPyTorch/evaluation/fit_evaluator.py
```diff
@@ -0,0 +1,317 @@
...
+        Y_test_pred: np.ndarray,
+    ) -> Tuple[Optional[float], Dict]:
+
+        # Abort if predictions contain NaNs
+        for y, s in [
+            [Y_valid_pred, 'validation'],
```
ArlindKadra, line 229: same.

ArlindKadra (approved):
Thanks for the PR, it looks really good :).
I left a few comments since I was uncertain on certain aspects.

autoPyTorch/evaluation/fit_evaluator.py
ravinkohli: but this X_valid is if the user provides a validation set. So separate from the train set if a user wants to provide a valid set, they can do so here

autoPyTorch/api/tabular_regression.py
```diff
@@ -244,12 +249,15 @@ def search(
...
+        assert isinstance(self.resampling_strategy, (CrossValTypes, HoldoutValTypes)), \
+            "Val Split is required for HPO search. " \
+            "Expected 'self.resampling_strategy' in" \
+            " '(CrossValTypes, HoldoutValTypes) got {}".format(self.resampling_strategy)
+
         self.get_dataset(X_train=X_train,
                          y_train=y_train,
                          X_test=X_test,
                          y_test=y_test,
-                         resampling_strategy=self.resampling_strategy,
```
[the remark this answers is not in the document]
ravinkohli: yeah its being done inside

autoPyTorch/evaluation/fit_evaluator.py
ArlindKadra: Okay, so this is if the user maybe wants to know how much it changed compared for example to a previous run when he ran on only `X_train`, compared to running on `X_train + X_valid`

ravinkohli: yeah or just having a validation set, I don't really think there's much point to it. but for now I think its okay. we may need to decide whether to keep it later for the library. As it is not gonna be used to report epoch wise statistics yet

ArlindKadra: True, we can change it in the future.

Issue #195, opened 2021-05-03
Imbalanced dataset problems


ArlindKadra:
Currently, there are a few problems that crash the library when working with imbalanced datasets.

- There are Null values and one unique value in the train split and Null values in the test set. The library will try to convert the feature column to float on the test set and it will throw an error. This condition should be removed.
- There are Null values in the train split and Null and a few values at the test split, the other non-null values in the test set should be ignored so that the run does not crash.

ravinkohli:
I'll leave this issue open for reference as it is still not fixed in the development branch

PR #197, opened 2021-05-04
Fixing issues with imbalanced datasets

ArlindKadra:
This PR fixes issue #195

### DIFF
--- a/autoPyTorch/data/tabular_feature_validator.py
+++ b/autoPyTorch/data/tabular_feature_validator.py
@@ -174,8 +174,6 @@ class TabularFeatureValidator(BaseFeatureValidator):
         if self.encoder is not None:
             X = self.encoder.transform(X)
 
-
-
         # Sparse related transformations
         # Not all sparse format support index sorting
         if scipy.sparse.issparse(X) and hasattr(X, 'sort_indices'):
--- a/autoPyTorch/data/tabular_feature_validator.py
+++ b/autoPyTorch/data/tabular_feature_validator.py
@@ -49,9 +49,13 @@ class TabularFeatureValidator(BaseFeatureValidator):
             # with nan values.
             # Columns that are completely made of NaN values are provided to the pipeline
             # so that later stages decide how to handle them
+
+            # Clear whatever null column markers we had previously
+            self.null_columns.clear()
             if np.any(pd.isnull(X)):
                 for column in X.columns:
                     if X[column].isna().all():
+                        self.null_columns.add(column)
                         X[column] = pd.to_numeric(X[column])
                         # Also note this change in self.dtypes
                         if len(self.dtypes) != 0:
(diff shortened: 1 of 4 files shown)


ravinkohli (changes requested):
Hey Arlind, thanks for the PR. Its essential in the corner case when we have all nans in train but not in test and vice versa. But, could you add tests for the validator to ensure we can handle the case properly. It would be great if we could test the two cases- 1. where we have all NaNs in train and not in test and 2. where we have all NaNs in test but not in train

ArlindKadra:
> Hey Arlind, thanks for the PR. Its essential in the corner case when we have all nans in train but not in test and vice versa. But, could you add tests for the validator to ensure we can handle the case properly. It would be great if we could test the two cases- 1. where we have all NaNs in train and not in test and 2. where we have all NaNs in test but not in train

Yep, I will do just that today. But overall, I wanted to get a quick idea on how the problem is handled now, in case I missed something on the fix.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -146,32 +151,35 @@ def transform(
...
+            for column in self.null_columns:
+                # The column is not null, make it null since it was null in fit.
+                if not X[column].isna().all():
+                    X[column] = np.NaN
+                X[column] = pd.to_numeric(X[column])
+
+            # for the test set, if we have columns with only null values
+            # they will probably have a numeric type. If these columns were not
+            # with only null values in the train set, they should be converted
+            # to the category type.
```
ravinkohli, line 167:
```suggestion
            # to the type it had while fitting.
```

I don't think we should wait for the tests to merge in case you choose to incorporate this suggestion

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -414,7 +422,7 @@ def infer_objects(self, X: pd.DataFrame) -> pd.DataFrame:
                     # In the case train data was interpreted as int
                     # and test data was interpreted as float, because of 0.0
                     # for example, honor training data
-                    X[key] = X[key].applymap(np.int64)
+                    X[key] = X[key].astype(np.int64)
```
ravinkohli, line 425: should we remove the if-else statement? as we are treating them equally

ArlindKadra: Yep, nice catch.

autoPyTorch/data/tabular_feature_validator.py
ArlindKadra: I also changed the comment accordingly.

test/test_data/test_feature_validator.py
```diff
@@ -515,11 +518,73 @@ def test_featurevalidator_new_data_after_fit(openml_id,
...
+    assert train_feature_types == ['categorical', 'numerical', 'numerical']
+
+    transformed_X_test = validator.transform(X_test)
+    transformed_X_test = pd.DataFrame(transformed_X_test)
+    null_columns = []
+    for column in transformed_X_test.columns:
+        if transformed_X_test[column].isna().all():
+            null_columns.append(column)
+
+    assert null_columns == [1]
```
ravinkohli, line 590: why is null_columns not empty here?

ArlindKadra: Because for the test set, column 'A' will have only null values, however, it will have a numeric type. While column 'C' will have 1, as it will detect correctly that it has 2 categories from the train set.

ravinkohli: I see, sorry I got confused as I expected none to be Null, I don't know why

ravinkohli (approved):

Issue #142, opened 2021-03-19
[refactor] Change name from `resampling_strategy` to `split_fn`


nabenabe0928:
I really would like to change the name `resampling_strategy` to `split_fn` for several reasons.

1. the current names are too long
2. usually we cannot write if statements in 1 lines because of the length
3. the names themselves are not familiar in general
In fact, the cases hit by google search are:
```
1. cross validation splitting function: 144,000,000
2. cross validation split function:      67,500,000
3. cross validation resampling function:  1,460,000
4. cross validation resampling strategy:    944,000
```
I actually did not understand what resampling strategy means at a first glance.
Since AutoPytorch is provided for wide range of users, these recognition gaps (about 100x difference) are quite big and should be considered more seriously.

PR #199, opened 2021-05-07
[refactor] Rename resampling strategy to split_fn

nabenabe0928:
Address the [Issue#142](https://github.com/automl/Auto-PyTorch/issues/142).


ArlindKadra:
I am actually not in favor of this PR. This diverges us from the autosklearn library and additionally, we still have issues that have a higher priority. Although I would say I understand the points you are mentioning, I think at this moment this is not what we want to pursue.

mfeurer:
> This diverges us from the autosklearn library and additionally

I think it would be great to apply such changes once there is further progress with [automl_common](https://github.com/automl/automl_common)

PR #200, opened 2021-05-07
[feature] Greedy Portfolio

ravinkohli:
This PR adds the greedy portfolio to the smac run.

It addresses the following-
1. Flag to determine whether to initialize smac with configurations from the greedy portfolio
2. The version on the development branch should be 0.1.0 and also we are now using smac instead of bohb, so I changed that. Feel free to make a suggestion if you disagree.


autoPyTorch/api/base_task.py
```diff
@@ -701,6 +701,7 @@ def _search(
         precision: int = 32,
         disable_file_output: List = [],
         load_models: bool = True,
+        run_greedy_portfolio: bool = False
```
nabenabe0928, line 704: Thanks for the enhancement:)
It looks good to me.
But can we make the argument more generic?
For example, it is likely to add another portfolio selection in the future, right?
So it would be better what kind of portfolio we build by an arg.
I would prefer `portfolio_selection='greedy'` rather than `run_greedy_portfolio`

autoPyTorch/api/base_task.py
```diff
@@ -771,7 +775,12 @@ def _search(
             disable_file_output (Union[bool, List]):
             load_models (bool), (default=True): Whether to load the
                 models after fitting AutoPyTorch.
-
+            portfolio_selection (str), (default="none"): If "greedy",
+                runs initial configurations present in
+                'autoPyTorch/optimizer/greedy_portfolio.json'.
```
franchuterivera, line 780: Do you think it is better to put it [here](https://github.com/automl/Auto-PyTorch/tree/development/autoPyTorch/configs)?

ravinkohli: yes. I moved it there

autoPyTorch/optimizer/smbo.py
```diff
@@ -205,6 +213,24 @@ def __init__(self,
                                               port=self.logger_port)
         self.logger.info("initialised {}".format(self.__class__.__name__))
 
+        # read and validate initial configurations
+        initial_configurations = json.load(open(os.path.join(os.path.dirname(__file__),
+                                                             '../configs/greedy_portfolio.json')))
+
+        self.initial_configurations: typing.Optional[typing.List[Configuration]] = None
+        assert portfolio_selection in ['none', 'greedy']
```
franchuterivera, line 221: Can you make this a value error so users can catch it if needed? And probably it makes sense to do so in the API so that we don't even proceed with an incorrect option.

Also, how can the user provide a greedy portfolio json file? Can this be a file provided by the user?

autoPyTorch/optimizer/smbo.py
```diff
@@ -323,7 +350,8 @@ def run_smbo(self, func: typing.Optional[typing.Callable] = None
                                    n_jobs=self.n_jobs,
                                    initial_budget=initial_budget,
                                    max_budget=max_budget,
-                                   dask_client=self.dask_client)
+                                   dask_client=self.dask_client,
+                                   initial_configurations=self.initial_configurations)
```
franchuterivera, line 354: Also I just noticed that there is no unit testing for this. Do you think we should at least check that some initial configurations are valid for Australian?

franchuterivera (changes requested):
Thanks a lot, everything looks fine. Maybe we should add a small unit test to make sure at least a configuration is being identified?

autoPyTorch/optimizer/smbo.py
ravinkohli: yeah that also makes sense. I'll add this

autoPyTorch/optimizer/smbo.py
ravinkohli: okay, I'll add a test.

autoPyTorch/api/base_task.py
```diff
@@ -781,6 +793,9 @@ def _search(
             raise ValueError("Incompatible dataset entered for current task,"
                              "expected dataset to have task type :{} got "
                              ":{}".format(self.task_type, dataset.task_type))
+        if portfolio_selection not in ["none", "greedy"]:
```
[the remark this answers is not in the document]
ravinkohli: yeah this also makes sense as it will allow us to distinguish between the greedy portfolio of the paper and something the user wants to use. and if its None then no portfolio selection. I'll adjust this and rerequest

autoPyTorch/optimizer/smbo.py
```diff
@@ -152,7 +158,15 @@ def __init__(self,
                 Allows to create a user specified SMAC object
             ensemble_callback (typing.Optional[EnsembleBuilderManager]):
                 A callback used in this scenario to start ensemble building subtasks
-
+            portfolio_selection (str), (default=None):
+                This argument controls the initial configurations that
+                AutoPyTorch uses to warm start SMAC for hyperparameter
+                optimization. By default, no warm-starting happens.
+                The user can provide a path to a json file containing
+                configurations, similar to (...herepathtogreedy...).
```
franchuterivera, line 166:
```suggestion
                configurations, similar to (autoPyTorch/configs/greedy_portfolio.json).
```

autoPyTorch/optimizer/smbo.py
```diff
@@ -205,6 +219,23 @@ def __init__(self,
                                               port=self.logger_port)
         self.logger.info("initialised {}".format(self.__class__.__name__))
 
+        self.initial_configurations: typing.Optional[typing.List[Configuration]] = None
+        if portfolio_selection is not None:
+            # read and validate initial configurations
+            portfolio_path = portfolio_selection if portfolio_selection != "greedy" else \
+                os.path.join(os.path.dirname(__file__), '../configs/greedy_portfolio.json')
+            initial_configurations = json.load(open(portfolio_path))
```
franchuterivera, line 227: Can you add an if to check if the provided path does not exists and notify the user with an error?

ravinkohli: sure

franchuterivera (approved):

autoPyTorch/optimizer/smbo.py
```diff
@@ -53,6 +56,8 @@ def get_smac_object(
         ta_kwargs (typing.Dict[str, typing.Any]): Arguments to the above ta
         n_jobs (int): Amount of cores to use for this task
         dask_client (dask.distributed.Client): User provided scheduler
+        initial_configurations (typing.List[Configuration]): List of initial
+            configurations which smac will run before starting the search process
```
nabenabe0928, line 60: [These variables](https://automl.github.io/SMAC3/master/apidoc/smac.facade.smac_ac_facade.html)?
If so, could you add the link?

nabenabe0928 (commented):
Sorry for the late response, here is some short comments and I will put other comments soon.

autoPyTorch/optimizer/smbo.py
```diff
@@ -205,6 +219,28 @@ def __init__(self,
                                               port=self.logger_port)
         self.logger.info("initialised {}".format(self.__class__.__name__))
 
+        self.initial_configurations: typing.Optional[typing.List[Configuration]] = None
```
nabenabe0928, line 222: Can we make L222-243 a separated function? (even not a function inside this class)
Then we can just test this function.

test/test_api/test_api.py
```diff
@@ -464,3 +465,92 @@ def test_do_dummy_prediction(dask_client, fit_dictionary_tabular):
     estimator._clean_logger()
 
     del estimator
+
+
+@unittest.mock.patch('autoPyTorch.evaluation.train_evaluator.eval_function',
+                     new=dummy_eval_function)
+@pytest.mark.parametrize('openml_id', (40981, ))
```
nabenabe0928, line 472: Those tests can be simplified by reflecting my previous comment.

autoPyTorch/optimizer/smbo.py
ravinkohli: Could you explain your comment? I didn't get what you mean.

autoPyTorch/optimizer/smbo.py
ravinkohli: sure

test/test_api/test_api.py
ravinkohli: exactly how would these tests be simplified?

PR #172, opened 2021-04-14
[Fix] Refactor development reproducibility

franchuterivera:
FIx reproducibility in AutoPyTorch.

This is limited to CPU runs, as GPU based still will have non-deterministic algorithms enabled to take advantage of speed.

PR #205, opened 2021-05-11
Reduce run time of the test 

ravinkohli:
This PR aims to reduce the time it takes to perform tests on autoPyTorch.

It includes the following:
1. API tests now don't fit autopytorch pipelines but use dummy pipelines.
2. number of epochs, in general, have been reduced to 5 unless more required to test the performance
3. mock patch train epoch where we don't care if training is happening or not
4. number of epochs in base training has been reduced to 50 wherever possible
5. Fixes for reproducibility in the feature preprocessors
6. skip tests nbest_remembered if python >=3.7. 
7. Also removes the examples workflow

Finally, on my laptop the tests now take 17 minutes which is a considerable improvement from the 35-45 it took before.


test/test_pipeline/components/training/base.py
```diff
@@ -21,10 +21,11 @@ class BaseTraining(unittest.TestCase):
 
     def prepare_trainer(self,
                         trainer: BaseTrainerComponent,
-                        task_type: int):
+                        task_type: int,
+                        epochs=50):
         if task_type in CLASSIFICATION_TASKS:
             X, y = make_classification(
-                n_samples=5000,
+                n_samples=500,
```
nabenabe0928, line 28: Can we define a constant variable for this so that we can easily change the value by one line modification in the future?

test/test_pipeline/components/training/test_training.py
```diff
@@ -165,7 +165,8 @@ def test_regression_epoch_training(self):
          _,
          epochs,
          logger) = self.prepare_trainer(StandardTrainer(),
-                                        constants.TABULAR_REGRESSION)
+                                        constants.TABULAR_REGRESSION,
+                                        1000)
```
nabenabe0928, line 169: Can we rewrite like this? (Same for the rest)
```
EPOCH = 1000

... = self.prepare_trainer(StandardTrainer(),
                                       constants.TABULAR_REGRESSION)
                                       constants.TABULAR_REGRESSION,
                                       epoch=1000)

```

test/test_pipeline/test_tabular_classification.py
```diff
@@ -405,6 +384,8 @@ def test_constant_pipeline_iris(fit_dictionary_tabular):
                                              search_space_updates=search_space_updates)
 
     fit_dictionary_tabular['additional_metrics'] = ['balanced_accuracy']
+    # increase number of epochs to test for performance
+    fit_dictionary_tabular['epochs'] = 50
```
nabenabe0928, line 388: Repeat from another comment.
Can we remove it or can we use NamedTuple instead? Because it is already defined in the contest and we do not want to change the variable locally. Especially, I am afraid of modifying the value without being recognized by anyone. If it is not appropriate change for this PR, I will raise an issue.

test/test_pipeline/components/setup/test_setup_networks.py
```diff
@@ -31,6 +31,9 @@ def test_pipeline_fit(self, fit_dictionary_tabular, embedding, backbone, head):
         """This test makes sure that the pipeline is able to fit
         every combination of network embedding, backbone, head"""
 
+        # increase number of epochs to test for performance
+        fit_dictionary_tabular['epochs'] = 50
```
nabenabe0928, line 35: Can we remove it or can we use NamedTuple instead? Because it is already defined in the contest and we do not want to change the variable locally. Especially, I am afraid of modifying the value without being recognized by anyone. If it is not appropriate change for this PR, I will raise an issue.

test/test_pipeline/components/training/test_training.py
ravinkohli: I can do this sure. but I didn't do it originally because its only used in two of the tests

test/test_pipeline/components/setup/test_setup_networks.py
ravinkohli: I didnt get the problem here. Could you explain again? For certain tests we want to check the performance of the pipelines, and hence we want higher number of epochs. Where do you want to use a NamedTuple?

test/test_pipeline/components/training/test_training.py
nabenabe0928: Ok, that also makes sense, but I just found out last time when I fixed the variable values in `conftest.py`, it was just annoying.

test/test_pipeline/components/training/base.py
```diff
@@ -42,7 +43,7 @@ def prepare_trainer(self,
 
         elif task_type in REGRESSION_TASKS:
             X, y = make_regression(
-                n_samples=5000,
+                n_samples=500,
```
franchuterivera, line 46: Why not usig here the `N_SAMPLES` global variable you created? (or something similar)

test/test_pipeline/test_tabular_classification.py
```diff
@@ -53,7 +52,6 @@ def _assert_pipeline_search_space(self, pipeline, search_space_updates):
             elif isinstance(hyperparameter, CategoricalHyperparameter):
                 assert update.value_range == hyperparameter.choices
 
-    @flaky.flaky(max_runs=2)
```
franchuterivera, line 56: We are sampling a random configuration, wouldn't it make it fail?

test/test_pipeline/components/setup/test_setup_networks.py
nabenabe0928: There are mainly two reasons:
1. it cannot reflect the changes directly in `fit_dictionary_tabular` in conftest
2. some people change `fit_dictionary_tabular` in conftest without noticing the local modifications

if we can modify the configuration, i.e. in this case `fit_dictionary_tabular`, when some people just thought that once we change the `fit_dictionary_tabular` in conftest, it will be reflected to all the functions, they suffer from these local modifications. Since dict can be modified easily, it is not secured with respect to this matter and it potentially increases our works in the future.

test/test_pipeline/test_tabular_classification.py
```diff
@@ -442,3 +423,36 @@ def test_constant_pipeline_iris(fit_dictionary_tabular):
...
+    fit_dictionary_tabular_dummy['epochs'] = 50
+
+    pipeline = TabularClassificationPipeline(
+        dataset_properties=fit_dictionary_tabular_dummy['dataset_properties'])
+
+    cs = pipeline.get_hyperparameter_search_space()
+    config = cs.get_default_configuration()
+    pipeline.set_hyperparameters(config)
+
+    pipeline.fit(fit_dictionary_tabular_dummy)
```
franchuterivera, line 445: I Think we removed the network check from then API, I meant [this](https://github.com/automl/Auto-PyTorch/blob/8447e917303ec64a180aa2c62e25ea1b3609a284/test/test_api/test_api.py#L137)

Can we add it here?

franchuterivera (changes requested):
Thanks, this change is great to make the unit test less painful! Some minor changes requested

test/test_pipeline/components/setup/test_setup_networks.py
ravinkohli: I think the changes to `fit_dictionary_tabular` in any local test are based on a need for that test. Therefore, if in a test we notice something should be changed, it can be done so using a dictionary and for me, it does not cause any issues. Using a NamedTuple for `fit_dictionary` will cause the following issues:-
1. As tuples are immutable, to make a single change we'll have to change the whole tuple instance.
2. `fit dictionary` is anyways a `Dict`that is used throughout the pipeline so we'll have to always change the namedTuple to a dictionary for each test. 
3. If we still want a named tuple for `fit_dictionary`, that is a large refactor to the whole AutoPyTorch, which I don't think we need. Moreover, we are moving to fit_params API of sklearn as soon as possible anyways.

test/test_pipeline/components/training/base.py
ravinkohli: that is completely true, I'll change to that.

test/test_pipeline/test_tabular_classification.py
ravinkohli: for the fit, predict functions, I dont think it should fail. I'll add it for the score function if that's what you mean

test/test_pipeline/test_tabular_classification.py
ravinkohli: yes I think it makes sense. Thanks

test/test_pipeline/components/setup/test_setup_networks.py
nabenabe0928: Ah, I did not know that `fit_dictionary` is used in the main parts.
OK, then let's not change this time. But can we define this epoch inside the function to fetch this dictionary?
Currently the function `get_fit_dictinary` in `conftest.py` does not support kwargs modifications, so can we enable to pass the variables to modify?
```
def get_fit_dictionary(X, y, validator, backend, **kwargs):
    ...
    fit_dictionary = {
        ....
    }
    
    if set(kwargs.keys()) - set(fit_dictionary.keys()) >= 1:
        raise ValueError('Unknown keys were provided.')

    fit_dicitonary.update(kwargs)
    ...
    return fit_dictionary
```

Just in case, I will leave some comments here and probably post on the Issue page.
First, I checked where it is used and there were only 5 lines which use `fit_dictionary`.
- `abstract_evaluator.py` and `train_evaluator.py`
According to the implementations, you are right and we need to update the dictionary sometimes.
So we cannot change it to tuple.

test/test_pipeline/test_tabular_classification.py
ravinkohli: I just checked that the score function is using default config

test/test_pipeline/components/setup/test_setup_networks.py
nabenabe0928: I do not see why it is going to be a lot of work.
Because **kwargs is not a mandatory argument and you do not need to change any other codes where you do not need to change the epoch.
It means you need to change the function and the places where you need to change the epoch.
If I count correctly, it will be only 4- or 5-lines modifications.
Am I wrong here? Maybe I am missing some points.

About this time, you do not need this change, because test codes are not complicated compared to the main parts. 

========================== REVIEW part =========================-
However, It is more like general advice and since it is quite easy to change this time, I do not see any reasons why we do not change.
You know the variable range, right?
We should not change variables in wider ranges. 
When some issues happen, it is quite hard to know where the bugs are coming from.
In fact, it happened when I was inspecting the memory issues.
That is why I assume it is a common bad practice in this project and I need to persistently mention to condition what and why your codes matter when I find such bad practices.

test/test_pipeline/test_tabular_regression.py
```diff
@@ -279,23 +288,18 @@ def test_set_range_search_space_updates(self, fit_dictionary_tabular):
...
-    # lower the learning rate of the optimizer until seeding properly works
-    # with the default learning rate of 0.01 regression sometimes does not converge
     pipeline = TabularRegressionPipeline(
         dataset_properties=fit_dictionary_tabular_dummy['dataset_properties'],
-        search_space_updates=HyperparameterSearchSpaceUpdates([
-            HyperparameterSearchSpaceUpdate("optimizer",
-                                            "AdamOptimizer:lr",
-                                            value_range=[0.0001, 0.001],
-                                            default_value=0.001)
-        ])
```
nabenabe0928, line 298: Could you explain why we can remove it? (I do not know what happens when we remove it)

test/test_pipeline/test_tabular_regression.py
```diff
@@ -144,7 +150,10 @@ def test_default_configuration(self, fit_dictionary_tabular, is_small_preprocess
         pipeline = TabularRegressionPipeline(
             dataset_properties=fit_dictionary_tabular['dataset_properties'])
 
-        pipeline.fit(fit_dictionary_tabular)
+        with unittest.mock.patch.object(pipeline.named_steps['trainer'].choice, 'train_epoch') \
+                as patch_train:
+            patch_train.return_value = 1, {}
+            pipeline.fit(fit_dictionary_tabular)
```
nabenabe0928, line 156: What is the intention behind L153-155?

test/test_pipeline/test_tabular_classification.py
nabenabe0928: @ravinkohli You mean the result for this test is deterministic and we do not need to repeat anymore?
If it is a stochastic test, it is better to keep it for more success rate. (flaky does not do anything bad if we pass the test by ,let's say, 99%, right?)

ravinkohli: if you're talking about the fit and predict functions, I'd say they are quite deterministic.

test/test_pipeline/test_tabular_regression.py
ravinkohli: are we are testing if a pipeline can be fit, and we have a separate test for testing training, I replaced the training steps with a mock patch object to make it faster

test/test_pipeline/components/setup/test_setup_networks.py
ravinkohli: Firstly, it may be a simple modification. But I don't like the changes we'll have to make to make this work. To explain further, for every function that needs this epoch value to be changed, we'll have to pass the epochs in the pytest.mark.parameterize function  `@pytest.mark.parametrize("fit_dictionary_tabular_dummy", [("classification", 50)], indirect=True)`, see [here](https://github.com/ravinkohli/Auto-PyTorch/blob/44dec19a9f00ff2eb3618d2acaa2bd4691758ad5/test/test_pipeline/test_tabular_classification.py#L428). After this we'll have to change the function call [here](https://github.com/ravinkohli/Auto-PyTorch/blob/44dec19a9f00ff2eb3618d2acaa2bd4691758ad5/test/conftest.py#L315) to accept a tuple and use the first one of the tuple for checking if its classification or not and the second part to send the epochs to the `get_fit_dictionary` where we'll also have to check if epochs were given or not. I don't think all this is necessary just for a test. 
Secondly, I don't see how this will make memory debugging process any more complicated. Please let me know if you want to. 
Thirdly, for changing this dictionary to tuple, as I previously wrote, we plan on making a refactor that will get rid of this fit_dictionary entirely. We'll instead be using fit_params of sklearn, which also means that changing fit dictionary to a tuple is moot. 
Finally, as for the "bad practices in this project", we have tried to ensure that we have proper checks for coding style and static type checking, and where we felt necessary, use namedTuples or child classes of that instead of using dictionaries. If you still find places that are not optimal, please raise an issue about this and we'll try to resolve it.

test/test_pipeline/test_tabular_regression.py
nabenabe0928: I checked the corresponding parts (base_trainer.py and base_trainer_choice.py) and it seems no problem.
Thanks.

test/test_api/test_api.py
```diff
@@ -315,9 +417,7 @@ def test_tabular_regression(openml_name, resampling_strategy, backend):
         model = estimator._backend.load_cv_model_by_seed_and_id_and_budget(
             estimator.seed, successful_num_run, run_key.budget)
         assert isinstance(model, VotingRegressor)
-        assert len(model.estimators_) == 5
-        assert isinstance(model.estimators_[0].named_steps['network'].get_network(),
-                          torch.nn.Module)
+        assert len(model.estimators_) == 2
```
nabenabe0928, line 420: This is `num_splits`, right?
Can we avoid hard-coding?

test/test_api/test_api.py
```diff
@@ -234,21 +335,23 @@ def test_tabular_regression(openml_name, resampling_strategy, backend):
...
+            optimize_metric='r2',
+            total_walltime_limit=35,
+            func_eval_time_limit_secs=5,
+            enable_traditional_pipeline=False,
+        )
 
     # Internal dataset has expected settings
     assert estimator.dataset.task_type == 'tabular_regression'
-    expected_num_splits = 1 if resampling_strategy == HoldoutValTypes.holdout_validation else 5
+    expected_num_splits = 1 if resampling_strategy == HoldoutValTypes.holdout_validation else 2
```
nabenabe0928, line 354: 2 --> `num_splits`

test/test_api/test_api.py
```diff
@@ -200,17 +296,22 @@ def test_tabular_classification(openml_id, resampling_strategy, backend):
...
 @pytest.mark.parametrize('openml_name', ("boston", ))
-@pytest.mark.parametrize('resampling_strategy', (HoldoutValTypes.holdout_validation,
-                                                 CrossValTypes.k_fold_cross_validation,
-                                                 ))
-def test_tabular_regression(openml_name, resampling_strategy, backend):
+@unittest.mock.patch('autoPyTorch.evaluation.train_evaluator.eval_function',
+                     new=dummy_eval_function)
+@pytest.mark.parametrize('resampling_strategy,resampling_strategy_args',
+                         ((HoldoutValTypes.holdout_validation, None),
+                          (CrossValTypes.k_fold_cross_validation, {'num_splits': 2})
```
nabenabe0928, line 303: 2 --> `num_splits` (Solution: make a global variable)

test/test_api/test_api.py
```diff
@@ -145,9 +243,7 @@ def test_tabular_classification(openml_id, resampling_strategy, backend):
         model = estimator._backend.load_cv_model_by_seed_and_id_and_budget(
             estimator.seed, successful_num_run, run_key.budget)
         assert isinstance(model, VotingClassifier)
-        assert len(model.estimators_) == 5
-        assert isinstance(model.estimators_[0].named_steps['network'].get_network(),
-                          torch.nn.Module)
+        assert len(model.estimators_) == 2
```
nabenabe0928, line 246: 2 --> `num_splits`

test/test_api/test_api.py
```diff
@@ -59,21 +156,23 @@ def test_tabular_classification(openml_id, resampling_strategy, backend):
...
+            optimize_metric='accuracy',
+            total_walltime_limit=50,
+            func_eval_time_limit_secs=10,
+            enable_traditional_pipeline=False,
+        )
 
     # Internal dataset has expected settings
     assert estimator.dataset.task_type == 'tabular_classification'
-    expected_num_splits = 1 if resampling_strategy == HoldoutValTypes.holdout_validation else 5
+    expected_num_splits = 1 if resampling_strategy == HoldoutValTypes.holdout_validation else 2
```
nabenabe0928, line 175: 2 --> `num_splits`

test/test_api/test_api.py
```diff
@@ -18,35 +18,132 @@
...
+        disable_file_output,
+        pipeline_config=None,
+        budget_type=None,
+        init_params=None,
+        logger_port=None,
+        all_supported_metrics=True,
+        search_space_updates=None,
+        instance: str = None,
+) -> None:
+    evaluator = TrainEvaluator(
```
nabenabe0928, line 103: DummyTrainEvaluator?

nabenabe0928 (changes requested):

test/test_api/test_api.py
ravinkohli: good catch. Thanks, I'll change it.

test/test_api/test_api.py
ravinkohli: yes, makes sense. I'll change this

test/test_pipeline/test_tabular_regression.py
ravinkohli: I think previously it was changing the default lr to 0.001 from 0.01 (which is the default in the AdamOptimizer class). As the comment says, it was to lower the learning rate until seeding worked properly. And as we have merged PR #172 we don't need it anymore. The pipeline converges with just 0.01 and hence we don't need to reduce the learning rate which was also slowing down the test (not significantly)

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/feature_preprocessing/base_feature_preprocessor.py
```diff
@@ -8,7 +12,13 @@
 class autoPyTorchFeaturePreprocessingComponent(autoPyTorchTabularPreprocessingComponent):
     _required_properties: List[str] = ['handles_sparse']
 
-    def __init__(self) -> None:
+    def __init__(self, random_state: Optional[np.random.RandomState] = None) -> None:
```
nabenabe0928, line 15: Once you set any kwargs, we do not need `-> None` for `__init__` function.
There are some that have `-> None`, so could you remove them?

ravinkohli: That would involve changing a lot of files through out the project which I don't think is in the scope of this PR

nabenabe0928: I mean only part that comes up in your fix 
I didn't say everything

ravinkohli: Okay, I'll do that.

nabenabe0928 (approved):

PR #210, opened 2021-05-12
Fixing the scheduler to update for epochs and not batch iterations

ArlindKadra:
The scheduler was wrongly updating for batches instead of epochs. 
This, in turn, affected training and cocktail ingredients.


ravinkohli (approved):

ravinkohli:
linking the corresponding issue for reference #211

Issue #211, opened 2021-05-12
[bug] LR Scheduler updating at wrong times


ravinkohli:
Hey,
LR Schedulers dependent on their type should be updated either epoch-wise or batch-wise. Currently, they are only happening batch-wise which is incompatible with CosineAnnealingWarmRestarts, see [here](https://github.com/automl/Auto-PyTorch/blob/ee07c7ef8169fe70945d603b10a263272690397d/autoPyTorch/pipeline/components/training/trainer/base_trainer.py#L321). 

One possible solution to this,
make `{'step'='epochs'}` property of each scheduler, and use this to decide in the base trainer when to make a step. 

I think this is an urgent fix needed.

ArlindKadra:
yeah, I think quite probably the results of the new autopytorch might even be better than the old one now ;)
We should delegate the fix from the cocktail branch as fast as possible

ArlindKadra:
Especially considering the range of `T_0 `is from `0 to 20`

PR #212, opened 2021-05-14
[Hotfix] Adapt LR scheduler to epoch wise

nabenabe0928:
Address the [issue#211](https://github.com/automl/Auto-PyTorch/issues/211).

### DIFF
--- a/autoPyTorch/pipeline/components/training/trainer/MixUpTrainer.py
+++ b/autoPyTorch/pipeline/components/training/trainer/MixUpTrainer.py
@@ -1,4 +1,4 @@
-import typing
+from typing import Any, Callable, Dict, Optional, Tuple
 
 from ConfigSpace.configuration_space import ConfigurationSpace
 from ConfigSpace.hyperparameters import (
@@ -11,13 +11,16 @@ import numpy as np
 import torch
 
 from autoPyTorch.constants import CLASSIFICATION_TASKS, STRING_TO_TASK_TYPES
-from autoPyTorch.pipeline.components.training.trainer.base_trainer import BaseTrainerComponent
+from autoPyTorch.pipeline.components.training.trainer.base_trainer import (
+    BaseTrainerComponent,
+    _CriterionPreparationParameters
+)
 from autoPyTorch.utils.common import HyperparameterSearchSpace, add_hyperparameter
 
 
 class MixUpTrainer(BaseTrainerComponent):
     def __init__(self, alpha: float, weighted_loss: bool = False,
-                 random_state: typing.Optional[np.random.RandomState] = None):
+                 random_state: Optional[np.random.RandomState] = None):
         """
         This class handles the training of a network for a single given epoch.
 
@@ -29,8 +32,8 @@ class MixUpTrainer(BaseTrainerComponent):
         self.weighted_loss = weighted_loss
         self.alpha = alpha
 
-    def data_preparation(self, X: np.ndarray, y: np.ndarray,
-                         ) -> typing.Tuple[np.ndarray, typing.Dict[str, np.ndarray]]:
+    def data_preparation(self, X: torch.Tensor, y: torch.Tensor,
+                         ) -> Tuple[np.ndarray, _CriterionPreparationParameters]:
         """
         Depending on the trainer choice, data fed to the network might be pre-processed
         on a different way. That is, in standard training we provide the data to the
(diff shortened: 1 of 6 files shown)


autoPyTorch/pipeline/components/training/trainer/StandardTrainer.py
```diff
@@ -23,39 +28,42 @@ def __init__(self, weighted_loss: bool = False,
...
         return lambda criterion, pred: criterion(pred, y_a)
 
     @staticmethod
-    def get_properties(dataset_properties: typing.Optional[typing.Dict[str, typing.Any]] = None
-                       ) -> typing.Dict[str, str]:
+    def get_properties(dataset_properties: Optional[Dict[str, Any]] = None
+                       ) -> Dict[str, str]:
         return {
             'shortname': 'StandardTrainer',
             'name': 'StandardTrainer',
```
ravinkohli, line 61:
```suggestion
            'name': 'Standard Trainer',
```

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -185,22 +203,34 @@ def prepare(
         metrics_during_training: bool,
         scheduler: _LRScheduler,
         task_type: int,
-        labels: Union[np.ndarray, torch.Tensor, pd.DataFrame]
+        labels: Union[np.ndarray, torch.Tensor, pd.DataFrame],
+        step_unit: Union[str, StepIntervalUnit] = StepIntervalUnit.batch
     ) -> None:
 
+        step_unit_types = [str, StepIntervalUnit]
+        if not any(isinstance(step_unit, step_unit_type) for step_unit_type in step_unit_types):
```
ravinkohli, line 211:
```suggestion
        if isinstance(step_unit, step_unit_types):
```
you can check if an object is in an iterable of types directly.

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -185,22 +203,34 @@ def prepare(
...
+            raise ValueError(f'step_unit must be {step_unit_types}, but got {type(step_unit)}.')
```
ravinkohli, line 212:
```suggestion
            raise ValueError(f'step_unit must be one of {step_unit_types}, but got {type(step_unit)}.')
```

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -185,22 +203,34 @@ def prepare(
...
+
         # Save the device to be used
         self.device = device
 
         # Setup the metrics
         self.metrics = metrics
 
         # Weights for the loss function
-        kwargs = {}
+        weight_dict = {}
```
ravinkohli, line 228: can we use weight_params? I think that would explain it better.

ravinkohli (changes requested):
Thank you for this PR. I have made a few suggestions. Also, do you think in the scheduler tests, we can check if the step of the scheduler is taken at the expected time. However, I could not find where the scheduler is passing its step interval anywhere? What am I missing?

nabenabe0928:
> I have made a few suggestions. 

Thanks, I will reflect those changes:)

> Also, do you think in the scheduler tests, we can check if the step of the scheduler is taken at the expected time. 

Do you mean we want another test for it? (Then I will do it.)

> However, I could not find where the scheduler is passing its step interval anywhere? What am I missing?

Since everything is set in `set_training_params`, I put it there.

autoPyTorch/pipeline/components/training/trainer/StandardTrainer.py
nabenabe0928: Why do we want space here?

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
nabenabe0928: Ok, I will change to the following:
```
if not isinstance(step_unit, step_unit_types):
```

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
nabenabe0928: I will change it to `either`.

autoPyTorch/pipeline/components/training/trainer/StandardTrainer.py
ravinkohli: its not important, but in general we have followed that short name is either acronym of the whole name or the whole name without spaces. You can definitely ignore this but I suggested for uniformity

ravinkohli:
> > I have made a few suggestions.
> 
> Thanks, I will reflect those changes:)
> 
> > Also, do you think in the scheduler tests, we can check if the step of the scheduler is taken at the expected time.
> 
> Do you mean we want another test for it? (Then I will do it.)
> 
> > However, I could not find where the scheduler is passing its step interval anywhere? What am I missing?
> 
> Since everything is set in `set_training_params`, I put it there.
but how is the scheduler telling the base trainer that it should be updated epoch or batch wise? I see that by default you are passing `batch` but its not connected to the scheduler right? For this to work, we should send this info via base_trainer_choice, which will pass it via `step_unit` of the `set_training_params`.

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -236,11 +273,25 @@ def on_epoch_end(self, X: Dict[str, Any], epoch: int) -> bool:
...
+    def _scheduler_step(
+        self,
+        step_interval: StepIntervalUnit,
+        loss: Optional[torch.Tensor] = None
+    ) -> None:
+        if self.step_unit != step_interval:
+            return
+
+        if self.scheduler:
+            if 'ReduceLROnPlateau' in self.scheduler.__class__.__name__:
```
franchuterivera, line 285: I think it will be better not to hardcode things in this, and let every scheduler decide what type of step it should take?

autoPyTorch/pipeline/components/training/trainer/MixUpTrainer.py
```diff
@@ -29,45 +32,50 @@ def __init__(self, alpha: float, weighted_loss: bool = False,
...
 
         mixed_x = lam * X + (1 - lam) * X[index, :]
         y_a, y_b = y, y[index]
-        return mixed_x, {'y_a': y_a, 'y_b': y_b, 'lam': lam}
-
-    def criterion_preparation(self, y_a: np.ndarray, y_b: np.ndarray = None, lam: float = 1.0
-                              ) -> typing.Callable:
+        return mixed_x, _NewLossParameters(y_a=y_a, y_b=y_b, lam=lam)
+
+    def _get_new_loss_fn(
```
franchuterivera, line 59: Can we try to make this in another PR? I think this does not directly helps to solve the `LR schedule`

franchuterivera:
Thanks a lot for the changes, for example, there are places where you coded certain pieces in a better way! Can these changes be done on a separate PR? I think this will be checked and approved right away as they are small format changes.

However, they make the actual change quite complicated to surface -- I also do not see right away how every scheduler will define whether the step should be done per epoch or per batch? 

Do you think we can also add unit testing to prevent this from happening again?

autoPyTorch/pipeline/components/training/trainer/StandardTrainer.py
nabenabe0928: Ok, I do not know where we are using those names, but if it is convention, I will follow this rule.

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
nabenabe0928: Yeah, I know what you mean, but it was the original code and I cannot guarantee the original behavior by changing here, so I would like to keep this for now.
Of course, if you have any good idea to solve this, I may adapt here immediately.

autoPyTorch/pipeline/components/training/trainer/MixUpTrainer.py
nabenabe0928: The renaming? Ok, I will do it in another PR.

ravinkohli:
As we have solved this issue with #256, I am closing this PR

PR #223, opened 2021-05-18
[ADD] Forkserver as default multiprocessing strategy

franchuterivera:
There is a big number of files changed for this enhancement, and first I would like to justify the change and then explain why so many files got modified:

**What we win with this PR**
- This PR includes two changes: enabling forking server strategy and no dask local-cluster if we are using just a single core.
- By enabling forkserver and only using dask only when needed we significantly reduce the chance of deadlock.
- Additionally, we are able to render experiments on our web page more nicely
- We align AutoML multiprocessing context, i.e., this is the proven recipe that works (by AutoSklean)

**Why so many changes**
It was hard to debug why forkserver did not work for autopytorch initially, and it was because every pipeline dynamically fills its steps, and with forkserver this was not happening. 
For example, encoding step has 2 choices (one-hot-encoder and no encoder), yet this step was left empty. The reason for the dynamic population of the choices of steps not triggering is that forkserver forces a re-initialization of the modules causing this [line](https://github.com/automl/Auto-PyTorch/blob/4493270fce6e450fa9bf11d4668f2b63c9a4d8e0/autoPyTorch/pipeline/components/base_component.py#L41) not to add the components. This is not the case, if the components are part of the `__init__.py`. I have to rename the choices from their original name to be `__init__.py` and that is why so many files were modified.


autoPyTorch/pipeline/components/preprocessing/image_preprocessing/normalise/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+_addons = ThirdPartyComponents(BaseNormalizer)
+
+
+def add_normalizer(normalizer: BaseNormalizer) -> None:
+    _addons.add_component(normalizer)
+
+
+class NormalizerChoice(autoPyTorchChoice):
+    """
+    Allows for dynamically choosing encoding component at runtime
```
ravinkohli, line 31:
```suggestion
    Allows for dynamically choosing normalizer component at runtime
```

autoPyTorch/pipeline/components/preprocessing/image_preprocessing/normalise/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+
+    def get_components(self) -> Dict[str, autoPyTorchComponent]:
+        """Returns the available normalizer components
+
+        Args:
+            None
+
+        Returns:
+            Dict[str, autoPyTorchComponent]: all BaseNormalise components available
```
ravinkohli, line 41:
```suggestion
            Dict[str, autoPyTorchComponent]: all BaseNormalizer components available
```

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/encoding/__init__.py
```diff
@@ -0,0 +1,137 @@
...
+                                                             ['NoEncoder'],
+                                                             default_value=default)
+            else:
+                preprocessor = CSH.CategoricalHyperparameter('__choice__',
+                                                             list(available_preprocessors.keys()),
+                                                             default_value=default)
+
+        cs.add_hyperparameter(preprocessor)
+
+        # add only child hyperparameters of early_preprocessor choices
```
ravinkohli, line 112:
```suggestion
        # add only child hyperparameters of preprocessor choices
```

autoPyTorch/pipeline/components/preprocessing/image_preprocessing/normalise/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+            preprocessor = CSH.CategoricalHyperparameter('__choice__',
+                                                         choice_hyperparameter.value_range,
+                                                         default_value=choice_hyperparameter.default_value)
+        else:
+            preprocessor = CSH.CategoricalHyperparameter('__choice__',
+                                                         list(available_preprocessors.keys()),
+                                                         default_value=default)
+        cs.add_hyperparameter(preprocessor)
+
+        # add only child hyperparameters of early_preprocessor choices
```
ravinkohli, line 96:
```suggestion
        # add only child hyperparameters of preprocessor choices
```

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/feature_preprocessing/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+                # Truncated SVD requires n_features > n_components
+                if len(dataset_properties['numerical_columns']) == 1:
+                    del available_['TruncatedSVD']
+                preprocessor = CSH.CategoricalHyperparameter('__choice__',
+                                                             list(available_.keys()),
+                                                             default_value=default)
+
+        cs.add_hyperparameter(preprocessor)
+
+        # add only child hyperparameters of early_preprocessor choices
```
ravinkohli, line 121:
```suggestion
        # add only child hyperparameters of preprocessor choices
```

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/scaling/__init__.py
```diff
@@ -0,0 +1,131 @@
...
+                preprocessor = CSH.CategoricalHyperparameter('__choice__',
+                                                             ['NoScaler'],
+                                                             default_value=default)
+            else:
+                preprocessor = CSH.CategoricalHyperparameter('__choice__',
+                                                             list(available_scalers.keys()),
+                                                             default_value=default)
+        cs.add_hyperparameter(preprocessor)
+
+        # add only child hyperparameters of early_preprocessor choices
```
ravinkohli, line 106:
```suggestion
        # add only child hyperparameters of preprocessor choices
```

autoPyTorch/pipeline/components/setup/network_embedding/__init__.py
```diff
@@ -0,0 +1,204 @@
...
+class NetworkEmbeddingChoice(autoPyTorchChoice):
+
+    def get_components(self) -> Dict[str, autoPyTorchComponent]:
+        """Returns the available embedding components
+
+        Args:
+            None
+
+        Returns:
+            Dict[str, autoPyTorchComponent]: all baseembedding components available
```
ravinkohli, line 40:
```suggestion
            Dict[str, autoPyTorchComponent]: all NetworkEmbeddingComponent components available
```

autoPyTorch/pipeline/components/setup/network_initializer/__init__.py
```diff
@@ -0,0 +1,179 @@
...
+class NetworkInitializerChoice(autoPyTorchChoice):
+
+    def get_components(self) -> Dict[str, autoPyTorchComponent]:
+        """Returns the available initializer components
+
+        Args:
+            None
+
+        Returns:
+            Dict[str, autoPyTorchComponent]: all BaseInitializerComponent components available
```
ravinkohli, line 40:
```suggestion
            Dict[str, autoPyTorchComponent]: all BaseNetworkInitializerComponent components available
```

autoPyTorch/utils/single_thread_client.py
```diff
@@ -0,0 +1,86 @@
+import typing
+from pathlib import Path
+
+import dask.distributed
+
+
+class DummyFuture(dask.distributed.Future):
+    """
+    A class that mimics a distributed Future, the outcome of
```
ravinkohli, line 9: Could you explain a bit more about why we need this? maybe in general what this whole file is doing?

test/test_utils/test_single_thread_client.py
```diff
@@ -0,0 +1,27 @@
...
+    assert sum(single_thread_client.nthreads().values()) == 1
+    single_thread_client.close()
+    single_thread_client.shutdown()
+
+    # Client/Futures are printed, so make sure str works
+    # str calls __rpr__ which is the purpose of below check
+    assert str(future) != ""
+    assert str(single_thread_client) != ""
+
+    with pytest.raises(NotImplementedError):
```
ravinkohli, line 26: could you add a comment to what this assertion is ensuring?

ravinkohli (changes requested):
Hey, thanks a lot for this PR. This will seriously simplify things when we are not working with a multi process environment. I have left a few comments, some are suggestions to fix the comments, and some are asking for clarifications. As they are resolved I think this PR can be merged.

autoPyTorch/pipeline/components/preprocessing/image_preprocessing/normalise/__init__.py
```diff
@@ -0,0 +1,106 @@
+import os
```
nabenabe0928, line 1: The series of changes you made should be achieved by writing the following in each `__init__.py`:
```
from autoPyTorch.pipeline.components.<something 1>.<something 2>.<something 3> import base_<something 4>_choice

<something 5>Choice = base_<something 4>_choice.<something 5>Choice
```

If there are any intentions behind the changes, let me know the reasons as a response here.

autoPyTorch/pipeline/image_classification.py
```diff
@@ -8,7 +8,7 @@
 
 from autoPyTorch.pipeline.base_pipeline import BasePipeline
 from autoPyTorch.pipeline.components.base_choice import autoPyTorchChoice
-from autoPyTorch.pipeline.components.preprocessing.image_preprocessing.normalise.base_normalizer_choice import (
```
nabenabe0928, line 11: Those paths are identical either ways (my suggestion and your changes.)

nabenabe0928 (changes requested):

autoPyTorch/api/base_task.py
```diff
@@ -187,7 +189,16 @@ def __init__(
...
+        # Single core, local runs should use fork
+        # to prevent the __main__ requirements in
+        # examples. Nevertheless, multi-process runs
+        # have spawn as requirement to reduce the
+        # possibility of a deadlock
         self._dask_client = None
+        self._multiprocessing_context = 'forkserver'
+        if self.n_jobs == 1:
+            self._multiprocessing_context = 'fork'
+            self._dask_client = SingleThreadedClient()
```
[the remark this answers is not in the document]
franchuterivera: This single threaded client aims to do this, while still providing an interface similar to dask client.

The problem is that many places in the code assume we are working with futures (because usually we are, as usually we run with more than 1 core). If running with 1 core, we do not want to use dask but also we do not want to add an `if` statement everywhere in smac/auto-sklearn/autopytorch. So this seems like a good tradeoff.

What this gives you is no new process (single thread) with the same dask interface.

autoPyTorch/pipeline/components/preprocessing/image_preprocessing/normalise/__init__.py
franchuterivera: I wrote this in the PR description, but let me state it here again. The problem is that forkserver, when communicating the pre-loaded modules from the parent process to the forked server client (the entity that spawns new process) seems to require the `__init__.py`has the choices defined on it. I made this change to mimic how the pipeline is implemented in auto-sklearn.

Defining the choice in the init is not sufficient, as the problem is how python organized the loading of the modules. Yes, maybe we could do some intelligent sourcing the choices via relative paths, but the criteria that I follow was to be similar to auto-sklearn so that we share both the same code (and solution to problems).

franchuterivera:
Hi @nabenabe0928 please let me know if you have further questions on this PR.

autoPyTorch/pipeline/components/preprocessing/image_preprocessing/normalise/__init__.py
nabenabe0928: > I made this change to mimic how the pipeline is implemented in auto-sklearn.

It is not a good practice to write many codes in `__init__.py`.
We should mimic good practices, but not bad practices. 

> Defining the choice in the init is not sufficient

Could you give me the error you got when you run?
I do not get any errors when I modify the codes accordingly.

autoPyTorch/api/base_task.py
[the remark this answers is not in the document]
nabenabe0928: Ok, I will check after a while and if there are some other issues, I will fix this part at this moment.

ravinkohli (approved):
This PR can be merged but I think there are some merge conflicts.

PR #224, opened 2021-05-18
[ADD] Coverage calculation

franchuterivera:
This PR enables coverage calculation in pytest. Additionally, it adds further tests to reach 85% coverage.

Some files that now reside in automl_common have very little coverage. Will address those problems there.

Low coverage files pending:

- [ ] Increase base task coverage!
- [x] TAE testing improvements
- [x] traditional pipeline testing
- [x] search_space testing


test/test_pipeline/components/training/test_image_data_loader.py
```diff
@@ -8,21 +8,28 @@
...
 
-        compose = loader.build_transform(fit_dictionary, mode='train')
+    compose = loader.build_transform(fit_dictionary, mode='train')
 
-        self.assertIsInstance(compose, torchvision.transforms.Compose)
+    assert isinstance(compose, torchvision.transforms.Compose)
 
-        # We expect to tensor and image augmenter
-        self.assertEqual(len(compose.transforms), 2)
+    # We expect to tensor and image augmenter
```
ravinkohli, line 27:
```suggestion
    # We expect there to be two transforms, to tensor and image augmenter
```

ravinkohli (approved):
hey, thanks for this PR. It introduces more robust testing. I have made just a small suggestion which is also not too important. Overall, this PR can be merged as-is in my opinion

franchuterivera:
My current guess about the page not showing is that we did not have a previous run. You would see the differential, but there is no baseline. In other words, after this PR is merged, future pushes should see it. Else, I will debug why that is not the case.

PR #233, opened 2021-05-21
Fix bugs in cutout training

ravinkohli:
This PR fixes the following bugs:

1. we were not sampling the indices with replacement.
2. we were taking the minimum between batch size and num features, which is not necessary
3. Numerical features were getting -1 to them, whereas 0 makes more sense

### DIFF
--- a/autoPyTorch/pipeline/components/training/trainer/RowCutMixTrainer.py
+++ b/autoPyTorch/pipeline/components/training/trainer/RowCutMixTrainer.py
@@ -39,7 +39,8 @@ class RowCutMixTrainer(MixUp, BaseTrainerComponent):
         # It is unlikely that the batch size is lower than the number of features, but
         # be safe
         size = min(X.shape[0], X.shape[1])
-        indices = torch.tensor(self.random_state.choice(range(1, size), max(1, np.int(size * lam))))
+        indices = torch.tensor(self.random_state.choice(range(1, size), max(1, np.int32(size * lam)),
+                                                        replace=False))
 
         X[:, indices] = X[index, :][:, indices]
 
--- a/autoPyTorch/pipeline/components/training/trainer/RowCutMixTrainer.py
+++ b/autoPyTorch/pipeline/components/training/trainer/RowCutMixTrainer.py
@@ -35,10 +35,7 @@ class RowCutMixTrainer(MixUp, BaseTrainerComponent):
         if beta <= 0 or r > self.alpha:
             return X, {'y_a': y, 'y_b': y[index], 'lam': 1}
 
-        # The mixup component mixes up also on the batch dimension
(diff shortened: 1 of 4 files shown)


autoPyTorch/pipeline/components/training/trainer/RowCutMixTrainer.py
```diff
@@ -39,7 +39,8 @@ def data_preparation(self, X: np.ndarray, y: np.ndarray,
         # It is unlikely that the batch size is lower than the number of features, but
         # be safe
         size = min(X.shape[0], X.shape[1])
```
ArlindKadra, line 41: This should also be changed to size=X.shape[1] right ?

ravinkohli: True, I missed that

autoPyTorch/pipeline/components/training/trainer/RowCutOutTrainer.py
```diff
@@ -34,17 +36,26 @@ def data_preparation(self, X: np.ndarray, y: np.ndarray,
...
-        # It is unlikely that the batch size is lower than the number of features, but
-        # be safe
-        size = min(X.shape[0], X.shape[1])
-        indices = torch.tensor(self.random_state.choice(range(1, size), max(1, np.int(size * self.patch_ratio))))
+        size = X.shape[1]
+        indices = self.random_state.choice(range(1, size), max(1, np.int32(size * self.patch_ratio)),
+                                           replace=False)
 
-        # We use an ordinal encoder on the tabular data
+        if not isinstance(self.numerical_columns, typing.Iterable):
```
ArlindKadra, line 43: What if the numerical columns are None, we should still continue with only categorical imputing in this case or not.

ArlindKadra: Also if there are only numerical columns, there should not be a conversion for categorical ones.

ravinkohli: Actually when there are no numerical columns, it is not none but it is an empty list. And indexing with an empty list does not affect the tensor so this should work

ArlindKadra: https://github.com/automl/Auto-PyTorch/blob/8b71ee2ad3c00277928ac462c0d364151d69aca0/autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py#L340-L341

Is 'numerical_columns' always in 'dataset_properties' ?

ravinkohli: when its tabular data then yeah

Issue #227, opened 2021-05-21
Add cronjobs


ravinkohli:
We should schedule cronjobs like [this](https://github.com/automl/auto-sklearn/blob/52dd9a1747152d5494812d81dd32b7519ba790e5/.github/workflows/pytest.yml#L8) to ensure that external changes don't break autopytorch

PR #234, opened 2021-05-21
[ADD] Pytest schedule

ravinkohli:
This PR adds a schedule to run the pytest workflow every Monday morning at 7 am UTC. This will allow us to ensure that any changes to the dependencies do not break autopytorch.

Addresses #227

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -32,8 +32,9 @@ jobs:
         echo "::set-output name=BEFORE::$(git status --porcelain -b)"
     - name: Run tests
       run: |
-        if [ ${{ matrix.code-cov }} ]; then codecov='--cov=autoPyTorch --cov-report=xml --cov-config=.coveragerc
-        --include="autoPyTorch/*"'; fi
+        if [ ${{ matrix.code-cov }} ]; then
+          codecov='--cov=autoPyTorch --cov-report=xml --cov-config=.coveragerc --include="autoPyTorch/*"';
+        fi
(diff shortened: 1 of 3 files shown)


franchuterivera (changes requested):
Thanks, with this every Monday we will run a scheduled pytest to detect if an external dependency broke the code.

But we want to make sure we checkout the development branch. One can pass a [ref](https://github.com/actions/checkout) to the [checkout](https://github.com/automl/auto-sklearn/blob/52dd9a1747152d5494812d81dd32b7519ba790e5/.github/workflows/pytest.yml#L28) command but we have to make it conditional on the schedule Alternatively we can create another file with just the schedule.

My point is, that I think the master branch will be checked out.

.github/workflows/pytest.yml
```diff
@@ -20,6 +20,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        ref: development
```
franchuterivera, line 24: The problem if you do this is that every PR onwards will checkout from development :)

franchuterivera: Yeah this ref development might work but if we create a parallel file? It will be duplicated code with `pytest.yml` but I cannot think of another way. 

So maybe let us only create a very short schedule.yml that just install with ref development and run pytest on 3.8 python?

ravinkohli: sure

nabenabe0928:
Thanks for the improvement, I do not see flaws now

franchuterivera (approved):

Issue #229, opened 2021-05-21
Create distribution of autopytorch


ravinkohli:
We should create a distribution for autopytorch like [this](https://github.com/automl/auto-sklearn/blob/master/.github/workflows/dist.yml)

PR #238, opened 2021-05-24
Add dist check worflow

ravinkohli:
Addresses #229


franchuterivera (approved):
thanks for the PR, I see this working [here](https://github.com/ravinkohli/Auto-PyTorch/runs/2655075049?check_suite_focus=true)

Issue #230, opened 2021-05-21
Stricter flake and mypy checks


ravinkohli:
There are a few `if` checks in the code that don't make sense. By making our checks stricter using`mypy --warn-unreachable`. Along with this, there are plenty of other checks that can help improve the quality of our code. Check [this](https://github.com/automl/automl_common/blob/93984e17464a2e9078d3b415528f06784515001e/.pre-commit-config.yaml#L6) for more info.

PR #240, opened 2021-05-25
[ADD] Stricter checks mypy

ravinkohli:
This PR partially addresses #230. 

1. Added mypy checks for warn-redundant-casts, warn-return-any, warn-unreachable,


nabenabe0928 (commented):
Thanks for the PR, I will put some good practices for mypy typing:

1. class rather than `Dict[str, Any]`
Because `Dict[str, Any]` is not supported by most languages, because of lower safety.

2. sometimes it is effective to declare typing variables as a global variable to comprehend inputs as a same group.

For python3.8 >=, 
3. Generic class
We can use something like `template` in C++
We can separate `Union` depending on files.

4. Literal
we can accept only specific variable rather than typing

5. Final
If we do not hope the change in variables, we can use it.

6. TypedDict
We can define meta class for dict. 

Also we need to discuss (with Frank as well) when we should stop the support of older python versions.
Just in case, python3.6 is already not supported officially and python3.7 will be not supported from this December.

autoPyTorch/pipeline/base_pipeline.py
```diff
@@ -152,18 +160,6 @@ def fit_estimator(self, X: Dict[str, Any],
...
-            return self._final_estimator.get_max_iter()
-        else:
-            raise NotImplementedError()
-
-    def configuration_fully_fitted(self) -> bool:
-        return self._final_estimator.configuration_fully_fitted()
-
-    def get_current_iter(self) -> int:
-        return self._final_estimator.get_current_iter()
-
```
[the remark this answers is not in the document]
ravinkohli: nah its an artifact of the code we used from autosklearn

autoPyTorch/pipeline/components/base_choice.py
```diff
@@ -222,7 +222,8 @@ def transform(self, X: Dict[str, Any]) -> Dict[str, Any]:
             (Dict[str, Any])
         """
         assert self.choice is not None, "Can not call transform without initialising the component"
-        return self.choice.transform(X)
+        X = self.choice.transform(X)
```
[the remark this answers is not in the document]
ravinkohli: but thats what we want write? the choice transform is only going to add stuff to the fit dictionary. I don't see where overwriting can cause issues

autoPyTorch/pipeline/tabular_regression.py
```diff
@@ -173,16 +172,19 @@ def _get_hyperparameter_search_space(self,
...
         The step itself has choices given via autoPyTorchChoice.
 
         Returns:
-            List[Tuple[str, autoPyTorchChoice]]: list of steps sequentially exercised
-                by the pipeline.
+            List[Tuple[str, Union[autoPyTorchComponent, autoPyTorchChoice]]]:
+                list of steps sequentially exercised by the pipeline.
         """
-        steps = []  # type: List[Tuple[str, autoPyTorchChoice]]
+        steps = []  # type: List[Tuple[str, Union[autoPyTorchComponent, autoPyTorchChoice]]]
```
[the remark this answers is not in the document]
ravinkohli: sure

autoPyTorch/pipeline/base_pipeline.py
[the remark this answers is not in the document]
nabenabe0928: Ok, I will close this conversation!

.pre-commit-config.yaml
```diff
@@ -3,21 +3,22 @@ repos:
...
         additional_dependencies:
           - flake8-print==3.1.4
           - flake8-import-order
+        name: flake8 autoPyTorch
+        files: autoPyTorch/.*
       - id: flake8
-        name: flake8 tests
-        files: test/.*
-        additional_dependencies:
-          - flake8-print==3.1.4
```
franchuterivera, line 22: why do you want to remove these additional dependencies?

autoPyTorch/api/base_task.py
```diff
@@ -1316,15 +1317,17 @@ def get_models_with_weights(self) -> List:
             self._load_models()
 
         assert self.ensemble_ is not None
-        return self.ensemble_.get_models_with_weights(self.models_)
+        models_with_weights: List = self.ensemble_.get_models_with_weights(self.models_)
```
franchuterivera, line 1320: This change I do not full understand :)... shouldn't the get_models_with_weights return a list?

If this has to happen, can you do `List[float, Pipeline]` ?

autoPyTorch/ensemble/ensemble_builder.py
```diff
@@ -274,7 +274,7 @@ def fit_and_return_ensemble(
     ensemble_nbest: int,
     max_models_on_disc: Union[float, int],
     seed: int,
-    precision: int,
+    precision: Union[int, str],
```
franchuterivera, line 277: I think we should only use int. It happened in autosklearn that we were comparing `'128' == 128`  which is false, and had unexpected behavior.

autoPyTorch/pipeline/traditional_tabular_classification.py
```diff
@@ -147,7 +147,7 @@ def _get_hyperparameter_search_space(
                 to honor when creating the configuration space
             exclude (Optional[Dict[str, Any]]): what hyper-parameter configurations
                 to remove from the configuration space
-            dataset_properties (Optional[Dict[str, Union[str, int]]]): Characteristics
+            dataset_properties (Optional[Dict[str, Any]]): Characteristics
```
franchuterivera, line 150: Can this really be any? Should we define a DATASET_PROPERTIES_TYPE in the dataset file?

Then we can specify the valid types which I think are List/int/float/str?

franchuterivera (changes requested):
Thanks for the PR. I left some questions. Let me know what you think about them.

autoPyTorch/pipeline/components/base_choice.py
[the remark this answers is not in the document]
nabenabe0928: The problem is both `X` and `X_transform` can remain on the user-side.
It means if they try to use `X` after using this function, unexpected behaviors will happen.

To illustrate what I mean, I will use an easy example.
```
In [1]: import numpy as np
   ...: 
   ...: 
   ...: def my_transform(a):
   ...:     a[0] = 100
   ...:     return a
   ...: 
   ...: 
   ...: a = np.random.randint(40, size=2)
   ...: print(a)
   ...: a_transform = my_transform(a)
   ...: print(a, a_transform)
   ...: 
   ...: a[1] = 300
   ...: print(a, a_transform)
[26  4]
[100   4] [100   4]
[100 300] [100 300]
```

In this case, as you can see, even though `a` was just modified, the changes propagate to the `a_transform` as well.
It means the ghost variable `a` might cause uncalled errors.
In other words, there are no benefits by leaving in this case.
Such problems might happen in your code as well.

autoPyTorch/ensemble/ensemble_builder.py
nabenabe0928: I agree with you.
Preferably, we would like to use `typing.Literal`, but it is not supported python<=3.7, so typing like `np.int64` also works better.
Anyways, we only have `32` or `64`, right? Then it is better to define by constant variables.

autoPyTorch/pipeline/components/base_choice.py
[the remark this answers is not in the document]
ravinkohli: I definitely agree with the problem you are mentioning. However, after a transform, we only care about a_transform in our case, and the old X is definitely not needed and its actually overwritten. I also agree we should fix this, but that will be better done with a new implementation for the fit dictionary, which is in our plans as we move towards sklearn 0.24 API once we are satisfied with the performance of the library. If you would like to, you can raise an issue and that can be referenced once we update our code.

.pre-commit-config.yaml
ravinkohli: I must have done it by mistake, I'll revert

autoPyTorch/pipeline/components/base_choice.py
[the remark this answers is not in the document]
nabenabe0928: OK, it means we will handle it later or we deprecate `fit_dictionary`?
If former, could you add a comment line saying `# This it not a destructive method now`
If latter, just resolve this comment.

autoPyTorch/evaluation/tae.py
```diff
@@ -34,7 +34,7 @@
 
 def fit_predict_try_except_decorator(
         ta: typing.Callable,
-        queue: multiprocessing.Queue, cost_for_crash: float, **kwargs: typing.Any) -> None:
+        queue: multiprocessing.Queue, cost_for_crash: float, **kwargs: typing.Any) -> typing.Optional[typing.Any]:
```
nabenabe0928, line 37: Because we do not use any return in the code (we take the results from pynisher), we do not need to return.
So I think the following is better.
```
try:
    ta(...)
except:
    ...
```

autoPyTorch/pipeline/components/base_component.py
```diff
@@ -127,14 +127,11 @@ def get_fit_requirements(self) -> Optional[List[FitRequirement]]:
...
-        if self._fit_requirements is not None:
-            self._fit_requirements.extend(requirements)
-        else:
-            self._fit_requirements = requirements
+        self._fit_requirements.extend(requirements)
 
     @staticmethod
     def get_properties(dataset_properties: Optional[Dict[str, str]] = None
-                       ) -> Dict[str, Any]:
+                       ) -> Dict[str, Union[str, bool]]:
```
nabenabe0928, line 134: Could you give an example what could be the bool value for this dict in the doc string?

autoPyTorch/pipeline/image_classification.py
```diff
@@ -134,7 +136,9 @@ def _get_hyperparameter_search_space(self,
         """
         cs = ConfigurationSpace()
 
-        if dataset_properties is None or not isinstance(dataset_properties, dict):
+        if not isinstance(dataset_properties, dict):
+            warnings.warn('The given dataset_properties argument contains an illegal value.'
```
nabenabe0928: `illegal value` is a common usage in python?
I have never seen `illegal value`, but see often `invalid value`.

Plus, `The given dataset_properties argument contains ..` sounds we have weird values inside the dict rather than `dataset_properties` is not a valid type.
So I think `dataset_properties must be dict` is easier to understand the warning.

autoPyTorch/pipeline/tabular_classification.py
```diff
@@ -185,10 +185,9 @@ def _get_hyperparameter_search_space(self,
         """
         cs = ConfigurationSpace()
 
-        if dataset_properties is None or not isinstance(dataset_properties, dict):
-            if not isinstance(dataset_properties, dict):
-                warnings.warn('The given dataset_properties argument contains an illegal value.'
-                              'Proceeding with the default value')
+        if not isinstance(dataset_properties, dict):
+            warnings.warn('The given dataset_properties argument contains an illegal value.'
```
nabenabe0928, line 189: illegal -> invalid

autoPyTorch/pipeline/tabular_regression.py
```diff
@@ -130,10 +130,9 @@ def _get_hyperparameter_search_space(self,
         """
         cs = ConfigurationSpace()
 
-        if dataset_properties is None or not isinstance(dataset_properties, dict):
-            if not isinstance(dataset_properties, dict):
-                warnings.warn('The given dataset_properties argument contains an illegal value.'
-                              'Proceeding with the default value')
+        if not isinstance(dataset_properties, dict):
+            warnings.warn('The given dataset_properties argument contains an illegal value.'
```
nabenabe0928, line 134: illegal -> invalid

autoPyTorch/pipeline/traditional_tabular_classification.py
```diff
@@ -156,10 +157,9 @@ def _get_hyperparameter_search_space(
         """
         cs = ConfigurationSpace()
 
-        if dataset_properties is None or not isinstance(dataset_properties, dict):
-            if not isinstance(dataset_properties, dict):
-                warnings.warn('The given dataset_properties argument contains an illegal value.'
-                              'Proceeding with the default value')
+        if not isinstance(dataset_properties, dict):
+            warnings.warn('The given dataset_properties argument contains an illegal value.'
```
nabenabe0928, line 160: illegal -> invalid

autoPyTorch/pipeline/base_pipeline.py
```diff
@@ -52,7 +55,7 @@ class BasePipeline(Pipeline):
 
 
     Attributes:
-        steps (List[Tuple[str, Union[autoPyTorch, autoPyTorchChoice]]]):
+        steps (List[Tuple[str, PipelineStepType]]):
             the steps of the current pipeline
```
nabenabe0928, line 59: Could you add what the step means?

autoPyTorch/api/base_task.py
ravinkohli: yes but mypy was raising a no-any-return error when I had no annotation but I have added your suggestion

autoPyTorch/ensemble/ensemble_builder.py
ravinkohli: okay I'll go with int then

autoPyTorch/pipeline/traditional_tabular_classification.py
ravinkohli: Yes this will be better.

autoPyTorch/evaluation/tae.py
ravinkohli: True, I'll change this

autoPyTorch/pipeline/components/base_component.py
ravinkohli: see [this](https://github.com/automl/Auto-PyTorch/blob/1818445d25415d9396ccd10fd7057dc315666418/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/encoding/OneHotEncoder.py#L35)

autoPyTorch/data/tabular_target_validator.py
```diff
@@ -194,8 +194,8 @@ def _check_data(
                 A set of features whose dimensionality and data type is going to be checked
         """
 
-        if not isinstance(y, (np.ndarray, pd.DataFrame, typing.List, pd.Series)) and not \
-                scipy.sparse.issparse(y):  # type: ignore[misc]
+        if not isinstance(y, (np.ndarray, pd.DataFrame,  # type: ignore[misc]
```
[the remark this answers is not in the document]
ravinkohli: So, we already have a variable for the types called `SUPPORTED_TARGET_TYPE` which we can use directly using `isinstance` checks, however, to check for sparsity we do not use `isinstance` rather we use scipy.issparse and hence I did not think it was appropriate to make another variable just for this type check.

Issue #239, opened 2021-05-25
Enable minority features coalescense  in autopytorch


franchuterivera:
Enable the pipeline to ignore the less frequent class. We should do this in a way that the pipeline can learn this. In other words, we should add a step in the pipeline that can ignore an infrequent features with a certain learnable criterion (maybe something like [this](https://stackoverflow.com/questions/55948491/scikitlearn-remove-less-frequent-categorical-classes)).

ravinkohli:
This issue has been resolved with PR #376

PR #242, opened 2021-05-25
[ADD] Minority Coalescer 

franchuterivera:
addresses #239

### DIFF
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -1072,11 +1072,17 @@ class BaseTask:
             self
         """
 
-        self.dataset_name = dataset.dataset_name
-
         if self._logger is None:
             self._logger = self._get_logger(str(self.dataset_name))
 
+        if self.models_ is None or len(self.models_) == 0 or self.ensemble_ is None:
+            self._load_models()
+
+        # Refit is not applicable when ensemble_size is set to zero.
+        if self.ensemble_ is None:
+            raise ValueError("Refit can only be called if 'ensemble_size != 0'")
+
+        self.dataset_name = dataset.dataset_name
         dataset_requirements = get_dataset_requirements(
             info=dataset.get_required_dataset_info(),
             include=self.include_components,
@@ -1085,26 +1091,20 @@ class BaseTask:
         dataset_properties = dataset.get_dataset_properties(dataset_requirements)
         self._backend.save_datamanager(dataset)
 
-        X: Dict[str, Any] = dict({'dataset_properties': dataset_properties,
-                                  'backend': self._backend,
-                                  'X_train': dataset.train_tensors[0],
-                                  'y_train': dataset.train_tensors[1],
-                                  'X_test': dataset.test_tensors[0] if dataset.test_tensors is not None else None,
-                                  'y_test': dataset.test_tensors[1] if dataset.test_tensors is not None else None,
-                                  'train_indices': dataset.splits[split_id][0],
-                                  'val_indices': dataset.splits[split_id][1],
-                                  'split_id': split_id,
-                                  'num_run': self._backend.get_next_num_run(),
-                                  })
-        X.update({**self.pipeline_options, **budget_config})
-        if self.models_ is None or len(self.models_) == 0 or self.ensemble_ is None:
-            self._load_models()
-
-        # Refit is not applicable when ensemble_size is set to zero.
-        if self.ensemble_ is None:
-            raise ValueError("Refit can only be called if 'ensemble_size != 0'")
-
         for identifier in self.models_:
+            X: Dict[str, Any] = dict({'dataset_properties': dataset_properties,
+                                      'backend': self._backend,
+                                      'X_train': dataset.train_tensors[0].copy(),
+                                      'y_train': dataset.train_tensors[1].copy(),
+                                      'X_test': dataset.test_tensors[0] if dataset.test_tensors is not None else None,
+                                      'y_test': dataset.test_tensors[1] if dataset.test_tensors is not None else None,
+                                      'train_indices': dataset.splits[split_id][0],
+                                      'val_indices': dataset.splits[split_id][1],
+                                      'split_id': split_id,
+                                      'num_run': self._backend.get_next_num_run(),
+                                      })
+            X.update({**self.pipeline_options, **budget_config})
+
             model = self.models_[identifier]
             # this updates the model inplace, it can then later be used in
             # predict method
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -12,11 +12,12 @@ import typing
 import unittest.mock
 import warnings
 from abc import abstractmethod
-from typing import Any, Callable, Dict, List, Optional, Tuple, Union, cast
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 from ConfigSpace.configuration_space import Configuration, ConfigurationSpace
 
 import dask
+import dask.distributed
 
 import joblib
 
@@ -38,7 +39,6 @@ from autoPyTorch.data.base_validator import BaseInputValidator
 from autoPyTorch.datasets.base_dataset import BaseDataset
 from autoPyTorch.datasets.resampling_strategy import CrossValTypes, HoldoutValTypes
 from autoPyTorch.ensemble.ensemble_builder import EnsembleBuilderManager
-from autoPyTorch.ensemble.ensemble_selection import EnsembleSelection
 from autoPyTorch.ensemble.singlebest_ensemble import SingleBest
 from autoPyTorch.evaluation.abstract_evaluator import fit_and_suppress_warnings
 from autoPyTorch.evaluation.tae import ExecuteTaFuncWithQueue, get_cost_of_crash
@@ -198,7 +198,7 @@ class BaseTask:
         # examples. Nevertheless, multi-process runs
         # have spawn as requirement to reduce the
         # possibility of a deadlock
-        self._dask_client = None
+        self._dask_client: Optional[dask.distributed.Client] = None
         self._multiprocessing_context = 'forkserver'
         if self.n_jobs == 1:
             self._multiprocessing_context = 'fork'
@@ -711,7 +711,8 @@ class BaseTask:
         precision: int = 32,
         disable_file_output: List = [],
         load_models: bool = True,
-        portfolio_selection: Optional[str] = None
+        portfolio_selection: Optional[str] = None,
+        dask_client: Optional[dask.distributed.Client] = None
     ) -> 'BaseTask':
         """
         Search for the best pipeline configuration for the given dataset.
@@ -857,10 +858,11 @@ class BaseTask:
         # If no dask client was provided, we create one, so that we can
         # start a ensemble process in parallel to smbo optimize
         if (
-            self._dask_client is None and (self.ensemble_size > 0 or self.n_jobs is not None and self.n_jobs > 1)
+            dask_client is None and (self.ensemble_size > 0 or self.n_jobs > 1)
         ):
             self._create_dask_client()
         else:
+            self._dask_client = dask_client
             self._is_dask_client_internally_created = False
 
         # Handle time resource allocation
@@ -1206,7 +1208,6 @@ class BaseTask:
 
         # Mypy assert
         assert self.ensemble_ is not None, "Load models should error out if no ensemble"
-        self.ensemble_ = cast(Union[SingleBest, EnsembleSelection], self.ensemble_)
 
         if isinstance(self.resampling_strategy, HoldoutValTypes):
             models = self.models_
@@ -1315,7 +1316,8 @@ class BaseTask:
             self._load_models()
 
         assert self.ensemble_ is not None
-        return self.ensemble_.get_models_with_weights(self.models_)
+        models_with_weights: List[Tuple[float, BasePipeline]] = self.ensemble_.get_models_with_weights(self.models_)
+        return models_with_weights
 
     def show_models(self) -> str:
         df = []
@@ -1323,7 +1325,8 @@ class BaseTask:
             representation = model.get_pipeline_representation()
             representation.update({'Weight': weight})
             df.append(representation)
-        return pd.DataFrame(df).to_markdown()
+        models_markdown: str = pd.DataFrame(df).to_markdown()
+        return models_markdown
 
     def _print_debug_info_to_log(self) -> None:
         """
--- a/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/MinorityCoalescer.py
+++ b/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/MinorityCoalescer.py
@@ -1,0 +1,52 @@
+from typing import Any, Dict, Optional, Union
+
+from ConfigSpace.configuration_space import ConfigurationSpace
+from ConfigSpace.hyperparameters import (
+    UniformFloatHyperparameter,
+)
+
+import numpy as np
+
+from autoPyTorch.pipeline.components.preprocessing.tabular_preprocessing.coalescer.base_coalescer import BaseCoalescer
+from autoPyTorch.utils.common import HyperparameterSearchSpace, add_hyperparameter
+from autoPyTorch.utils.implementations import MinorityCoalescing
+
+
+class MinorityCoalescer(BaseCoalescer):
+    """
+    Groups together classes in a categorical feature if the frequency
+    of occurrence is less than minimum_fraction
+    """
+    def __init__(self, minimum_fraction: float, random_state: Optional[Union[np.random.RandomState, int]] = None):
+        super().__init__()
+        self.minimum_fraction = minimum_fraction
+        self.random_state = random_state
+
+    def fit(self, X: Dict[str, Any], y: Any = None) -> BaseCoalescer:
+
+        self.check_requirements(X, y)
+
+        self.preprocessor['categorical'] = MinorityCoalescing(minimum_fraction=self.minimum_fraction)
+        return self
+
+    @staticmethod
+    def get_properties(dataset_properties: Optional[Dict[str, Any]] = None) -> Dict[str, Union[str, bool]]:
+        return {
+            'shortname': 'MinorityCoalescer',
+            'name': 'Minority Feature-class coalescer',
+            'handles_sparse': False
+        }
+
+    @staticmethod
+    def get_hyperparameter_search_space(
+        dataset_properties: Optional[Dict] = None,
+        minimum_fraction: HyperparameterSearchSpace = HyperparameterSearchSpace(hyperparameter="minimum_fraction",
+                                                                                value_range=(0.0001, 0.5),
+                                                                                default_value=0.01,
+                                                                                log=True),
+    ) -> ConfigurationSpace:
+        cs = ConfigurationSpace()
+
+        add_hyperparameter(cs, minimum_fraction, UniformFloatHyperparameter)
+
+        return cs
--- a/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/MinorityCoalescer.py
+++ b/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/MinorityCoalescer.py
@@ -17,7 +17,7 @@ class MinorityCoalescer(BaseCoalescer):
     Groups together classes in a categorical feature if the frequency
     of occurrence is less than minimum_fraction
     """
-    def __init__(self, minimum_fraction: float, random_state: Optional[Union[np.random.RandomState, int]] = None):
+    def __init__(self, minimum_fraction: float, random_state: np.random.RandomState):
         super().__init__()
         self.minimum_fraction = minimum_fraction
         self.random_state = random_state
--- a/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/NoCoalescer.py
+++ b/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/NoCoalescer.py
@@ -1,0 +1,51 @@
+from typing import Any, Dict, Optional, Union
+
+import numpy as np
+
+from autoPyTorch.pipeline.components.preprocessing.tabular_preprocessing.coalescer.base_coalescer import BaseCoalescer
+
+
+class NoCoalescer(BaseCoalescer):
+    """
+    Don't perform NoCoalescer on categorical features
+    """
+    def __init__(self,
+                 random_state: Optional[Union[np.random.RandomState, int]] = None
+                 ):
+        super().__init__()
+        self.random_state = random_state
+
+    def fit(self, X: Dict[str, Any], y: Any = None) -> BaseCoalescer:
+        """
+        The fit function calls the fit function of the underlying model
+        and returns the transformed array.
+        Args:
+            X (np.ndarray): input features
+            y (Optional[np.ndarray]): input labels
+
+        Returns:
+            instance of self
+        """
+        self.check_requirements(X, y)
+
+        return self
+
+    def transform(self, X: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Adds the self into the 'X' dictionary and returns it.
+        Args:
+            X (Dict[str, Any]): 'X' dictionary
+
+        Returns:
+            (Dict[str, Any]): the updated 'X' dictionary
+        """
+        X.update({'coalescer': self.preprocessor})
+        return X
+
+    @staticmethod
+    def get_properties(dataset_properties: Optional[Dict[str, Any]] = None) -> Dict[str, Union[str, bool]]:
+        return {
+            'shortname': 'NoCoalescer',
+            'name': 'No Coalescer',
+            'handles_sparse': True
+        }
--- a/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/NoCoalescer.py
+++ b/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/NoCoalescer.py
@@ -10,18 +10,20 @@ class NoCoalescer(BaseCoalescer):
     Don't perform NoCoalescer on categorical features
     """
     def __init__(self,
-                 random_state: Optional[Union[np.random.RandomState, int]] = None
+                 random_state: np.random.RandomState,
                  ):
         super().__init__()
         self.random_state = random_state
 
-    def fit(self, X: Dict[str, Any], y: Any = None) -> BaseCoalescer:
+    def fit(self, X: Dict[str, Any], y: Optional[Any] = None) -> BaseCoalescer:
         """
-        The fit function calls the fit function of the underlying model
-        and returns the transformed array.
+        As no coalescing happens, the input fit dictionary is unchanged.
+
         Args:
-            X (np.ndarray): input features
-            y (Optional[np.ndarray]): input labels
+        X (Dict[str, Any]):
+            input fit dictionary
+        y (Optional[Any]):
+            Parameter to comply with scikit-learn API. Not used.
 
         Returns:
             instance of self
@@ -32,7 +34,7 @@ class NoCoalescer(BaseCoalescer):
 
     def transform(self, X: Dict[str, Any]) -> Dict[str, Any]:
         """
-        Adds the self into the 'X' dictionary and returns it.
+        Add self into the 'X' dictionary and return the modified dict.
         Args:
             X (Dict[str, Any]): 'X' dictionary
 
--- a/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
+++ b/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
@@ -99,8 +99,8 @@ class CoalescerChoice(autoPyTorchChoice):
             if len(dataset_properties['categorical_columns']) == 0:
                 default = 'NoCoalescer'
                 if include is not None and default not in include:
-                    raise ValueError("Provided {} in include, however, the dataset "
-                                     "is incompatible with it".format(include))
+                    raise ValueError("Provided coalescer {} are incompatible with "
+                                     "the dataset without categorical columns.".format(include))
                 preprocessor = CSH.CategoricalHyperparameter('__choice__',
                                                              ['NoCoalescer'],
                                                              default_value=default)
--- a/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
+++ b/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
@@ -65,7 +65,7 @@ class CoalescerChoice(autoPyTorchChoice):
             raise ValueError("no coalescer found, please add a coalescer")
 
         if default is None:
-            defaults = ['MinorityCoalescer', 'NoCoalescer']
+            defaults = ['NoCoalescer', 'MinorityCoalescer']
             for default_ in defaults:
                 if default_ in available_preprocessors:
                     if include is not None and default_ not in include:
@@ -111,8 +111,10 @@ class CoalescerChoice(autoPyTorchChoice):
 
         # add only child hyperparameters of early_preprocessor choices
         for name in preprocessor.choices:
+            updates = self._get_search_space_updates(prefix=name)
             preprocessor_configuration_space = available_preprocessors[name].\
-                get_hyperparameter_search_space(dataset_properties)
+                get_hyperparameter_search_space(dataset_properties,
+                                                **updates)
             parent_hyperparameter = {'parent': preprocessor, 'value': name}
             cs.add_configuration_space(name, preprocessor_configuration_space,
                                        parent_hyperparameter=parent_hyperparameter)
--- a/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
+++ b/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
@@ -115,8 +115,7 @@ class CoalescerChoice(autoPyTorchChoice):
             # Call arg is ignored on mypy as the search space dynamically
             # provides different args
             preprocessor_configuration_space = available_preprocessors[name].\
-                get_hyperparameter_search_space(dataset_properties,  # type:ignore[call-arg]
-                                                **updates)
+                get_hyperparameter_search_space(dataset_properties, **updates)  # type:ignore
             parent_hyperparameter = {'parent': preprocessor, 'value': name}
             cs.add_configuration_space(name, preprocessor_configuration_space,
                                        parent_hyperparameter=parent_hyperparameter)
--- a/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
+++ b/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
@@ -1,0 +1,137 @@
+import os
+from collections import OrderedDict
+from typing import Any, Dict, List, Optional
+
+import ConfigSpace.hyperparameters as CSH
+from ConfigSpace.configuration_space import ConfigurationSpace
+
+from autoPyTorch.pipeline.components.base_choice import autoPyTorchChoice
+from autoPyTorch.pipeline.components.base_component import (
+    ThirdPartyComponents,
+    autoPyTorchComponent,
+    find_components,
+)
+from autoPyTorch.pipeline.components.preprocessing.tabular_preprocessing.coalescer.base_coalescer import BaseCoalescer
+
+
+coalescer_directory = os.path.split(__file__)[0]
+_coalescer = find_components(__package__,
+                             coalescer_directory,
+                             BaseCoalescer)
+_addons = ThirdPartyComponents(BaseCoalescer)
+
+
+def add_coalescer(coalescer: BaseCoalescer) -> None:
+    _addons.add_component(coalescer)
+
+
+class CoalescerChoice(autoPyTorchChoice):
+    """
+    Allows for dynamically choosing coalescer component at runtime
+    """
+
+    def get_components(self) -> Dict[str, autoPyTorchComponent]:
+        """Returns the available coalescer components
+
+        Args:
+            None
+
+        Returns:
+            Dict[str, autoPyTorchComponent]: all BaseCoalescer components available
+                as choices for coalescer the categorical columns
+        """
+        components = OrderedDict()
+        components.update(_coalescer)
+        components.update(_addons.components)
+        return components
+
+    def get_hyperparameter_search_space(self,
+                                        dataset_properties: Optional[Dict[str, Any]] = None,
+                                        default: Optional[str] = None,
+                                        include: Optional[List[str]] = None,
+                                        exclude: Optional[List[str]] = None) -> ConfigurationSpace:
+        cs = ConfigurationSpace()
+
+        if dataset_properties is None:
+            dataset_properties = dict()
+
+        dataset_properties = {**self.dataset_properties, **dataset_properties}
+
+        available_preprocessors = self.get_available_components(dataset_properties=dataset_properties,
+                                                                include=include,
+                                                                exclude=exclude)
+
+        if len(available_preprocessors) == 0:
+            raise ValueError("no coalescer found, please add a coalescer")
+
+        if default is None:
+            defaults = ['MinorityCoalescer', 'NoCoalescer']
+            for default_ in defaults:
+                if default_ in available_preprocessors:
+                    if include is not None and default_ not in include:
+                        continue
+                    if exclude is not None and default_ in exclude:
+                        continue
+                    default = default_
+                    break
+
+        updates = self._get_search_space_updates()
+        if '__choice__' in updates.keys():
+            choice_hyperparameter = updates['__choice__']
+            if not set(choice_hyperparameter.value_range).issubset(available_preprocessors):
+                raise ValueError("Expected given update for {} to have "
+                                 "choices in {} got {}".format(self.__class__.__name__,
+                                                               available_preprocessors,
+                                                               choice_hyperparameter.value_range))
+            if len(dataset_properties['categorical_columns']) == 0:
+                assert len(choice_hyperparameter.value_range) == 1
+                assert 'MinorityCoalescer' in choice_hyperparameter.value_range, \
+                    "Provided {} in choices, however, the dataset " \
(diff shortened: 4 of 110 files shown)


autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/MinorityCoalescer.py
```diff
@@ -0,0 +1,52 @@
...
+
+        self.preprocessor['categorical'] = MinorityCoalescing(minimum_fraction=self.minimum_fraction)
+        return self
+
+    @staticmethod
+    def get_properties(dataset_properties: Optional[Dict[str, Any]] = None) -> Dict[str, Union[str, bool]]:
+        return {
+            'shortname': 'MinorityCoalescer',
+            'name': 'Minority Feature-class coalescer',
+            'handles_sparse': False
```
ravinkohli, line 37: doesn't it? 
```suggestion
            'handles_sparse': True
```

autoPyTorch/utils/implementations.py
```diff
@@ -61,3 +65,76 @@ def __call__(self, y: Union[np.ndarray, torch.Tensor]) -> np.ndarray:
     @staticmethod
     def get_properties() -> Dict[str, Any]:
         return {'supported_losses': ['BCEWithLogitsLoss']}
+
+
+class MinorityCoalescing(BaseEstimator, TransformerMixin):
+    """ Group together categories which occurence is less than a specified
+    minimum fraction. Coalesced categories get index of one.
```
ravinkohli, line 72:
```suggestion
    minimum fraction. Coalesced categories are set to -2.
```

ravinkohli (changes requested):
Hey, thanks for the PR. I have suggested minor changes, and I think after they are resolved, this PR can be merged

autoPyTorch/api/base_task.py
```diff
@@ -1085,26 +1091,20 @@ def refit(
...
-        X.update({**self.pipeline_options, **budget_config})
-        if self.models_ is None or len(self.models_) == 0 or self.ensemble_ is None:
-            self._load_models()
-
-        # Refit is not applicable when ensemble_size is set to zero.
-        if self.ensemble_ is None:
-            raise ValueError("Refit can only be called if 'ensemble_size != 0'")
-
         for identifier in self.models_:
+            X: Dict[str, Any] = dict({'dataset_properties': dataset_properties,
```
ravinkohli, line 1095: Maybe not as a part of this PR, but could you raise an issue to change this to also use dask if possible? Similar to the fit_pipeline method in the cocktails branch.
Btw, I'll try and add fit_pipeline to `development` as well

ravinkohli (approved):
Hey, thanks for the PR. This is a great feature to have. I have left just a single comment other than that, I think we can merge this PR.

autoPyTorch/utils/implementations.py
```diff
@@ -61,3 +65,76 @@ def __call__(self, y: Union[np.ndarray, torch.Tensor]) -> np.ndarray:
...
+
+        self.do_not_coalesce_ = do_not_coalesce
+        return self
+
+    def transform(self, X: np.ndarray) -> np.ndarray:
+        self.check_X(X)
+
+        if self.minimum_fraction is None:
+            return X
+
```
nabenabe0928, line 117: The codes are a bit complicated and not simple, so I rewrote the codes.
```
def suggestion(X: Union[sparse.csr_matrix, np.ndarray]) -> None:
    n_features = X.shape[1]
    is_sparse = sparse.issparse(X)

    for col in range(n_features):
        if is_sparse:
            indptr_start = X.indptr[col],
            indptr_end = X.indptr[col + 1]
            col_data = X.data[indptr_start:indptr_end]
        else:
            col_data = X[:, col]

        mask = ~np.isin(col_data, do_not_coalesce_)
        col_data[mask] = -2

    return X
```

Just in case, I also checked if it works in the identical way.

```
size = (400000, 100)
n_vals = 10
do_not_coalesce_ = [3, 5]

X = np.random.randint(n_vals, size=size)
X_copy = np.copy(X)

n_wrongs = (original(X) != suggestion(X_copy)).sum()
n_wrongs

>>> 0  # It works identically
```

Plus, I also checked the benchmark speed.

```
size = (400000, 100)
n_vals = 10
do_not_coalesce_ = [3, 5]

X = np.random.randint(n_vals, size=size)
%timeit original(X)
>>> 1.24 s ± 25.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

X = np.random.randint(n_vals, size=size)
%timeit suggestion(X)
>>> 549 ms ± 13.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

test/test_utils/test_coalescer.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def test_coalesce_10_percent(X1):
+    X = X1
+    Y = MinorityCoalescing(minimum_fraction=.1).fit_transform(X)
+    for col in range(Y.shape[1]):
+        hist = np.histogram(Y[:, col], bins=np.arange(-2, 7))
+        np.testing.assert_array_almost_equal(hist[0], [10, 0, 0, 0, 0, 30, 30, 30])
+    # Assert no copies were made
+    assert id(X) == id(Y)
```
nabenabe0928, line 58: If you would like this method to share the address before and after the transformation, we should not return the array as a result, because it just means if we make any changes to the `X`, it means `Y` will also change unexpectedly.
It is not safe.

autoPyTorch/utils/implementations.py
```diff
@@ -61,3 +65,76 @@ def __call__(self, y: Union[np.ndarray, torch.Tensor]) -> np.ndarray:
...
+    def __init__(self, minimum_fraction: Optional[float] = None):
+        self.minimum_fraction = minimum_fraction
+
+    def check_X(self, X: np.array) -> None:
+        X_data = X.data if sparse.issparse(X) else X
+        if np.nanmin(X_data) <= -2:
+            raise ValueError("X needs to contain only integers greater than -2.")
+
+    def fit(self, X: np.array, y: Optional[np.ndarray] = None) -> 'MinorityCoalescing':
```
nabenabe0928, line 83: I will check `fit` method later.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/base_coalescer.py
```diff
@@ -0,0 +1,32 @@
+from typing import Any, Dict, List
```
nabenabe0928, line 1: Check this file later.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
```diff
@@ -0,0 +1,141 @@
+import os
```
nabenabe0928, line 1: Check this file later.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/NoCoalescer.py
```diff
@@ -0,0 +1,51 @@
+from typing import Any, Dict, Optional, Union
```
nabenabe0928, line 1: Check this file later.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/MinorityCoalescer.py
```diff
@@ -0,0 +1,52 @@
+from typing import Any, Dict, Optional, Union
```
nabenabe0928, line 1: Check this file later.

nabenabe0928 (changes requested):
Thanks for the PR.
This helps to prevent unexpectedly low performance.
However, some codes seem to include vulnerabilities, so first let's remove those issues.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/MinorityCoalescer.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class MinorityCoalescer(BaseCoalescer):
+    """
+    Groups together classes in a categorical feature if the frequency
+    of occurrence is less than minimum_fraction
+    """
+    def __init__(self, minimum_fraction: float, random_state: Optional[Union[np.random.RandomState, int]] = None):
+        super().__init__()
+        self.minimum_fraction = minimum_fraction
```
nabenabe0928, line 22: `minimum_fraction` -> `min_fraction` (convention)

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/MinorityCoalescer.py
nabenabe0928, line 37: `handle_sparse` (imperative form)

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/NoCoalescer.py
```diff
@@ -0,0 +1,51 @@
...
+
+import numpy as np
+
+from autoPyTorch.pipeline.components.preprocessing.tabular_preprocessing.coalescer.base_coalescer import BaseCoalescer
+
+
+class NoCoalescer(BaseCoalescer):
+    """
+    Don't perform NoCoalescer on categorical features
```
nabenabe0928, line 10: Why?

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/MinorityCoalescer.py
```diff
@@ -0,0 +1,52 @@
...
+from autoPyTorch.utils.common import HyperparameterSearchSpace, add_hyperparameter
+from autoPyTorch.utils.implementations import MinorityCoalescing
+
+
+class MinorityCoalescer(BaseCoalescer):
+    """
+    Groups together classes in a categorical feature if the frequency
+    of occurrence is less than minimum_fraction
+    """
+    def __init__(self, minimum_fraction: float, random_state: Optional[Union[np.random.RandomState, int]] = None):
```
nabenabe0928, line 20: `random_state` is now only `np.random.RandomState`, right?

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/NoCoalescer.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+    def __init__(self,
+                 random_state: Optional[Union[np.random.RandomState, int]] = None
```
nabenabe0928, line 13: only `np.random.RandomState`

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/NoCoalescer.py
```diff
@@ -0,0 +1,51 @@
...
+                 ):
+        super().__init__()
+        self.random_state = random_state
+
+    def fit(self, X: Dict[str, Any], y: Any = None) -> BaseCoalescer:
+        """
+        The fit function calls the fit function of the underlying model
+        and returns the transformed array.
```
nabenabe0928, line 21: The transformed feature array? (it means we transform only `X`)

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/NoCoalescer.py
```diff
@@ -0,0 +1,51 @@
...
+        Returns:
+            instance of self
+        """
+        self.check_requirements(X, y)
+
+        return self
+
+    def transform(self, X: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Adds the self into the 'X' dictionary and returns it.
```
nabenabe0928, line 35: Add ... and return ... (imperative form)

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/NoCoalescer.py
```diff
@@ -0,0 +1,51 @@
...
+        Args:
+            X (Dict[str, Any]): 'X' dictionary
```
nabenabe0928, line 37: What is `X` dictionary?
`fit_dictionary`? (Ravin says the `fit_dictionary` will deprecate soon, but do you have any idea when?)

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/NoCoalescer.py
```diff
@@ -0,0 +1,51 @@
...
+        """
+        X.update({'coalescer': self.preprocessor})
+        return X
+
+    @staticmethod
+    def get_properties(dataset_properties: Optional[Dict[str, Any]] = None) -> Dict[str, Union[str, bool]]:
+        return {
+            'shortname': 'NoCoalescer',
+            'name': 'No Coalescer',
+            'handles_sparse': True
```
nabenabe0928, line 50: `handle_sparse` (imperative form)

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+            dataset_properties = dict()
+
+        dataset_properties = {**self.dataset_properties, **dataset_properties}
+
+        available_preprocessors = self.get_available_components(dataset_properties=dataset_properties,
+                                                                include=include,
+                                                                exclude=exclude)
+
+        if len(available_preprocessors) == 0:
+            raise ValueError("no coalescer found, please add a coalescer")
```
nabenabe0928, line 65: `No coalescer found, add coalescers from <where>`

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+
+        if default is None:
```
nabenabe0928, line 67: check below from here later

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/base_coalescer.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    def __init__(self) -> None:
+        super().__init__()
+        self.add_fit_requirements([
+            FitRequirement('categorical_columns', (List,), user_defined=True, dataset_property=True),
+            FitRequirement('categories', (List,), user_defined=True, dataset_property=True)])
+
+    def transform(self, X: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Adds the self into the 'X' dictionary and returns it.
```
nabenabe0928, line 21: Add and return (imperative form)

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/base_coalescer.py
```diff
@@ -0,0 +1,32 @@
...
+        Args:
+            X (Dict[str, Any]): 'X' dictionary
```
nabenabe0928, line 23: What is `X`?

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/base_coalescer.py
```diff
@@ -0,0 +1,32 @@
...
+
+        Returns:
+            (Dict[str, Any]): the updated 'X' dictionary
+        """
+        if self.preprocessor['numerical'] is None and self.preprocessor['categorical'] is None:
```
nabenabe0928, line 28: `and`, but not `or` ? (I do not know the exact behavior of preprocessor, so please let me know how it works.)

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/base_coalescer.py
```diff
@@ -0,0 +1,32 @@
...
+            raise ValueError("cant call transform on {} without fitting first."
```
nabenabe0928, line 29: Cannot or Can't
without fitting first -> before fitting

nabenabe0928 (changes requested):

autoPyTorch/utils/implementations.py
```diff
@@ -61,3 +65,76 @@ def __call__(self, y: Union[np.ndarray, torch.Tensor]) -> np.ndarray:
...
+                        X.data[indptr_start:indptr_end][
+                            X.data[indptr_start:indptr_end] == unique_value] = -2
+            else:
+                unique = np.unique(X[:, column])
+                unique_values = [unique_value for unique_value in unique
+                                 if unique_value not in self.do_not_coalesce_[column]]
+                mask = np.isin(X[:, column], unique_values)
+                # The imputer uses -1 for unknown categories
+                # Then -2 means coalesced categories
+                X[mask, column] = -2
```
nabenabe0928, line 136:
```suggestion
        n_features = X.shape[1]
        is_sparse = sparse.issparse(X)
        
        for col in range(n_features):
            if is_sparse:
                indptr_start = X.indptr[col],
                indptr_end = X.indptr[col + 1]
                col_data = X.data[indptr_start:indptr_end]
            else:
                col_data = X[:, col]

            # The imputer uses -1 for unknown categories
            # Then -2 means coalesced categories
            mask = ~np.isin(col_data, self.do_not_coalesce_)
            col_data[mask] = -2
```

autoPyTorch/utils/implementations.py
```diff
@@ -61,3 +65,76 @@ def __call__(self, y: Union[np.ndarray, torch.Tensor]) -> np.ndarray:
...
+            raise ValueError("X needs to contain only integers greater than -2.")
+
+    def fit(self, X: np.array, y: Optional[np.ndarray] = None) -> 'MinorityCoalescing':
+        self.check_X(X)
+
+        if self.minimum_fraction is None:
+            return self
+
+        # Remember which values should not be coalesced
+        do_not_coalesce: List[Set[int]] = list()
```
nabenabe0928, line 90: Clean this function (**SHUHEI TODO**)

nabenabe0928 (changes requested):

autoPyTorch/utils/implementations.py
```diff
@@ -61,3 +65,76 @@ def __call__(self, y: Union[np.ndarray, torch.Tensor]) -> np.ndarray:
...
+                    X.data[indptr_start:indptr_end], return_counts=True)
+                colsize = indptr_end - indptr_start
+            else:
+                unique, counts = np.unique(X[:, column], return_counts=True)
+                colsize = X.shape[0]
+
+            for unique_value, count in zip(unique, counts):
+                fraction = float(count) / colsize
+                if fraction >= self.minimum_fraction:
+                    do_not_coalesce[-1].add(unique_value)
```
nabenabe0928, line 107: I found some issues and refactored the codes.
**Issues**
1. colsize changes depending on if the feature matrix is sparse or not (`indptr_end - indptr_start` is not correct)
2. Unneeded usage of for loop (rewrite by numpy to eliminate for loop, which I did)

This function can also be written nicely.
```suggestion
        n_instances, n_features = X.shape
        do_not_coalesce = [set() for _ in range(n_features)]
        is_sparse = sparse.issparse(X)

        for col, no_coal in enumerate(do_not_coalesce):
            if is_sparse:
                indptr_start = X.indptr[column]
                indptr_end = X.indptr[column + 1]
                col_data = X.data[indptr_start:indptr_end]
            else:
                col_data = X[:, col]

            unique_vals, counts = np.unique(col_data, return_counts=True)
            frac = counts / n_instances
            no_coal.update(unique_vals[frac >= min_frac])
```

nabenabe0928 (changes requested):
I think `fit` and `transform` methods in `MinorityCoalescing` class can be merged when I refactor them.
Please take a look at the refactored version of both methods.

Just in case, I also put the refactored version of the merged method.

```
        n_instances, n_features = X.shape
        is_sparse = sparse.issparse(X)

        for col in range(n_features):
            if is_sparse:
                indptr_start = X.indptr[column]
                indptr_end = X.indptr[column + 1]
                col_data = X.data[indptr_start:indptr_end]
            else:
                col_data = X[:, col]

            unique_vals, counts = np.unique(col_data, return_counts=True)
            frac = counts / n_instances
            coalesce = set(unique_vals[frac <= min_frac])

            # The imputer uses -1 for unknown categories
            # Then -2 means coalesced categories
            mask = np.isin(col_data, coalesce)
            col_data[mask] = -2
```

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+            preprocessor = CSH.CategoricalHyperparameter('__choice__',
+                                                         choice_hyperparameter.value_range,
+                                                         default_value=choice_hyperparameter.default_value)
+        else:
+            # add only no coalescer to choice hyperparameters in case the dataset is only numerical
+            if len(dataset_properties['categorical_columns']) == 0:
+                default = 'NoCoalescer'
+                if include is not None and default not in include:
+                    raise ValueError("Provided {} in include, however, the dataset "
+                                     "is incompatible with it".format(include))
```
nabenabe0928, line 101: The sentence is a bit confusing because the first sentence is an imperative form and the second sentence is not.
It is better to divide into two sentences.

```suggestion
                    raise ValueError("Provided coalescer {} are incompatible with "
                                     "the dataset without categorical columns.".format(include))
```

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+                                                               choice_hyperparameter.value_range))
+            if len(dataset_properties['categorical_columns']) == 0:
+                assert len(choice_hyperparameter.value_range) == 1
+                assert 'MinorityCoalescer' in choice_hyperparameter.value_range, \
+                    "Provided {} in choices, however, the dataset " \
+                    "is incompatible with it".format(choice_hyperparameter.value_range)
+
+            preprocessor = CSH.CategoricalHyperparameter('__choice__',
+                                                         choice_hyperparameter.value_range,
+                                                         default_value=choice_hyperparameter.default_value)
```
nabenabe0928, line 94: Could you make a method for this?
The method is already too long.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+                    if exclude is not None and default_ in exclude:
+                        continue
+                    default = default_
+                    break
+
+        updates = self._get_search_space_updates()
+        if '__choice__' in updates.keys():
+            choice_hyperparameter = updates['__choice__']
+            if not set(choice_hyperparameter.value_range).issubset(available_preprocessors):
+                raise ValueError("Expected given update for {} to have "
```
nabenabe0928, line 82: Does this mean like this (I did not understand this message)?:
`The update for {self.__class__.__name__} expected {available_preprocessors}, but got {set(choice_hyperparameter.value_range) - set(available_preprocessors)}`

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+                if include is not None and default not in include:
+                    raise ValueError("Provided {} in include, however, the dataset "
+                                     "is incompatible with it".format(include))
+                preprocessor = CSH.CategoricalHyperparameter('__choice__',
+                                                             ['NoCoalescer'],
+                                                             default_value=default)
+            else:
+                preprocessor = CSH.CategoricalHyperparameter('__choice__',
+                                                             list(available_preprocessors.keys()),
+                                                             default_value=default)
```
nabenabe0928, line 108: Same here. Please divide this process with an understandable method name.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+        for name in preprocessor.choices:
+            updates = self._get_search_space_updates(prefix=name)
+            # Call arg is ignored on mypy as the search space dynamically
+            # provides different args
+            preprocessor_configuration_space = available_preprocessors[       # type:ignore[call-arg]
+                name                                                          # type:ignore[call-arg]
+            ].get_hyperparameter_search_space(dataset_properties, **updates)  # type:ignore[call-arg]
+            parent_hyperparameter = {'parent': preprocessor, 'value': name}
+            cs.add_configuration_space(name, preprocessor_configuration_space,
+                                       parent_hyperparameter=parent_hyperparameter)
```
nabenabe0928, line 122: If you need a comment for the process, please divide the process into a small method.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+        Returns:
+            Dict[str, autoPyTorchComponent]: all BaseCoalescer components available
+                as choices for coalescer the categorical columns
+        """
+        components = OrderedDict()
+        components.update(_coalescer)
+        components.update(_addons.components)
+        return components
+
+    def get_hyperparameter_search_space(self,
```
nabenabe0928, line 48: Overall 78 lines for this method.
Please divide the method appropriately.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+        if default is None:
+            defaults = ['NoCoalescer', 'MinorityCoalescer']
+            for default_ in defaults:
+                if default_ in available_preprocessors:
+                    if include is not None and default_ not in include:
+                        continue
+                    if exclude is not None and default_ in exclude:
+                        continue
+                    default = default_
+                    break
```
nabenabe0928, line 76: This process will be a better choice to divide.
say `def _get_default_value()`

nabenabe0928 (changes requested):

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
franchuterivera: thanks a lot for the suggestion, but all Choice objects have the same structure. For instance check `encoding/__init__.py`.

I am not saying we should not change it. By all means if you think this is something that will improve readability we should do this, but then better raise an issue where we change all Choice objects and not just this one.

For this PR I suggest to prevent making this Choice the only one that is different and following this request here in this issue: #267

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
franchuterivera: thanks a lot for the suggestion, but all Choice objects have the same structure. For instance check `encoding/__init__.py`.

I am not saying we should not change it. By all means if you think this is something that will improve readability we should do this, but then better raise an issue where we change all Choice objects and not just this one.

For this PR I suggest to prevent making this Choice the only one that is different and following this request here in this issue: #267

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
franchuterivera: thanks a lot for the suggestion, but all Choice objects have the same structure. For instance check `encoding/__init__.py`.

I am not saying we should not change it. By all means if you think this is something that will improve readability we should do this, but then better raise an issue where we change all Choice objects and not just this one.

For this PR I suggest to prevent making this Choice the only one that is different and following this request here in this issue: #267

autoPyTorch/api/base_task.py
```diff
@@ -1072,11 +1072,17 @@ def refit(
...
-
         if self._logger is None:
             self._logger = self._get_logger(str(self.dataset_name))
 
+        if self.models_ is None or len(self.models_) == 0 or self.ensemble_ is None:
+            self._load_models()
+
+        # Refit is not applicable when ensemble_size is set to zero.
+        if self.ensemble_ is None:
+            raise ValueError("Refit can only be called if 'ensemble_size != 0'")
```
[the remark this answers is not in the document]
franchuterivera: Auto-sklearn, the old Auto-PyTorch, Autogluon and scikit learn all use this naming convention. The name refit.
I will recommend staying with this name?

test/test_utils/test_coalescer.py
franchuterivera: As the name of the method (transform) indicates, it is within the expectation that the input argument will be transformed.

Normally the input can be an array of thousands of elements. Creating a copy every time we transform is not feasible, so we do expect to transform it.

autoPyTorch/utils/implementations.py
franchuterivera: Thanks for the suggestion, I have fixed this.

test/test_pipeline/components/preprocessing/test_coalesce.py
```diff
@@ -0,0 +1,53 @@
...
+        self.assertListEqual(
+            sorted(cs.get_hyperparameter('__choice__').choices),
+            sorted(list(coalescer_choice.get_components().keys()))
+        )
+
+        # Make sure we can properly set some random configs
+        # Whereas just one iteration will make sure the algorithm works,
+        # doing five iterations increase the confidence. We will be able to
+        # catch component specific crashes
+        for i in range(5):
```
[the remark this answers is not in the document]
franchuterivera: Thanks for the suggestion, I have fixed this.

autoPyTorch/utils/implementations.py
```diff
@@ -61,3 +65,76 @@ def __call__(self, y: Union[np.ndarray, torch.Tensor]) -> np.ndarray:
...
+    minimum fraction. Coalesced categories get index of one.
+    """
+
+    def __init__(self, minimum_fraction: Optional[float] = None):
+        self.minimum_fraction = minimum_fraction
+
+    def check_X(self, X: np.array) -> None:
+        X_data = X.data if sparse.issparse(X) else X
+        if np.nanmin(X_data) <= -2:
+            raise ValueError("X needs to contain only integers greater than -2.")
```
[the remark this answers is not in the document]
franchuterivera: Thanks for pointing this out, I have improved the message and added more doc-strings.

autoPyTorch/utils/implementations.py
```diff
@@ -61,3 +65,76 @@ def __call__(self, y: Union[np.ndarray, torch.Tensor]) -> np.ndarray:
...
+
+            for unique_value, count in zip(unique, counts):
+                fraction = float(count) / colsize
+                if fraction >= self.minimum_fraction:
+                    do_not_coalesce[-1].add(unique_value)
+
+        self.do_not_coalesce_ = do_not_coalesce
+        return self
+
+    def transform(self, X: np.ndarray) -> np.ndarray:
```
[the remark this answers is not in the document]
franchuterivera: Thanks for pointing out the typing. The method transform should indeed transform the object, as the name of the method suggested.

We should not copy arrays every time we transform the data, it is not memory feasible.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/MinorityCoalescer.py
franchuterivera: I don't think there is an agreement that the code is written in imperative form. Most of the core is written in the third person, as it was taken from auto-sklearn. You are very right that [google's code style](https://google.github.io/styleguide/pyguide.html) suggest that we use imperative and our docstrings are in google style format. I will recommend to disentangle this decision from this PR and have @nabenabe0928 and @ravinkohli  decide on this (and how to do so in automlcommon).

The reason for this is that `handles_*` is used not only on this new code but everywhere else.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/NoCoalescer.py
franchuterivera: This is a `NoCoalescer` class. This allows the BO model to enable/disable coalescing. 

The choice object selects between MinorityCoalescer and NoCoalescer depending on what gives better performance.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/MinorityCoalescer.py
franchuterivera: Yes, thank you very much for the catch, I have fixed this.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/NoCoalescer.py
franchuterivera: Thanks, I have improved the message.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/NoCoalescer.py
franchuterivera: As this is an isolated change, I have done this here.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/NoCoalescer.py
franchuterivera: Scikit-learn supports passing a dictionary alongside the data. See [here](https://github.com/scikit-learn/scikit-learn/blob/2beed55847ee70d363bdbfe14ee4401438fba057/sklearn/pipeline.py#L266)

It makes a lot of sense to use it instead of X as a fit_dictionary.

From all of the refactoring changes, this is to me the most important.

When** depends on when there is a contributor that wants to do this change :)

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/NoCoalescer.py
franchuterivera: See answer above please.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
franchuterivera: Thanks for the suggestion, I have improved this message.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/base_coalescer.py
franchuterivera: I have added a classification inside of the code.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/base_coalescer.py
franchuterivera: This part is something that @ravinkohli can clarify better. All preprocessors like encoders have this.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/base_coalescer.py
franchuterivera: Thanks for the suggestion, I have fixed this.

autoPyTorch/utils/implementations.py
```diff
@@ -61,3 +65,76 @@ def __call__(self, y: Union[np.ndarray, torch.Tensor]) -> np.ndarray:
...
+
+class MinorityCoalescing(BaseEstimator, TransformerMixin):
+    """ Group together categories which occurence is less than a specified
+    minimum fraction. Coalesced categories get index of one.
+    """
+
+    def __init__(self, minimum_fraction: Optional[float] = None):
+        self.minimum_fraction = minimum_fraction
+
+    def check_X(self, X: np.array) -> None:
```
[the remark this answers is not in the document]
franchuterivera: Thanks for the catch, I have fixed this problem

autoPyTorch/utils/implementations.py
```diff
@@ -61,3 +65,76 @@ def __call__(self, y: Union[np.ndarray, torch.Tensor]) -> np.ndarray:
...
+            else:
+                unique, counts = np.unique(X[:, column], return_counts=True)
+                colsize = X.shape[0]
+
+            for unique_value, count in zip(unique, counts):
+                fraction = float(count) / colsize
+                if fraction >= self.minimum_fraction:
+                    do_not_coalesce[-1].add(unique_value)
+
+        self.do_not_coalesce_ = do_not_coalesce
```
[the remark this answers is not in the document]
franchuterivera: Yes, I have made the change. This is something we also have not discussed before, as it does not have any meaning for python.. It is how you want to make the convention. Maybe you can sync up with Ravin and have this be added to the wiki.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
franchuterivera: thanks a lot for the suggestion, but all Choice objects have the same structure. For instance check `encoding/__init__.py`.

I am not saying we should not change it. By all means if you think this is something that will improve readability we should do this, but then better raise an issue where we change all Choice objects and not just this one.

For this PR I suggest to prevent making this Choice the only one that is different and following this request here in this issue: #267

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
franchuterivera: Thanks I will make this change more clearn.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/__init__.py
franchuterivera: Please see answer above.

autoPyTorch/utils/implementations.py
franchuterivera: Yes, but we do not want to merge fit and transform as we are using the scikit-learn API.  You can read more about it [here](https://towardsdatascience.com/fit-vs-transform-in-scikit-libraries-for-machine-learning-3c70e6300ded)

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/MinorityCoalescer.py
franchuterivera: I don't think we have a convention for this, and in this case, having the complete word is more clear. If possible I would like to preserve it.

autoPyTorch/api/base_task.py
franchuterivera: I have created https://github.com/automl/Auto-PyTorch/issues/268

test/test_utils/test_coalescer.py
nabenabe0928: Yes, the copy take memory, so I would like it to be a destructive method.
Because we potentially change the original (the one before transformed) and those changes will be reflected to the transformed version as well.
In other words, we do not have to return the transformed version.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/MinorityCoalescer.py
nabenabe0928: Pytorch uses `min` rather than `minimum`.
When it uses `minimum`, it is only for the element-wise minimum.
But if you would like to stick to it, it is also fine.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/MinorityCoalescer.py
nabenabe0928: Ok, then let's go with `handles_*` now.
About the third person matter, have you ever seen any functions whose name is `gets_<something>` or `processes_<something>`?
But here, I got you and let's stay without the changes.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/NoCoalescer.py
nabenabe0928: I mean i did not get if you mean `Do not perform NO coalescer` or `Do not perform coalescer`.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/NoCoalescer.py
nabenabe0928: Sorry I did not get you, so please add your ideas to the doc-string as well?
Especially, I do not get why the meaning behind 'X' in the sentence.
But still it is a bit confusing for me.
Do you know why sklearn uses `X` for both `fit_dictionary` and `feature_data`?

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/coalescer/base_coalescer.py
nabenabe0928: OK, if you guys reached the agreements.

autoPyTorch/utils/implementations.py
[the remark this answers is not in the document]
nabenabe0928: Ok, I will do it, but mostly I just follow pep8 rules.
Btw, is `<var name>_` naming from another language?

autoPyTorch/utils/implementations.py
nabenabe0928: Ok, then I will respond to the corresponding lines separately.

nabenabe0928:
I added some comments mostly small comments as responses.
And about two large suggestions.
Those two return exactly the same results, but my codes use numpy nicely.

ravinkohli:
we have merged this with #376

PR #245, opened 2021-06-02
Cocktail hotfixes

ArlindKadra:
1. Updated the dropout implementation, since the dropout rates were not calculated correctly.
2. Fixed the dropout implementation.
3. Removed the flatten layer in NoHead.
4. Updated the normalization procedure in the backbones since the addition of conditional skip connection broke things.

### DIFF
--- a/autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
+++ b/autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
@@ -41,7 +41,7 @@ class ResNetBackbone(NetworkBackboneComponent):
                     out_features=self.config["num_units_%d" % i],
                     blocks_per_group=self.config["blocks_per_group_%d" % i],
                     last_block_index=(i - 1) * self.config["blocks_per_group_%d" % i],
-                    dropout=self.config[f'dropout_{i}'],
+                    dropout=self.config[f'dropout_{i}'] if self.config['use_dropout'] else None,
                 )
             )
         if self.config['use_batch_norm']:
@@ -52,7 +52,7 @@ class ResNetBackbone(NetworkBackboneComponent):
         return backbone
 
     def _add_group(self, in_features: int, out_features: int,
-                   blocks_per_group: int, last_block_index: int, dropout: float
+                   blocks_per_group: int, last_block_index: int, dropout: Optional[float]
                    ) -> nn.Module:
         """
         Adds a group into the main backbone.
@@ -64,7 +64,8 @@ class ResNetBackbone(NetworkBackboneComponent):
             out_features (int): output dimensionality for the current block
             blocks_per_group (int): Number of ResNet per group
             last_block_index (int): block index for shake regularization
-            dropout (float): dropout value for the group
+            dropout (None, float): dropout value for the group. If none,
+                no dropout is applied.
         """
         blocks = list()
         for i in range(blocks_per_group):
@@ -245,7 +246,7 @@ class ResBlock(nn.Module):
         out_features: int,
         blocks_per_group: int,
         block_index: int,
-        dropout: float,
+        dropout: Optional[float],
         activation: nn.Module
     ):
         super(ResBlock, self).__init__()
@@ -303,7 +304,7 @@ class ResBlock(nn.Module):
             layers.append(nn.BatchNorm1d(out_features))
         layers.append(self.activation())
 
-        if self.config['use_dropout']:
+        if self.dropout is not None:
             layers.append(nn.Dropout(self.dropout))
         layers.append(nn.Linear(out_features, out_features))
 
--- a/autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
+++ b/autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
@@ -292,10 +292,11 @@ class ResBlock(nn.Module):
             layers.append(self.activation())
         else:
             # if start norm is not None and skip connection is None
-            # we will never apply the start_norm for the first block,
+            # we will never apply the start_norm for the first layer in the block,
             # which is why we should account for this case.
             if not self.config['use_skip_connection']:
-                layers.append(nn.BatchNorm1d(in_features))
+                if self.config['use_batch_norm']:
+                    layers.append(nn.BatchNorm1d(in_features))
                 layers.append(self.activation())
 
         layers.append(nn.Linear(in_features, out_features))
--- a/autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
+++ b/autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
@@ -41,7 +41,7 @@ class ResNetBackbone(NetworkBackboneComponent):
                     out_features=self.config["num_units_%d" % i],
                     blocks_per_group=self.config["blocks_per_group_%d" % i],
                     last_block_index=(i - 1) * self.config["blocks_per_group_%d" % i],
-                    dropout=self.config['use_dropout']
+                    dropout=self.config[f'dropout_{i}'],
                 )
             )
         if self.config['use_batch_norm']:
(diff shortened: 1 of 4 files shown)


autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
```diff
@@ -40,8 +40,12 @@ def build_backbone(self, input_shape: Tuple[int, ...]) -> None:
...
         if self.config['use_dropout'] and self.config["max_dropout"] > 0.05:
             dropout_shape = get_shaped_neuron_counts(
-                self.config['dropout_shape'], 0, 0, 1000, self.config['num_groups']
-            )
+                self.config['dropout_shape'],
+                0,
+                self.config["max_dropout"],
+                1000,
+                self.config['num_groups'] + 2,
+            )[:-1]
```
ravinkohli, line 48: why are we leaving out the last of the shape? Wouldn't it mean that it won't have a dropout value for the last block?

autoPyTorch/pipeline/components/setup/network_head/no_head.py
```diff
@@ -20,7 +20,7 @@ class NoHead(NetworkHeadComponent):
     """
 
     def build_head(self, input_shape: Tuple[int, ...], output_shape: Tuple[int, ...]) -> nn.Module:
-        layers = [nn.Flatten()]
+        layers = []
         in_features = np.prod(input_shape).item()
         out_features = np.prod(output_shape).item()
         layers.append(_activations[self.config["activation"]]())
```
ravinkohli, line 26: hey, just for correctness, could you set the 'handles_image' and 'handles_time_series' flag to false?

test/test_pipeline/components/setup/test_setup.py
```diff
@@ -441,7 +440,10 @@ def test_dummy_forward_backward_pass(self, task_type_input_output_shape):
         if task_type in constants.CLASSIFICATION_TASKS:
             dataset_properties["num_classes"] = output_shape[0]
 
-        cs = network_head_choice.get_hyperparameter_search_space(dataset_properties=dataset_properties)
+        cs = network_head_choice.get_hyperparameter_search_space(
```
ravinkohli, line 443: this issue should be fixed also, if you change the flag

autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
```diff
@@ -320,6 +329,7 @@ def forward(self, x: torch.FloatTensor) -> torch.FloatTensor:
             if self.config["use_skip_connection"]:
                 residual = self.shortcut(x)
 
+        # TODO make the below code better
```
ravinkohli, line 332: maybe you can also add the fix in line no. 256 which we discussed offline.

ravinkohli (changes requested):
Hey, thanks for the PR. After merging this PR, we should see a considerable improvement over the current results. I have requested a few small changes. One of them is, maybe you can fix the skip connection bug in the ResBlock module which we discussed on mattermost.

autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
```diff
@@ -289,13 +290,21 @@ def _build_block(self, in_features: int, out_features: int) -> nn.Module:
             if self.config['use_batch_norm']:
                 layers.append(nn.BatchNorm1d(in_features))
             layers.append(self.activation())
+        else:
+            # if start norm is not None and skip connection is None
+            # we will never apply the start_norm for the first block,
+            # which is why we should account for this case.
+            if not self.config['use_skip_connection']:
+                layers.append(nn.BatchNorm1d(in_features))
```
ravinkohli, line 298: should this not have a condition to check if we should use batch norm?

autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
ArlindKadra: Hmm no, as the number of layers which is passed is `self.config['num_groups'] + 2`. Thinking about it more, this should actually be +1, since dropout is not applied after the first input layer, but only between the layers between the `ResBlock`.
The reason why we pass one more and we skip the last one, is because here https://github.com/automl/Auto-PyTorch/blob/2b633cec1d44f8ca7f1126f2c9ca2ca987cfdba8/autoPyTorch/pipeline/components/setup/network_backbone/utils.py#L133 at all shapes in the end we pass `output_features`. We actually want a shape and output features in the probably author implementation is meant for the number of classes. So, we can pass 0 for the output features to actually have the shapes and never actually include it, because it would not satisfy the shapes, for example if we want a `triangle` shape, including the last count, would mean including 0, which breaks the implementation.

autoPyTorch/pipeline/components/setup/network_head/no_head.py
ArlindKadra: Good point, we should actually investigate this in the future if it is indeed a problem to have a flatten layer or not, with all the problems there were already in there, maybe it was confusing. We should open an issue about it.

test/test_pipeline/components/setup/test_setup.py
ArlindKadra: That is true, I will change the flag and revert the implementation.

autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
ArlindKadra: I actually noticed that it is correct, if you look at https://arxiv.org/pdf/1705.07485.pdf `Page 2, Figure 1`, to the left it actually has the residual connection still flowing (Same for https://arxiv.org/pdf/1802.02375.pdf, `Page 3 Figure 1`. It confused me initially for `shake-shake` and `shake-drop` since I only looked at the right.  So, the residual connection can be applied for all 3 and it is correct in my pov.

autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
ArlindKadra: Ups, I missed it. Very good catch.

autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
```diff
@@ -30,25 +30,27 @@ def build_backbone(self, input_shape: Tuple[int, ...]) -> None:
...
+                0,
                 self.config["max_dropout"],
-                1000,
-                self.config['num_groups'] + 2,
+                self.config['num_groups'] + 1,
             )[:-1]
 
             dropout_shape = [
-                dropout / 1000 * self.config["max_dropout"] for dropout in dropout_shape
+                dropout for dropout in dropout_shape
```
ArlindKadra, line 53: Now that I see it, this is actually the same list.

autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
ravinkohli: okay that makes sense then

autoPyTorch/pipeline/components/setup/network_head/no_head.py
ravinkohli: yes, I'll make one

autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
ravinkohli: okay thanks for the explanation

autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
ravinkohli: btw could you add this as a comment above the code.

ArlindKadra: Good idea, I also had to look at it to understand what was happening.

ArlindKadra: I am also going to remove the max_dropout from the if condition above, since I think it serves no purpose and it will give a crash if it does not go through it because of that.

ravinkohli: yeas that makes sense

ravinkohli (approved):

PR #247, opened 2021-06-04
[fix] Dropout bug fix

ravinkohli:
This PR fixes the bug where dropout was making the input tensor zero.

### DIFF
--- a/autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
+++ b/autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
@@ -41,7 +41,7 @@ class ResNetBackbone(NetworkBackboneComponent):
                     out_features=self.config["num_units_%d" % i],
                     blocks_per_group=self.config["blocks_per_group_%d" % i],
                     last_block_index=(i - 1) * self.config["blocks_per_group_%d" % i],
-                    dropout=self.config['use_dropout']
+                    dropout=self.config[f'dropout_{i}'] if self.config['use_dropout'] else None,
                 )
             )
 
@@ -52,7 +52,7 @@ class ResNetBackbone(NetworkBackboneComponent):
         return backbone
 
     def _add_group(self, in_features: int, out_features: int,
-                   blocks_per_group: int, last_block_index: int, dropout: bool
+                   blocks_per_group: int, last_block_index: int, dropout: Optional[float]
                    ) -> nn.Module:
         """
         Adds a group into the main backbone.
@@ -206,7 +206,7 @@ class ResBlock(nn.Module):
         out_features: int,
         blocks_per_group: int,
         block_index: int,
-        dropout: bool,
+        dropout: Optional[float],
         activation: nn.Module
     ):
         super(ResBlock, self).__init__()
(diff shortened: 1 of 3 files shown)


autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
```diff
@@ -38,10 +38,10 @@ def build_backbone(self, input_shape: Tuple[int, ...]) -> None:
         self.config.update(
             {"num_units_%d" % (i): num for i, num in enumerate(neuron_counts)}
         )
-        if self.config['use_dropout'] and self.config["max_dropout"] > 0.05:
+        if self.config['use_dropout']:
             dropout_shape = get_shaped_neuron_counts(
                 self.config['resnet_shape'], 0, 0, 1000, self.config['num_groups']
-            )
+            )[:-1]
```
nabenabe0928, line 44: What is it supposed to mean?

nabenabe0928 (commented):
Thanks for the PR.
My first impression about this PR is Oh....
That is why they always return 0 outputs because the dropout rate is 1.
Anyways, it is good that we could fix this issue promptly.

autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
ArlindKadra: Since this was taken from my pr, to not write the same thing again, please take a look at https://github.com/automl/Auto-PyTorch/pull/245 at the first resolved comment.

autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
```diff
@@ -38,10 +38,14 @@ def build_backbone(self, input_shape: Tuple[int, ...]) -> None:
...
+        # we are skipping the last layer, as the function get_shaped_neuron_counts
+        # is built for getting neuron counts, so it will add the out_features to
+        # the last layer. However, in dropout we dont want to have that, we just
+        # want to use the shape and not worry about the output.
+        if self.config['use_dropout']:
             dropout_shape = get_shaped_neuron_counts(
-                self.config['resnet_shape'], 0, 0, 1000, self.config['num_groups']
-            )
+                self.config['resnet_shape'], 0, 0, 1000, self.config['num_groups'] + 1
+            )[:-1]
```
franchuterivera, line 48: Hi, I immediately do not understand this.  The comment is very clear and I get the intention... Can we add a unit test for this? 

So that we call `build_backbone` and we get a model. Then we check that dropout was added as expected?

autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
```diff
@@ -206,7 +206,7 @@ def __init__(
         out_features: int,
         blocks_per_group: int,
         block_index: int,
-        dropout: bool,
+        dropout: Optional[float],
```
franchuterivera, line 209: Can you please add a unit test where you create a pipeline with a config with dropout=0.123, and just check that the dropout of the layer is set to `0.123` to make sure this does not happen again?

ravinkohli: sure

autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
ravinkohli: sure

Issue #248, opened 2021-06-04
Weighted Loss Possible Issue


ArlindKadra:
I am noticing some very large fluctuations when working with weighted loss, in particular for task `233109`.
Are we sure we are passing the weights correctly ? Using the weighted loss results to a drop in performance from 65.86 to 54.

Uncommenting the line of code below:
https://github.com/automl/Auto-PyTorch/blob/ee685a915869c9943dec9d7d682cc7d4aad05fca/autoPyTorch/utils/implementations.py#L38

Yields 68.64 test set performance.

PR #249, opened 2021-06-07
Bug fixes

ArlindKadra:
Addressed #248

**Description:**

- Removed wrongly added condition for the starting normalization
- Refactoring the code for the multi-branch options.
- Fixing the implementation for the weighted loss.


autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
```diff
@@ -353,8 +341,9 @@ def forward(self, x: torch.FloatTensor) -> torch.FloatTensor:
                     x.is_cuda,
                 )
                 x = shake_drop(x, alpha, beta, bl)
+            else:
+                x = self.layers(x)
 
-        if self.config["use_skip_connection"]:
```
ravinkohli, line 357: why are you removing this here? residual will not exist if there is no skip connection see line 309.

ravinkohli (changes requested):
I think the residual being added needs to be fixed. Otherwise thanks for the PR, it looks great. Btw, did you test this locally? I remember you saying there was a 15% increase after changing the weighted loss issue. Did you see an increase also for the other changes?

autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
ArlindKadra: Hey @ravinkohli, it is inside of the skip connection condition now https://github.com/automl/Auto-PyTorch/blob/3b7f5598b32ce4614d18605f18e00b05fa89b65a/autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py#L327 
Maybe just looking at the changes gives the wrong view, however, you can look at the file and let me know if you see a bug again.

ArlindKadra:
@ravinkohli I just reverted my "bad" change for the skip connection from the previous PR, that extra change did not seem to affect performance. So, the only thing fixed here is the weighted loss. However, I am testing something small locally since I think there might be another small issue.

ravinkohli:
> @ravinkohli I just reverted my "bad" change for the skip connection from the previous PR, that extra change did not seem to affect performance. So, the only thing fixed here is the weighted loss. However, I am testing something small locally since I think there might be another small issue.

Do you think it's better to add all the bugs to this PR. Or would you prefer merging a PR per fix? I personally prefer the former as we can isolate all the changes in a single commit basically. and in the meantime we can work of this branch to run experiments.

ArlindKadra:
> > @ravinkohli I just reverted my "bad" change for the skip connection from the previous PR, that extra change did not seem to affect performance. So, the only thing fixed here is the weighted loss. However, I am testing something small locally since I think there might be another small issue.
> 
> Do you think it's better to add all the bugs to this PR? Or would you prefer merging a PR per fix? I personally prefer the former as we can isolate all the changes in a single commit basically. and in the meantime, we can work off this branch to run experiments.

Yep, I prefer to have the first option. I made the PR so you and Francisco can also have a look at the differences more clearly.

autoPyTorch/datasets/base_dataset.py
```diff
@@ -322,7 +322,7 @@ def create_holdout_val_split(
             self.random_state, val_share, self._get_indices(), **kwargs)
         return train, val
 
-    def get_dataset_for_training(self, split_id: int, train: bool) -> Dataset:
+    def get_dataset_for_training(self, split_id: int, train: bool, subset: int = 0) -> Dataset:
```
ravinkohli, line 325: just a suggestion, but do you think we can just use the train flag to use the subset. because if we have train=True we need 0 and if train=False we need 1?

autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
ravinkohli: yes thats true. Looking at the whole file I think these changes are fine. Thank you

autoPyTorch/pipeline/components/training/trainer/AdversarialTrainer.py
```diff
@@ -226,9 +226,17 @@ def get_hyperparameter_search_space(
...
         if dataset_properties is not None:
             if STRING_TO_TASK_TYPES[dataset_properties['task_type']] in CLASSIFICATION_TASKS:
                 add_hyperparameter(cs, weighted_loss, CategoricalHyperparameter)
+        """
+        # TODO, decouple the weighted loss from the trainer. Uncomment the code above and
+        # remove the code below. Also update the method signature, so the weighted loss
+        # is not a constant.
+        if dataset_properties is not None:
+            if STRING_TO_TASK_TYPES[dataset_properties['task_type']] in CLASSIFICATION_TASKS:
+                add_hyperparameter(cs, weighted_loss, Constant)
```
ravinkohli, line 240: why are we making it a constant again?

autoPyTorch/pipeline/components/training/trainer/RowCutOutTrainer.py
```diff
@@ -40,22 +40,23 @@ def data_preparation(self, X: np.ndarray, y: np.ndarray,
         indices = self.random_state.choice(range(1, size), max(1, np.int32(size * self.patch_ratio)),
                                            replace=False)
 
-        if not isinstance(self.numerical_columns, typing.Iterable):
+        """if not isinstance(self.numerical_columns, typing.Iterable):
             raise ValueError("{} requires numerical columns information of {}"
                              "to prepare data got {}.".format(self.__class__.__name__,
                                                               typing.Iterable,
                                                               self.numerical_columns))
         numerical_indices = torch.tensor(self.numerical_columns)
```
ravinkohli, line 48:
```suggestion
        numerical_indices = torch.tensor([index for index in indices if index in self.numerical_columns)
```

We can use this code and then the bug should be resolved

ravinkohli (changes requested):
Hey, all in all, the changes look good. I have left a few suggestions and some questions.

ravinkohli (approved):

autoPyTorch/utils/implementations.py
```diff
@@ -35,7 +35,7 @@ def __call__(self, y: Union[np.ndarray, torch.Tensor]) -> np.ndarray:
             weights = (np.ones(y.shape[1]) * weight_per_class) / np.maximum(counts, 1)
         else:
             classes, counts = np.unique(y, axis=0, return_counts=True)
```
nabenabe0928, line 37: Just to make sure, The index of the weights should be correspondence to the class index, right?

autoPyTorch/pipeline/tabular_classification.py
```diff
@@ -91,12 +91,7 @@ def _predict_proba(self, X: np.ndarray) -> np.ndarray:
         loader = self.named_steps['data_loader'].get_loader(X=X)
         pred = self.named_steps['network'].predict(loader)
         if isinstance(self.dataset_properties['output_shape'], int):
-            proba = pred[:, :self.dataset_properties['output_shape']]
-            normalizer = proba.sum(axis=1)[:, np.newaxis]
-            normalizer[normalizer == 0.0] = 1.0
-            proba /= normalizer
-
-            return proba
+          return pred
```
nabenabe0928, line 94: Could you explain the reason behind this change?
Btw, the indent level is different in the new change.

autoPyTorch/pipeline/tabular_classification.py
```diff
@@ -147,8 +142,8 @@ def predict_proba(self, X: np.ndarray, batch_size: Optional[int] = None) -> np.n
 
         # Neural networks might not be fit to produce a [0-1] output
         # For instance, after small number of epochs.
-        y = np.clip(y, 0, 1)
-        y = sklearn.preprocessing.normalize(y, axis=1, norm='l1')
+        # y = np.clip(y, 0, 1)
+        # y = sklearn.preprocessing.normalize(y, axis=1, norm='l1')
```
nabenabe0928, line 146: So the comments above is guaranteed in fact?

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -537,8 +537,8 @@ def get_hyperparameter_search_space(
         dataset_properties: Optional[Dict] = None,
         weighted_loss: HyperparameterSearchSpace = HyperparameterSearchSpace(
             hyperparameter="weighted_loss",
-            value_range=[True, False],
-            default_value=True),
+            value_range=[1],
```
nabenabe0928, line 540:
```suggestion
            value_range=(1, ),
```

autoPyTorch/pipeline/components/training/trainer/cutout_utils.py
```diff
@@ -20,7 +20,7 @@
 class CutOut:
     def __init__(self, patch_ratio: float,
                  cutout_prob: float,
-                 weighted_loss: bool = False,
+                 weighted_loss: int = 0,
```
nabenabe0928, line 23: What is the role of `weighted_loss`?

autoPyTorch/pipeline/components/training/trainer/cutout_utils.py
```diff
@@ -63,8 +63,8 @@ def get_hyperparameter_search_space(
             dataset_properties: Optional[Dict] = None,
             weighted_loss: HyperparameterSearchSpace = HyperparameterSearchSpace(
                 hyperparameter="weighted_loss",
-                value_range=[True, False],
-                default_value=True),
+                value_range=[1],
```
nabenabe0928, line 66:
```suggestion
                value_range=(1, ),
```

autoPyTorch/pipeline/components/training/trainer/mixup_utils.py
```diff
@@ -61,8 +61,8 @@ def get_hyperparameter_search_space(
         dataset_properties: Optional[Dict] = None,
         weighted_loss: HyperparameterSearchSpace = HyperparameterSearchSpace(
             hyperparameter="weighted_loss",
-            value_range=[True, False],
-            default_value=True),
+            value_range=[1],
```
nabenabe0928, line 64:
```suggestion
            value_range=(1, ),
```

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -175,7 +175,7 @@ class BaseTrainerComponent(autoPyTorchTrainingComponent):
     """
     Base class for training
     Args:
-        weighted_loss (bool, default=False): In case for classification, whether to weight
+        weighted_loss (int, default=0): In case for classification, whether to weight
```
nabenabe0928, line 178: Why not `bool`?

autoPyTorch/pipeline/components/training/trainer/AdversarialTrainer.py
```diff
@@ -159,8 +159,8 @@ def get_hyperparameter_search_space(
         dataset_properties: Optional[Dict] = None,
         weighted_loss: HyperparameterSearchSpace = HyperparameterSearchSpace(
             hyperparameter="weighted_loss",
-            value_range=[True, False],
-            default_value=True),
+            value_range=[1],
```
nabenabe0928, line 162:
```suggestion
            value_range=(1, ),
```

nabenabe0928 (commented):
Thanks for the PR, I will put the comments of the second half of the changes in the github `Files changed` page.

autoPyTorch/pipeline/components/setup/network_head/no_head.py
```diff
@@ -23,7 +23,6 @@ def build_head(self, input_shape: Tuple[int, ...], output_shape: Tuple[int, ...]
         layers = []
         in_features = np.prod(input_shape).item()
         out_features = np.prod(output_shape).item()
-        layers.append(_activations[self.config["activation"]]())
```
nabenabe0928, line 26: Why deleted?

autoPyTorch/pipeline/components/training/data_loader/base_data_loader.py
```diff
@@ -148,6 +148,8 @@ def get_loader(self, X: np.ndarray, y: Optional[np.ndarray] = None, batch_size:
         dataset = BaseDataset(
             train_tensors=(X, y),
             # This dataset is used for loading test data in a batched format
+            seed=self.random_state.get_state()[1][0],
+            shuffle=False,
             train_transforms=self.test_transform,
             val_transforms=self.test_transform,
         )
```
nabenabe0928, line 155:
```suggestion
        # The seed set for this `np.random.RandomState`
        seed = self.random_state.get_state()[1][0]
        dataset = BaseDataset(
            train_tensors=(X, y),
            # This dataset is used for loading test data in a batched format
            seed=seed,
            shuffle=False,
            train_transforms=self.test_transform,
            val_transforms=self.test_transform,
        )
```

autoPyTorch/pipeline/components/training/trainer/RowCutOutTrainer.py
```diff
@@ -40,22 +40,23 @@ def data_preparation(self, X: np.ndarray, y: np.ndarray,
...
         # We use an ordinal encoder on the categorical columns of tabular data
         # -1 is the conceptual equivalent to 0 in a image, that does not
         # have color as a feature and hence the network has to learn to deal
         # without this data. For numerical columns we use 0 to cutout the features
         # similar to the effect that setting 0 as a pixel value in an image.
         X[:, categorical_indices.long()] = self.CATEGORICAL_VALUE
         X[:, numerical_indices.long()] = self.NUMERICAL_VALUE
-
+        """
+        X[:, indices] = 0
```
nabenabe0928, line 59: So all the `nan`s in categorical columns are now `0`? 
```suggestion
        # For both numerical and categorical, 0 is equivalent to NaN
        X[:, indices] = 0
```

autoPyTorch/datasets/base_dataset.py
nabenabe0928: I agree with this suggestion.

autoPyTorch/pipeline/components/setup/network/base_network.py
```diff
@@ -131,13 +131,14 @@ def _predict(self, network: torch.nn.Module, loader: torch.utils.data.DataLoader
...
         Y_batch_preds = list()
 
-        for i, (X_batch, Y_batch) in enumerate(loader):
-            # Predict on batch
-            X_batch = X_batch.float().to(self.device)
-            Y_batch_pred = network(X_batch)
-            if self.final_activation is not None:
-                Y_batch_pred = self.final_activation(Y_batch_pred)
-            Y_batch_preds.append(Y_batch_pred.detach().cpu())
+        with torch.no_grad():
```
nabenabe0928, line 134:
```suggestion
        # `torch.no_grad` reduces memory usage even after `model.eval()` 
        with torch.no_grad():
```

autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
```diff
@@ -261,18 +261,19 @@ def __init__(
         # if the shortcut needs a layer we apply batchnorm and activation to the shortcut
```
nabenabe0928, line 261: I have not checked this file yet.

nabenabe0928 (changes requested):
Just to make sure that I have not checked one file.

autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
```diff
@@ -327,8 +320,7 @@ def forward(self, x: torch.FloatTensor) -> torch.FloatTensor:
             # if in_features != out_features
             # -> result = W_shortcut(A(BN(x))) + W_2(~D(A(BN(W_1(A(BN(x))))))
             x = self.start_norm(x)
-            if self.config["use_skip_connection"]:
-                residual = self.shortcut(x)
+            residual = self.shortcut(x)
```
nabenabe0928, line 323: From here to until the return, I would suggest the following change:
```suggestion
        if not self.config["use_skip_connection"]:
            # Early-return
            return self.layers(x)

        # We use skip connection
        residual = self.shortcut(x)

        if self.config["multi_branch_choice"] == 'shake-shake':
            x1 = self.layers(x)
            x2 = self.shake_shake_layers(x)
            alpha, beta = shake_get_alpha_beta(self.training, x.is_cuda)
            x = shake_shake(x1, x2, alpha, beta)
        elif self.config["multi_branch_choice"] == 'shake-drop':
            x = self.layers(x)
            alpha, beta = shake_get_alpha_beta(self.training, x.is_cuda)
            bl = shake_drop_get_bl(
                self.block_index,
                1 - self.config["max_shake_drop_probability"],
                self.num_blocks,
                self.training,
                x.is_cuda,
            )
            x = shake_drop(x, alpha, beta, bl)
        else:
            x = self.layers(x)

        return x + residual
```

autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
```diff
@@ -290,14 +291,6 @@ def _build_block(self, in_features: int, out_features: int) -> nn.Module:
...
                 layers.append(nn.BatchNorm1d(in_features))
             layers.append(self.activation())
-        else:
-            # if start norm is not None and skip connection is None
-            # we will never apply the start_norm for the first layer in the block,
-            # which is why we should account for this case.
-            if not self.config['use_skip_connection']:
-                if self.config['use_batch_norm']:
-                    layers.append(nn.BatchNorm1d(in_features))
-                layers.append(self.activation())
```
nabenabe0928, line 300: Why can we delete them?

autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
```diff
@@ -261,18 +261,19 @@ def __init__(
         # if the shortcut needs a layer we apply batchnorm and activation to the shortcut
         # as well (start_norm)
         if in_features != out_features:
```
nabenabe0928, line 263:
```suggestion
        if in_features != out_features and self.config["use_skip_connection"]:
            self.shortcut = nn.Linear(in_features, out_features)
            initial_normalization = list()
            if self.config['use_batch_norm']:
                initial_normalization.append(
                    nn.BatchNorm1d(in_features)
                )
            initial_normalization.append(
                self.activation()
            )
            self.start_norm = nn.Sequential(
                *initial_normalization
            )
```

autoPyTorch/pipeline/tabular_classification.py
ArlindKadra: I think this change was suggested by Ravin. The idea was that the predictions did not need to be normalized, @ravinkohli can explain it better. I think at a certain time the code below was creating an error, but not anymore.

autoPyTorch/pipeline/tabular_classification.py
ArlindKadra: @ravinkohli

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
ArlindKadra: to check, if passing a tuple is what is expected.

autoPyTorch/pipeline/components/training/data_loader/base_data_loader.py
ArlindKadra: I do not agree with creating an extra variable. It is unnecessary and it does not make it clearer unless I am missing another pro in the code change.

autoPyTorch/pipeline/components/training/trainer/RowCutOutTrainer.py
ArlindKadra: More like certain features for certain rows are selected to be dropped via a mask.

autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
ArlindKadra: To check

autoPyTorch/pipeline/tabular_classification.py
nabenabe0928, line 94:
```suggestion
            # The final layer is always softmax now (it already gives pseudo proba)
            return pred
```

nabenabe0928 (changes requested):

autoPyTorch/pipeline/tabular_classification.py
ArlindKadra: On another topic, what do you think the else is doing here in this code block @ravinkohli @nabenabe0928, what would be the usage?

autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
ArlindKadra: I think your code has a small problem if I validated it correctly, it is the case where there is a skip connection and there is no need for a shortcut to match the dimension change, so when the skip connection dimensions have not changed. And I think I addressed it, other than that it is better than was what before.

autoPyTorch/api/base_task.py
```diff
@@ -408,7 +412,7 @@ def _create_dask_client(self) -> None:
             dask.distributed.LocalCluster(
                 n_workers=self.n_jobs,
                 processes=True,
-                threads_per_worker=1,
+                threads_per_worker=2,
```
ravinkohli, line 415: should we have this as a user-defined parameter?

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -9,15 +9,99 @@
...
+    columns associated with their preprocessors.
+
+    Args:
+        preprocessors (Dict[str, List[BaseEstimator]]):
+            Dictionary containing list of numerical and categorical preprocessors.
+        numerical_columns (List[str]):
+            List of names of numerical columns
+        categorical_columns (List[str]):
+            List of names of categorical columns
+    Returns:
```
ravinkohli, line 39:
```suggestion

    Returns:
```

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -39,87 +123,54 @@ def _fit(
...
+            # this saves the index of the column.
             for i, type_ in enumerate(self.feat_type):
                 if 'numerical' in type_:
                     self.numerical_columns.append(i)
                 else:
                     self.categorical_columns.append(i)
 
         # Lastly, store the number of features
-        self.num_features = np.shape(X)[1]
+        self.num_features = len(X.columns)
```
ravinkohli, line 172: can we undo this change? As X can also be a sparse matrix and they don't have columns. The original code `np.shape(X)[1])` works for both dataframe and sparse matrices.

ravinkohli:
```suggestion
        self.num_features = np.shape(X)[1]
```

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -130,57 +181,77 @@ def transform(
...
+        # having a value for a categorical column.
+        # We need to convert the column in test data to
+        # object otherwise the test column is interpreted as float
+        if len(self.categorical_columns) > 0:
+            categorical_columns = self.column_transformer.transformers_[0][-1]
+            for column in categorical_columns:
+                if X[column].isna().all():
+                    X[column] = X[column].astype('object')
+
+        X = self.column_transformer.transform(X)
```
ravinkohli, line 254:
```suggestion
        if self.column_transformer is not None:
            X = self.column_transformer.transform(X)
```

ravinkohli: same, for sparse matrices we wont have a column transformer and hence we should check if it was created or not

autoPyTorch/api/base_task.py
ArlindKadra: yes, we should have it.

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
```diff
@@ -110,19 +131,36 @@ def shake_get_alpha_beta(is_training: bool, is_cuda: bool
...
+    """
+    The sampling of Bernoulli random variable
+    based on Eq. (4) in the paper
+    Args:
+        block_index (int): The index of the block from the input layer
+        min_prob_no_shake (float): The initial shake probability
+        num_blocks (int): The total number of building blocks
+        is_training (bool): Whether it is training
+        is_cuda (bool): Whether the tensor is on CUDA
+    Returns:
```
ravinkohli, line 149:
```suggestion

    Returns:
```

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -318,8 +318,12 @@ def on_epoch_end(self, X: Dict[str, Any], epoch: int) -> bool:
                 if self.use_snapshot_ensemble:
                     assert self.model_snapshots is not None, "model snapshots container can't be " \
                                                              "none when snapshot ensembling is enabled"
-                    model_copy = deepcopy(self.swa_model) if self.use_stochastic_weight_averaging \
-                        else deepcopy(self.model)
+                    is_last_epoch = (epoch == self.budget_tracker.max_epochs)
```
ravinkohli, line 321:
```suggestion
                    is_last_epoch = self.budget_tracker.is_max_epoch_reached(epoch)
```

autoPyTorch/pipeline/components/training/trainer/cutout_utils.py
```diff
@@ -63,8 +63,8 @@ def get_hyperparameter_search_space(
             dataset_properties: Optional[Dict] = None,
             weighted_loss: HyperparameterSearchSpace = HyperparameterSearchSpace(
                 hyperparameter="weighted_loss",
-                value_range=[True, False],
-                default_value=True),
+                value_range=(1, ),
```
ravinkohli, line 66: also here

ravinkohli (changes requested):
I have made some minor suggestions, otherwise it looks ready to be merged. Thanks for this PR.

autoPyTorch/api/base_task.py
ArlindKadra: This is weird, I have it as 2 on my branch?

autoPyTorch/pipeline/components/training/trainer/AdversarialTrainer.py
```diff
@@ -159,8 +159,8 @@ def get_hyperparameter_search_space(
         dataset_properties: Optional[Dict] = None,
         weighted_loss: HyperparameterSearchSpace = HyperparameterSearchSpace(
             hyperparameter="weighted_loss",
-            value_range=[True, False],
-            default_value=True),
+            value_range=(1, ),
```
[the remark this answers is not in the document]
ArlindKadra: We stick with it for what we will have as a branch for the paper. When we further progress with another branch and we evaluate that it is better we can update the reg_cocktail branch.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/scaling/base_scaler.py
```diff
@@ -28,5 +28,5 @@ def transform(self, X: Dict[str, Any]) -> Dict[str, Any]:
         if self.preprocessor['numerical'] is None and self.preprocessor['categorical'] is None:
             raise ValueError("cant call transform on {} without fitting first."
                              .format(self.__class__.__name__))
-        X.update({'scaler': self.preprocessor})
+        # X.update({'scaler': self.preprocessor})
```
nabenabe0928, line 31: no transformation occurs

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
```diff
@@ -92,15 +93,35 @@ def backward(ctx: typing.Any,
...
-def shake_get_alpha_beta(is_training: bool, is_cuda: bool
-                         ) -> typing.Tuple[torch.tensor, torch.tensor]:
-    if is_training:
+def shake_get_alpha_beta(
+    is_training: bool,
+    is_cuda: bool,
+    method: str
+) -> typing.Tuple[torch.Tensor, torch.Tensor]:
+    """
+    The methods used in this function have been introduced in 'ShakeShake Regularisation'
```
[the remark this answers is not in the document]
ArlindKadra: Sure, we can.

autoPyTorch/utils/implementations.py
```diff
@@ -35,7 +35,7 @@ def __call__(self, y: Union[np.ndarray, torch.Tensor]) -> np.ndarray:
...
-            classes, counts = classes[::-1], counts[::-1]
+            # classes, counts = classes[::-1], counts[::-1]
```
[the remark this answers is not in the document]
ArlindKadra: Sure, we can.

autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
```diff
@@ -72,6 +72,7 @@ def build_backbone(self, input_shape: Tuple[int, ...]) -> None:
             )
         if self.config['use_batch_norm']:
             layers.append(torch.nn.BatchNorm1d(self.config["num_units_%i" % self.config['num_groups']]))
+        layers.append(_activations[self.config["activation"]]())
```
nabenabe0928, line 75: We need it, right? (this change originates in this PR, right?)

autoPyTorch/pipeline/tabular_classification.py
ravinkohli: its for multilabel multi-class classification. But I don't think we have tested it yet.

autoPyTorch/utils/backend.py
```diff
@@ -205,12 +205,12 @@ def temporary_directory(self) -> str:
 
     def _make_internals_directory(self) -> None:
         try:
-            os.makedirs(self.internals_directory)
+            os.makedirs(self.internals_directory, exist_ok=True)
```
nabenabe0928, line 208: probably we should warn or ask users if we should allow to overwrite log files

autoPyTorch/pipeline/components/training/trainer/RowCutOutTrainer.py
```diff
@@ -36,34 +32,23 @@ def data_preparation(self, X: np.ndarray, y: np.ndarray,
...
-
-        # We use an ordinal encoder on the categorical columns of tabular data
-        # -1 is the conceptual equivalent to 0 in a image, that does not
-        # have color as a feature and hence the network has to learn to deal
-        # without this data. For numerical columns we use 0 to cutout the features
-        # similar to the effect that setting 0 as a pixel value in an image.
-        X[:, categorical_indices.long()] = self.CATEGORICAL_VALUE
-        X[:, numerical_indices.long()] = self.NUMERICAL_VALUE
+        size: int = np.shape(X)[1]
+        indices = self.random_state.choice(
```
nabenabe0928, line 36: `indices` -> `shuffled_indices` for the compatibility with row-cut-mix trainer

nabenabe0928 (commented):
The review at [the version right after the merge](https://github.com/automl/Auto-PyTorch/pull/249/commits/23466f0c7210db7b1d946f5accd086d0a78b9e98).
I have not checked:
- [ ] base_task.py
- [ ] tabular_feature_validator.py
- [ ] ResNetBackbone.py
- [x] network_backbone/utils.py

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/base_imputer.py
```diff
@@ -29,5 +29,5 @@ def transform(self, X: Dict[str, Any]) -> Dict[str, Any]:
         if self.preprocessor['numerical'] is None and self.preprocessor['categorical'] is None:
             raise ValueError("cant call transform on {} without fitting first."
                              .format(self.__class__.__name__))
-        X.update({'imputer': self.preprocessor})
```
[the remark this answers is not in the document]
ravinkohli: I think we have discussed that i'll make another PR for preprocessing after this is merged to fix this.

autoPyTorch/utils/backend.py
ravinkohli: Yeah I had a warning in the PR I made to automl common. Maybe we can copy it as is.

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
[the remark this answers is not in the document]
nabenabe0928:
```suggestion
    """
    The methods used in this function have been introduced in 'ShakeShake Regularisation'
    Each method name is available in the referred paper.
    Currently, this function supports `even-even`, `shake-even`, `shake-shake` and `M3`.

    Args:
        is_training (bool): Whether the computation for the training
        is_cuda (bool): Whether the tensor is on CUDA
        method (str): The shake method either `even-even`, `shake-even`, `shake-shake` or `M3`

    Returns:
        alpha, beta (Tuple[float, float]):
            alpha (in [0, 1]) is the weight coefficient  for the forward pass
            beta (in [0, 1]) is the weight coefficient for the backward pass

    Reference:
        Title: Shake-shake regularization
        Author: Xavier Gastaldi
        URL: https://arxiv.org/abs/1705.07485
    """
```

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
```diff
@@ -92,15 +93,35 @@ def backward(ctx: typing.Any,
...
+    elif method.endswith('shake'):
+        beta = torch.rand(1)
+    elif method == 'M3':
+        # Table 4 in the paper `Shake-Shake regularization`
+        rnd = torch.rand(1)
+        beta = torch.FloatTensor(
+            [rnd * (0.5 - alpha) + alpha if alpha < 0.5 else rnd * (alpha - 0.5) + 0.5]
+        )
+    else:
+        raise ValueError("Unknown method for ShakeShakeRegularisation in NetworkBackbone")
```
nabenabe0928, line 124:
```suggestion
        raise ValueError(f"Unknown method `{method}` for ShakeShakeRegularisation in NetworkBackbone")
```

autoPyTorch/data/base_feature_validator.py
```diff
@@ -35,60 +35,61 @@ class BaseFeatureValidator(BaseEstimator):
...
-        self.dtypes = []  # type: typing.List[str]
-        self.column_order = []  # type: typing.List[str]
+        self.feat_type: Optional[List[str]] = None
+        self.data_type: Optional[type] = None
+        self.dtypes: List[str] = []
+        self.column_order: List[str] = []
 
-        self.encoder = None  # type: typing.Optional[BaseEstimator]
-        self.enc_columns = []  # type: typing.List[str]
+        self.encoder: Optional[BaseEstimator] = None
```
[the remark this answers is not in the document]
ArlindKadra: Okay, I removed it.

autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
ArlindKadra: Yes, makes it consistent with the `ResNetBackbone`.

autoPyTorch/utils/backend.py
ArlindKadra: Would you like to commit it then @ravinkohli ?

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/encoding/base_encoder.py
```diff
@@ -28,5 +28,5 @@ def transform(self, X: Dict[str, Any]) -> Dict[str, Any]:
         if self.preprocessor['numerical'] is None and self.preprocessor['categorical'] is None:
             raise ValueError("cant call transform on {} without fitting first."
                              .format(self.__class__.__name__))
-        X.update({'encoder': self.preprocessor})
+        # X.update({'encoder': self.preprocessor})
```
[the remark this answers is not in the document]
ArlindKadra: @ravinkohli  can confirm, but it is the same answer as before if I am not mistaken.

autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
```diff
@@ -310,48 +320,51 @@ def _build_block(self, in_features: int, out_features: int) -> nn.Module:
...
-                x1 = self.layers(x)
-                x2 = self.shake_shake_layers(x)
-                alpha, beta = shake_get_alpha_beta(self.training, x.is_cuda)
-                x = shake_shake(x1, x2, alpha, beta)
+            residual = self.shortcut(x)
+        else:
+            if not self.config["use_skip_connection"]:
+                # Early-return
+                return self.layers(x)
             else:
```
nabenabe0928, line 339: I think this order (reversed) is easier to understand.
```suggestion
        elif self.config["use_skip_connection"]:
            # We use a skip connection but we do not need to match dimensions
            residual = x
        else:  # Early-return because no need of skip connection
            return self.layers(x)
```

autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
```diff
@@ -310,48 +320,51 @@ def _build_block(self, in_features: int, out_features: int) -> nn.Module:
...
+                is_training=self.training,
+                is_cuda=x.is_cuda,
+                method=self.config['shake_shake_update_func'],
+            )
+            x = shake_shake(x1, x2, alpha, beta)
+        elif self.config["multi_branch_choice"] == 'shake-drop':
+            x = self.layers(x)
+            alpha, beta = shake_get_alpha_beta(
+                self.training,
+                x.is_cuda,
```
nabenabe0928, line 356:
```suggestion
                is_training=self.training,
                is_cuda=x.is_cuda,
```

autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
nabenabe0928: Thanks, I did not check the process before the lines!

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
```diff
@@ -98,8 +98,27 @@ def shake_get_alpha_beta(
...
+            beta (in [0, 1]) is the weight coefficient for the backward pass
+
+    Reference:
+        Title: Shake-shake regularization
+        Author: Xavier Gastaldi
+        URL: https://arxiv.org/abs/1705.07485
+    """
     https://arxiv.org/abs/1705.07485. The names have been taken from the paper as well.
     Currently, this function supports `even-even`, `shake-even`, `shake-shake` and `M3`.
     """
```
nabenabe0928: Can we do like this?:
```
# current looking
"""
doc-string A
"""
doc-string B
"""
doc-string C
"""

# After modification
"""
doc-string B
"""
```

autoPyTorch/api/tabular_classification.py
```diff
@@ -27,11 +27,14 @@
 class TabularClassificationTask(BaseTask):
     """
     Tabular Classification API to the pipelines.
+
     Args:
         seed (int):
             seed to be used for reproducibility.
         n_jobs (int), (default=1):
             number of consecutive processes to spawn.
+        nr_threads (int), (default=1):
```
nabenabe0928, line 36: `n_threads` for the compatibility?

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
```diff
@@ -111,15 +112,54 @@ def backward(ctx: typing.Any,
...
 
-def shake_get_alpha_beta(is_training: bool, is_cuda: bool
-                         ) -> typing.Tuple[torch.tensor, torch.tensor]:
-    if is_training:
+def shake_get_alpha_beta(
+    is_training: bool,
+    is_cuda: bool,
+    method: str
+) -> typing.Tuple[torch.Tensor, torch.Tensor]:
+    """
```
nabenabe0928, line 120:
```suggestion
```

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
```diff
@@ -111,15 +112,54 @@ def backward(ctx: typing.Any,
...
+            beta (in [0, 1]) is the weight coefficient for the backward pass
+
+    Reference:
+        Title: Shake-shake regularization
+        Author: Xavier Gastaldi
+        URL: https://arxiv.org/abs/1705.07485
+    """
+    https://arxiv.org/abs/1705.07485. The names have been taken from the paper as well.
+    Currently, this function supports `even-even`, `shake-even`, `shake-shake` and `M3`.
+    """
```
nabenabe0928, line 143:
```suggestion
```

autoPyTorch/api/base_task.py
```diff
@@ -101,7 +101,10 @@ class BaseTask:
     Base class for the tasks that serve as API to the pipelines.
     Args:
         seed (int), (default=1): seed to be used for reproducibility.
-        n_jobs (int), (default=1): number of consecutive processes to spawn.
+        n_jobs (int), (default=1):
+            number of consecutive processes to spawn.
+        nr_threads (int), (default=1):
```
nabenabe0928, line 106: `n_threads` for the compatibility?

autoPyTorch/utils/backend.py
ravinkohli: actually, I didn't do it there. but I can add it definitely.

autoPyTorch/pipeline/components/training/trainer/RowCutOutTrainer.py
ArlindKadra: masked_indices would be good I guess and is in between of both things

ravinkohli: I think having only indices is a bit confusing. Let's stick with mentioning that these are column indices. Also, cut is fine because that's also the name of the algorithm, although I don't feel too strongly about this point. But for consistency, we can stick with cut_column_indices as we are using in `RowCutMixTrainer`

autoPyTorch/api/base_task.py
```diff
@@ -1333,6 +1367,205 @@ def fit_pipeline(self,
...
+                Model fitting will be terminated if the machine
+                learning algorithm runs over the time limit. Set
+                this value high enough so that typical machine
+                learning algorithms can be fit on the training
+                data.
+                When set to None, this time will automatically be set to
+                total_walltime_limit // 2 to allow enough time to fit
+                at least 2 individual machine learning algorithms.
+                Set to np.inf in case no time limit is desired.
+        Returns:
```
ravinkohli, line 1421:
```suggestion

        Returns:
```

autoPyTorch/utils/backend.py
ravinkohli: I have done that.

ArlindKadra: Thanks.

autoPyTorch/pipeline/components/training/trainer/RowCutOutTrainer.py
ArlindKadra: Sure then, I agree.

autoPyTorch/api/tabular_classification.py
ArlindKadra: Sure, done.

PR #251, opened 2021-06-10
[ADD] Enable long running regression

franchuterivera:
The idea is to trigger a regression every Tuesday that will take the greedy and default portfolio and make sure we achieve a deterministic performance.


ravinkohli (approved):
Hey, thanks for this PR. It enables us to ensure a minimum performance for our portfolio and the API in general. I think this can be merged

PR #256, opened 2021-06-15
[feat] Add flexible step-wise LR scheduler with minimum changes

nabenabe0928:
Since this PR is urgent now, I made another PR to quickly merge the feature.
The original PR is available [here](https://github.com/automl/Auto-PyTorch/pull/212/files).


ravinkohli (changes requested):
Hey, Thanks for this PR. The changes look good. However, I still don't see where epoch or batch is passed by an LR scheduler. So basically, how does the fit dictionary 'X' get updated to contain the step unit? To that end, could we add a test where we run a few schedulers in a training and we verify that the step was made at the appropriate time. This can be achieved by using the internal parameters of the torch scheduler for example, in `CosineAnnealingWarmRestarts` we have `T_0` or `T_curr` which contains the epochs since the last step. And we can verify if it is done epoch wise or batch wise there.

nabenabe0928:
> However, I still don't see where epoch or batch is passed by an LR scheduler. So basically, how does the fit dictionary 'X' get updated to contain the step unit? 

I will check promptly and give another commit soon

> To that end, could we add a test where we run a few schedulers in a training and we verify that the step was made at the appropriate time. 

Should I do it separetly as in the latest commit of test codes by me or should I really run it from API?
I think the former is sufficient and faster, but if something wrong happens in the system, we cannot potentially detect the errors. 
But in my opinion, when you make tests, they should test speicific parts, so the former is enough. 
What do you think?

ravinkohli:
> > However, I still don't see where epoch or batch is passed by an LR scheduler. So basically, how does the fit dictionary 'X' get updated to contain the step unit?
> 
> I will check promptly and give another commit soon
> 
> > To that end, could we add a test where we run a few schedulers in a training and we verify that the step was made at the appropriate time.
> 
> Should I do it separetly as in the latest commit of test codes by me or should I really run it from API?
> I think the former is sufficient and faster, but if something wrong happens in the system, we cannot potentially detect the errors.
> But in my opinion, when you make tests, they should test speicific parts, so the former is enough.
> What do you think?

yeah so the idea is to train a pipeline with the different schedulers and make sure that the step is made at the right time. The code you added in itself might run, but the tests I would like to see would ensure that proper information is passed by the chosen lr scheduler and it is able to take a step at the appropriate time. This will ensure that  CosineAnnealing or some other scheduler if selected in a configuration is able to take the appropriate step

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -20,6 +21,22 @@
 from autoPyTorch.utils.implementations import get_loss_weight_strategy
 
 
+class StepIntervalUnit(Enum):
```
ravinkohli, line 24: hey as this class is something that is used in lr schedulers as well and more importantly, this is something to do with the lr schedulers. Do you think it makes sense to move it to a utils file of the lr_scheduler folder?

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -388,18 +423,20 @@ def evaluate(self, test_loader: torch.utils.data.DataLoader, epoch: int,
                         epoch * len(test_loader) + step,
                     )
 
+        self._scheduler_step(step_interval=StepIntervalUnit.valid, loss=loss_sum / N)
```
ravinkohli, line 426: why do we need this?

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -254,11 +273,29 @@ def on_epoch_end(self, X: Dict[str, Any], epoch: int) -> bool:
...
+        loss: Optional[float] = None
+    ) -> None:
+
+        if self.step_interval != step_interval:
+            return
+
+        if not self.scheduler:  # skip if no scheduler defined
+            return
+
+        try:
```
ravinkohli, line 288: I think it was better before. why do you want to catch an exception and then handle it?

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
nabenabe0928: It is from [ReduceLROnPlateau](https://pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.ReduceLROnPlateau.html#torch.optim.lr_scheduler.ReduceLROnPlateau).
It says step should be performed after validation.

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
nabenabe0928: because we do not know what kind of `lr_scheduler` we add in the future.
In reality, I would like to have got the arguments for each function, but the simplest way I can achieve this was to catch by TypeError.

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
nabenabe0928: It makes sense
I will do it

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
ravinkohli: Okay, I see.  Could you create an issue for this? So we don't forget to update this. Moreover, as we allow the users to add an LR scheduler, we should think of a neat way to pass this info.

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
ravinkohli: oh, I didn't know that. Thanks.

ravinkohli (approved):

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
nabenabe0928: Here, I posted
https://github.com/automl/Auto-PyTorch/issues/270

ravinkohli (approved):

PR #260, opened 2021-06-17
[ADD] Post-Hoc ensemble fitting

ravinkohli:
Adds a fit ensemble method enabling users to post hoc build an ensemble after search is done. This feature allows the user to run a search and sequentially build an ensemble later.  This PR also adds tests for the functionality as well as updating `traditional_run_history.json`.

### DIFF
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -786,7 +786,7 @@ class BaseTask:
                 metrics supporting current task will be calculated
                 for each pipeline and results will be available via cv_results
             precision (int), (default=32): Numeric precision used when loading
-                ensemble data. Can be either '16', '32' or '64'.
+                ensemble data. Can be either 16, 32 or 64.
             disable_file_output (Union[bool, List]):
             load_models (bool), (default=True): Whether to load the
                 models after fitting AutoPyTorch.
@@ -931,12 +931,12 @@ class BaseTask:
             self._logger.info("Starting ensemble")
             ensemble_task_name = 'ensemble'
             self._stopwatch.start_task(ensemble_task_name)
-            proc_ensemble = self._fit_ensemble(time_left_for_ensembles=time_left_for_ensembles,
-                                               ensemble_size=self.ensemble_size,
-                                               ensemble_nbest=self.ensemble_nbest,
-                                               precision=precision,
-                                               optimize_metric=self.opt_metric
-                                               )
+            proc_ensemble = self._init_ensemble_builder(time_left_for_ensembles=time_left_for_ensembles,
+                                                        ensemble_size=self.ensemble_size,
+                                                        ensemble_nbest=self.ensemble_nbest,
+                                                        precision=precision,
+                                                        optimize_metric=self.opt_metric
+                                                        )
 
         # ==> Run SMAC
         smac_task_name: str = 'runSMAC'
@@ -1174,14 +1174,32 @@ class BaseTask:
     def fit_ensemble(
         self,
         ensemble_nbest: Optional[int] = None,
-        ensemble_size: Optional[int] = None,
+        ensemble_size: int = 50,
         precision: int = 32
     ) -> 'BaseTask':
+        """
+        Enables post-hoc fitting of the ensemble after the `search()`
+        method is finished. This method creates an ensemble using all
+        the models stored on disk during the smbo run
+        Args:
+            ensemble_nbest (Optional[int]):
+                only consider the ensemble_nbest models to build the ensemble.
+                If None, uses the value stored in class attribute `ensemble_nbest`.
+            ensemble_size (int) (default=50):
+                Number of models added to the ensemble built by
+                Ensemble selection from libraries of models.
+                Models are drawn with replacement.
+            precision (int), (default=32): Numeric precision used when loading
+                ensemble data. Can be either 16, 32 or 64.
 
+        Returns:
+            self
+        """
         # Make sure that input is valid
         if self.dataset is None or self.opt_metric is None:
-            raise ValueError("fit_ensemble() can only be called after fit. Please call the "
-                             "estimator search() method prior to fit_ensemble().")
+            raise ValueError("fit_ensemble() can only be called after `search()`. "
+                             "Please call the `search()` method of {} prior to "
+                             "fit_ensemble().".format(self.__class__.__name__))
 
         if self._logger is None:
             self._logger = self._get_logger(self.dataset.dataset_name)
@@ -1192,11 +1210,11 @@ class BaseTask:
         else:
             self._is_dask_client_internally_created = False
 
-        manager = self._fit_ensemble(
+        manager = self._init_ensemble_builder(
             time_left_for_ensembles=self._time_for_task,
             optimize_metric=self.opt_metric,
             precision=precision if precision is not None else self.precision,
-            ensemble_size=ensemble_size if ensemble_size is not None else self.ensemble_size,
+            ensemble_size=ensemble_size,
             ensemble_nbest=ensemble_nbest if ensemble_nbest is not None else self.ensemble_nbest,
         )
 
@@ -1204,15 +1222,15 @@ class BaseTask:
         future = manager.futures.pop()
         result = future.result()
         if result is None:
-            raise ValueError("Error building the ensemble - please check the log file and command "
-                             "line output for error messages.")
+            raise ValueError("Errors occurred while building the ensemble - please"
+                             " check the log file and command line output for error messages.")
         self.ensemble_performance_history, _, _, _ = result
 
         self._load_models()
         self._close_dask_client()
         return self
 
-    def _fit_ensemble(
+    def _init_ensemble_builder(
         self,
         time_left_for_ensembles: float,
         optimize_metric: str,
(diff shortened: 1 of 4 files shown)


autoPyTorch/api/base_task.py
```diff
@@ -1180,6 +1169,95 @@ def fit(self,
...
+        optimize_metric: str,
+        ensemble_nbest: int,
+        ensemble_size: int,
+        precision: int = 32,
+    ) -> EnsembleBuilderManager:
+
+        assert self._logger is not None, "logger should be initialised to fit ensemble"
+        if self.dataset is None:
+            raise ValueError("ensemble can only be fitted after fit. Please call the "
+                             "estimator search() method prior to fit_ensemble().")
```
nabenabe0928, line 1225: Can we clarify which search() method?
For example, users would like to know which `class` or `file` to check.

autoPyTorch/api/base_task.py
```diff
@@ -1180,6 +1169,95 @@ def fit(self,
...
+            precision=precision if precision is not None else self.precision,
+            ensemble_size=ensemble_size if ensemble_size is not None else self.ensemble_size,
+            ensemble_nbest=ensemble_nbest if ensemble_nbest is not None else self.ensemble_nbest,
+        )
+
+        manager.build_ensemble(self._dask_client)
+        future = manager.futures.pop()
+        result = future.result()
+        if result is None:
+            raise ValueError("Error building the ensemble - please check the log file and command "
```
nabenabe0928, line 1205: Errors occurred during building the ensemble.

autoPyTorch/api/base_task.py
```diff
@@ -183,6 +183,9 @@ def __init__(
         self.trajectory: Optional[List] = None
         self.dataset_name: Optional[str] = None
         self.cv_models_: Dict = {}
+        self.precision: Optional[int] = None
+        self.opt_metric: Optional[str] = None
+        self.dataset: Optional[BaseDataset] = None
```
nabenabe0928, line 188: Although each `BaseDataset` shares the same address, it is not a good idea to have `BaseDataset` everywhere (because we do not know when and where and how it can change).
So can we take `self.dataset` as an argument when we would like to use?

autoPyTorch/api/base_task.py
```diff
@@ -1180,6 +1169,95 @@ def fit(self,
         self._clean_logger()
         return pipeline
 
+    def fit_ensemble(
```
nabenabe0928, line 1172: Is this the post-hoc ensemble fitting?
It is confusing for me to have `_fit_ensemble` and `fit_ensemble`.

nabenabe0928 (changes requested):
Thanks for the PR.
I left some comments.

autoPyTorch/api/base_task.py
ravinkohli: so `self.dataset` is only modified in the task.search method. It is already an argument to the functions `fit` and `refit` so that allows the user to use a different dataset to refit all the models or even fitting a pipeline. and the `fit_ensemble` as it is intended to only be used after `search()` does not have an option to take a datasert as we need to ensure its the same dataset as the search was called on. Taking it as an argument does not seem to provide that much advantanges especially considering creating a dataset is hard, we dont have a function currently that will create this dataset and the feature validators and fit and transform them for the user. And needing to give a dataset object for each function would thus become tougher. I have merged a `fit_pipeline` [PR](https://github.com/automl/Auto-PyTorch/pull/173) to the cocktails branch which would introduce the create_dataset method. I'll soon make a PR to add this to `development` as well. Maybe we can revisit this conversation then?

autoPyTorch/api/base_task.py
ravinkohli: yeah let me fix the documentation for this, sorry. But basically, `_fit_ensemble` is a private method which helps to not repeat the same code as it is being done in the`search` method already. I'll change the name for this method to also clear the confusion

autoPyTorch/api/base_task.py
ravinkohli: sure

autoPyTorch/api/base_task.py
```diff
@@ -1182,6 +1171,133 @@ def fit(self,
...
+        self,
+        ensemble_nbest: Optional[int] = None,
```
franchuterivera, line 1176: can you make this 50, not optional?

autoPyTorch/api/base_task.py
```diff
@@ -1182,6 +1171,133 @@ def fit(self,
...
+        manager.build_ensemble(self._dask_client)
+        future = manager.futures.pop()
+        result = future.result()
+        if result is None:
+            raise ValueError("Errors occurred while building the ensemble - please"
+                             " check the log file and command line output for error messages.")
+        self.ensemble_performance_history, _, _, _ = result
+
+        self._load_models()
+        self._close_dask_client()
```
franchuterivera, line 1230: I think we should also close the logger. Could you create please a function that is called like `clean_up()` that closes both dask and the logger?

autoPyTorch/api/base_task.py
```diff
@@ -1182,6 +1171,133 @@ def fit(self,
...
+                Ensemble selection from libraries of models.
+                Models are drawn with replacement.
+            precision (int), (default=32): Numeric precision used when loading
+                ensemble data. Can be either 16, 32 or 64.
+
+        Returns:
+            EnsembleBuilderManager
+
+        """
+        assert self._logger is not None, "logger should be initialised to fit ensemble"
```
franchuterivera, line 1262: Probably this should be a value error

requirements.txt
```diff
@@ -4,7 +4,7 @@ torchvision
 tensorboard
 scikit-learn>=0.24.0,<0.25.0
 numpy
-scipy
+scipy<1.7.0
```
franchuterivera, line 7: I think you are gonna update this one.

test/test_api/test_api.py
```diff
@@ -714,3 +714,61 @@ def test_build_pipeline(api_type, fit_dictionary_tabular):
...
+    estimator.fit_ensemble(ensemble_size=2)
+    assert isinstance(estimator.ensemble_performance_history, list)
+    assert 'train_accuracy' in estimator.ensemble_performance_history[0]
+    assert 'test_accuracy' in estimator.ensemble_performance_history[0]
+
+    assert os.path.exists(os.path.join(estimator._backend.internals_directory, 'ensembles'))
+    assert len(os.listdir(os.path.join(estimator._backend.internals_directory, 'ensembles'))) > 0
+    assert any(['.ensemble' in file for file in os.listdir(os.path.join(
+        estimator._backend.internals_directory, 'ensembles'))])
+    assert any(['ensemble_' or '_ensemble.npy' in os.listdir(estimator._backend.internals_directory)])
```
franchuterivera, line 759: Can you do predict from the estimator? This will tell us that the ensemble got written to disc and we can read it back

test/test_api/test_api.py
```diff
@@ -714,3 +714,61 @@ def test_build_pipeline(api_type, fit_dictionary_tabular):
     pipeline = api.build_pipeline(fit_dictionary_tabular['dataset_properties'])
     assert isinstance(pipeline, BaseEstimator)
     assert len(pipeline.steps) > 0
+
+
+@unittest.mock.patch('autoPyTorch.evaluation.train_evaluator.eval_function',
+                     new=dummy_eval_function)
+@pytest.mark.parametrize('dataset_name', ('iris',))
+def test_fit_ensemble(backend, n_samples, dataset_name):
```
franchuterivera, line 722: Can you also check that the ensemble history got populated correctly with at least one run?

franchuterivera (changes requested):
Thanks a lot for the changes. They look good but I am requesting mainly two things to make it even better:

Add an example where we fit the API sequentially. First search is called with 1 core, then fit the ensemble

Then add some extra checks as mentioned to gain more confidence on the chance.

autoPyTorch/api/base_task.py
nabenabe0928: Thanks for the clarification:)
In fact, this is very general discussion.
It is just my opinion and it will be great if we can revisit some times soon.
FYI, it is worth taking a look at the concept "encapsulation of object" and "unsafety from multiple inheritance".

nabenabe0928 (commented):
I did not find large issues except what Francisco already mentioned.
I will wait for another major change.

autoPyTorch/api/base_task.py
```diff
@@ -1259,7 +1274,8 @@ def _init_ensemble_builder(
             EnsembleBuilderManager
 
         """
-        assert self._logger is not None, "logger should be initialised to fit ensemble"
+        if self._logger is None:
+            raise ValueError("logger should be initialised to fit ensemble")
```
nabenabe0928, line 1278: It does not matter so much, but I think you often use American English, right?
Then let's go by `initialized`

nabenabe0928, line 1278: It is not important, but can we go by American English as you seem to often use American spelling.
`initialised` -> `initialized`

autoPyTorch/api/base_task.py
```diff
@@ -1182,6 +1171,95 @@ def fit(self,
...
+    def _fit_ensemble(
+        self,
+        time_left_for_ensembles: float,
+        optimize_metric: str,
+        ensemble_nbest: int,
+        ensemble_size: int,
+        precision: int = 32,
+    ) -> EnsembleBuilderManager:
+
+        assert self._logger is not None, "logger should be initialised to fit ensemble"
```
nabenabe0928: initialized

autoPyTorch/api/base_task.py
ravinkohli:
```suggestion
            raise ValueError("logger should be initialized to fit ensemble")
```

Issue #110, opened 2021-02-23
Remove the temporal .pth files left by early stopping


franchuterivera:
We use the temporary directory python feature, but it does not automatically clean the file after use.

To save disk space, delete all .pth files that are no longer needed, saved under the backend.temproary_directory area.

PR #261, opened 2021-06-18
Remove temporal paths left out during training.

franchuterivera:
Addresses #110

### DIFF
--- a/autoPyTorch/pipeline/components/training/trainer/__init__.py
+++ b/autoPyTorch/pipeline/components/training/trainer/__init__.py
@@ -1,6 +1,7 @@
 import collections
 import logging.handlers
 import os
+import shutil
 import tempfile
 import time
 from typing import Any, Dict, List, Optional, Tuple, cast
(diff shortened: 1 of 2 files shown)


ravinkohli (approved):
Thanks for this PR. I think it can be merged

autoPyTorch/pipeline/components/training/trainer/__init__.py
```diff
@@ -384,12 +395,6 @@ def early_stop_handler(self, X: Dict[str, Any]) -> bool:
             torch.save(X['network'].state_dict(), best_path)
 
         if epochs_since_best > X['early_stopping']:
```
nabenabe0928, line 397: `return epochs_since_best > X['early_stopping']`

nabenabe0928 (changes requested):
Thanks for the PR.
I suggested one minor change.

autoPyTorch/pipeline/components/training/trainer/__init__.py
franchuterivera: Thanks for the suggestion, I have incorporated this.

PR #263, opened 2021-06-18
Fixes problems with weighted cross-entropy

franchuterivera:
address #248

### DIFF
--- a/autoPyTorch/utils/implementations.py
+++ b/autoPyTorch/utils/implementations.py
@@ -25,17 +25,17 @@ class LossWeightStrategyWeighted():
     def __call__(self, y: Union[np.ndarray, torch.Tensor]) -> np.ndarray:
         if isinstance(y, torch.Tensor):
             y = y.detach().cpu().numpy() if y.is_cuda else y.numpy()
-        if isinstance(y[0], str):
-            y = y.astype('float64')
         counts = np.sum(y, axis=0)
         total_weight = y.shape[0]
 
-        if len(y.shape) > 1:
+        if len(y.shape) > 1 and y.shape[1] != 1:
+            # In this case, the second axis represents classes
             weight_per_class = total_weight / y.shape[1]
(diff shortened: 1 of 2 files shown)


ravinkohli (approved):
Hey, this PR looks great. I think this can be merged. Thanks for the added tests to ensure we don't mess this up going forward.

autoPyTorch/utils/implementations.py
```diff
@@ -50,10 +50,8 @@ class LossWeightStrategyWeightedBinary():
     def __call__(self, y: Union[np.ndarray, torch.Tensor]) -> np.ndarray:
         if isinstance(y, torch.Tensor):
             y = y.detach().cpu().numpy() if y.is_cuda else y.numpy()
-        if isinstance(y[0], str):
-            y = y.astype('float64')
         counts_one = np.sum(y, axis=0)
-        counts_zero = counts_one + (-y.shape[0])
+        counts_zero = y.shape[0] - counts_one
         weights = counts_zero / np.maximum(counts_one, 1)
```
nabenabe0928, line 55: Where do we use those weights?
If we weight the data with label 0 by this weights, `counts_zero / y.shape[0]` seems correct to me.

nabenabe0928 (commented):
Thanks for the PR.
I put a comment regarding the validity of the calculation.
Please check the current implementation.

autoPyTorch/utils/implementations.py
franchuterivera: This follows the recommendation from pytorch, please look into [this](https://pytorch.org/docs/stable/generated/torch.nn.BCEWithLogitsLoss.html).

The problem nevertheless here was that we had negative weights. For this reason I changed line 54. Line 55 Looks ok for me, but let me know if you find otherwise.

franchuterivera: Thanks for the suggestion, I have done so now.

PR #264, opened 2021-06-18
Enable python 3.9

franchuterivera:
This PR enables python 3.9

### DIFF
--- a/autoPyTorch/pipeline/components/setup/lr_scheduler/CosineAnnealingWarmRestarts.py
+++ b/autoPyTorch/pipeline/components/setup/lr_scheduler/CosineAnnealingWarmRestarts.py
@@ -19,7 +19,7 @@ from autoPyTorch.utils.common import HyperparameterSearchSpace, add_hyperparamet
 class CosineAnnealingWarmRestarts(BaseLRComponent):
     """
     Set the learning rate of each parameter group using a cosine annealing schedule,
-    where \eta_{max}ηmax is set to the initial lr, T_{cur} is the number of epochs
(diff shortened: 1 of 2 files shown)


ravinkohli (approved):
Thanks for this PR. I think it can be merged

autoPyTorch/pipeline/components/setup/lr_scheduler/CosineAnnealingWarmRestarts.py
```diff
@@ -16,7 +16,7 @@
 class CosineAnnealingWarmRestarts(BaseLRComponent):
     """
     Set the learning rate of each parameter group using a cosine annealing schedule,
-    where \eta_{max}ηmax is set to the initial lr, T_{cur} is the number of epochs
```
nabenabe0928, line 19: Why did you delete `\` ?

The usage of the Inline math represetations is nicely described [here](https://numpydoc.readthedocs.io/en/latest/format.html#notes).

```
where :math:`\eta_{\max}` is set to the initial lr, :math:`T_{\text{cur}}` is the number of epochs
since the last restart and :math:`T_{i}` is the number of epochs between two warm
```

nabenabe0928 (changes requested):
Thanks for the PR.
I put a small comment.

autoPyTorch/pipeline/components/setup/lr_scheduler/CosineAnnealingWarmRestarts.py
franchuterivera: Hello, sorry this was not an intentional change. 

My vim configuration changed this. It did so because this should be a raw string, as `\e` is not a valid escaped sequence.

PR #272, opened 2021-06-23
[Fix] long running regression

franchuterivera:
Pytest approximate does not work with compare equal.

Also, simplified the additional_metrics code, which caused problem to the long regression.

### DIFF
--- a/autoPyTorch/pipeline/components/training/trainer/__init__.py
+++ b/autoPyTorch/pipeline/components/training/trainer/__init__.py
@@ -249,14 +249,16 @@ class TrainerChoice(autoPyTorchChoice):
         )
 
         # Support additional user metrics
-        additional_metrics = X['additional_metrics'] if 'additional_metrics' in X else []
-        if 'optimize_metric' in X:
-            additional_metrics.append(X['optimize_metric'])
+        metrics = get_metrics(dataset_properties=X['dataset_properties'])
+        print(type(metrics[0]))
+        if 'additional_metrics' in X:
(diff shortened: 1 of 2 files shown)


autoPyTorch/pipeline/components/training/trainer/__init__.py
```diff
@@ -249,15 +249,15 @@ def _fit(self, X: Dict[str, Any], y: Any = None, **kwargs: Any) -> 'TrainerChoic
...
 
         # Support additional user metrics
-        additional_metrics = X['additional_metrics'] if 'additional_metrics' in X else None
-        if 'optimize_metric' in X:
-            additional_metrics = additional_metrics.append(X['optimize_metric']) if additional_metrics is not None \
-                else [X['optimize_metric']]
+        metrics = get_metrics(dataset_properties=X['dataset_properties'])
+        if 'additional_metrics' in X:
+            metrics.extend(get_metrics(dataset_properties=X['dataset_properties'], names=X['additional_metrics']))
+        if 'optimize_metric' in X and 'optimize_metric' not in [m.name for m in metrics]:
```
nabenabe0928, line 255: Can you put a single-line comment for what `'optimize_metric' not in [m.name for m in metrics]` this means?

nabenabe0928 (commented):
Thanks for the PR.
I would like to ask you how this change relates to `long running regression`?

autoPyTorch/pipeline/components/training/trainer/__init__.py
ravinkohli, line 255:
```suggestion
        if 'optimize_metric' in X and X['optimize_metric'] not in [m.name for m in metrics]:
```

PR #286, opened 2021-08-04
Cocktail fixes time debug

ravinkohli:
This PR addresses the time issue faced by the library where the data is being preprocessed for every batch which is taking an exceptionally long time for larger datasets. 
Additionally, I have moved the code from PR #260 and closed that PR as we needed the code for one of our experiments. I have verified the functionality as well as added an example of how to use this functionality. 
This PR also updates the torch type usage that caused errors when using `torch> 1.8`.

### DIFF
--- a/autoPyTorch/data/tabular_feature_validator.py
+++ b/autoPyTorch/data/tabular_feature_validator.py
@@ -110,7 +110,7 @@ class TabularFeatureValidator(BaseFeatureValidator):
         features (from categorical for example) to a numerical value that further stages
         will be able to use
 
-        Arguments:
+        Args:
             X (SUPPORTED_FEAT_TYPES):
                 A set of features that are going to be validated (type and dimensionality
                 checks) and a encoder fitted in the case the data needs encoding
@@ -180,7 +180,7 @@ class TabularFeatureValidator(BaseFeatureValidator):
         Validates and fit a categorical encoder (if needed) to the features.
         The supported data types are List, numpy arrays and pandas DataFrames.
 
-        Arguments:
+        Args:
             X_train (SUPPORTED_FEAT_TYPES):
                 A set of features, whose categorical features are going to be
                 transformed
@@ -279,7 +279,7 @@ class TabularFeatureValidator(BaseFeatureValidator):
         """
         Feature dimensionality and data type checks
 
-        Arguments:
+        Args:
             X (SUPPORTED_FEAT_TYPES):
                 A set of features that are going to be validated (type and dimensionality
                 checks) and a encoder fitted in the case the data needs encoding
@@ -355,7 +355,7 @@ class TabularFeatureValidator(BaseFeatureValidator):
         """
         Return the columns to be encoded from a pandas dataframe
 
-        Arguments:
+        Args:
             X (pd.DataFrame)
                 A set of features that are going to be validated (type and dimensionality
                 checks) and a encoder fitted in the case the data needs encoding
@@ -425,7 +425,7 @@ class TabularFeatureValidator(BaseFeatureValidator):
 
         If test data is provided, we proactively match it to train data
 
-        Arguments:
+        Args:
             X_train (SUPPORTED_FEAT_TYPES):
                 A set of features that are going to be validated (type and dimensionality
                 checks) and a encoder fitted in the case the data needs encoding
@@ -459,7 +459,7 @@ class TabularFeatureValidator(BaseFeatureValidator):
         """
         Converts a numpy array to pandas for type inference
 
-        Arguments:
+        Args:
             X (np.ndarray):
                 data to be interpreted.
 
@@ -474,7 +474,7 @@ class TabularFeatureValidator(BaseFeatureValidator):
 
         This has to be done once, so the test and train data are treated equally
 
-        Arguments:
+        Args:
             X (pd.DataFrame):
                 data to be interpreted.
 
@@ -512,12 +512,12 @@ def has_object_columns(
     """
     Indicate whether on a Series of dtypes for a Pandas DataFrame
     there exists one or more object columns.
+            
+    Args:
+        feature_types (pd.Series): The feature types for a DataFrame.
 
-    Arguments:
-        feature_types (pd.Series):
-            The feature types for a DataFrame.
     Returns:
-        bool:
+        bool: 
             True if the DataFrame dtypes contain an object column, False
             otherwise.
     """
--- a/autoPyTorch/data/base_feature_validator.py
+++ b/autoPyTorch/data/base_feature_validator.py
@@ -79,7 +79,7 @@ class BaseFeatureValidator(BaseEstimator):
         The supported data types are List, numpy arrays and pandas DataFrames.
         CSR sparse data types are also supported
 
-        Arguments:
+        Args:
             X_train (SUPPORTED_FEAT_TYPES):
                 A set of features that are going to be validated (type and dimensionality
                 checks) and a encoder fitted in the case the data needs encoding
@@ -115,7 +115,7 @@ class BaseFeatureValidator(BaseEstimator):
         X: SUPPORTED_FEAT_TYPES,
     ) -> BaseEstimator:
         """
-        Arguments:
+        Args:
             X (SUPPORTED_FEAT_TYPES):
                 A set of features that are going to be validated (type and dimensionality
                 checks) and a encoder fitted in the case the data needs encoding
@@ -133,7 +133,7 @@ class BaseFeatureValidator(BaseEstimator):
         """
         Feature dimensionality and data type checks
 
-        Arguments:
+        Args:
             X (SUPPORTED_FEAT_TYPES):
                 A set of features that are going to be validated (type and dimensionality
                 checks) and a encoder fitted in the case the data needs encoding
@@ -146,7 +146,7 @@ class BaseFeatureValidator(BaseEstimator):
         X: SUPPORTED_FEAT_TYPES,
     ) -> np.ndarray:
         """
-        Arguments:
+        Args:
             X_train (SUPPORTED_FEAT_TYPES):
                 A set of features, whose categorical features are going to be
                 transformed
@@ -168,7 +168,7 @@ class BaseFeatureValidator(BaseEstimator):
 
         If test data is provided, we proactively match it to train data
 
-        Arguments:
+        Args:
             X_train (SUPPORTED_FEAT_TYPES):
                 A set of features that are going to be validated (type and dimensionality
                 checks) and a encoder fitted in the case the data needs encoding
--- a/autoPyTorch/data/base_target_validator.py
+++ b/autoPyTorch/data/base_target_validator.py
@@ -75,7 +75,7 @@ class BaseTargetValidator(BaseEstimator):
         Validates and fit a categorical encoder (if needed) to the targets
         The supported data types are List, numpy arrays and pandas DataFrames.
 
-        Arguments:
+        Args:
             y_train (SUPPORTED_TARGET_TYPES)
                 A set of targets set aside for training
             y_test (Union[SUPPORTED_TARGET_TYPES])
@@ -132,7 +132,7 @@ class BaseTargetValidator(BaseEstimator):
         y_test: Optional[SUPPORTED_TARGET_TYPES] = None,
     ) -> BaseEstimator:
         """
-        Arguments:
+        Args:
             y_train (SUPPORTED_TARGET_TYPES)
                 The labels of the current task. They are going to be encoded in case
                 of classification
@@ -146,7 +146,7 @@ class BaseTargetValidator(BaseEstimator):
         y: Union[SUPPORTED_TARGET_TYPES],
     ) -> np.ndarray:
         """
-        Arguments:
+        Args:
             y (SUPPORTED_TARGET_TYPES)
                 A set of targets that are going to be encoded if the current task
                 is classification
@@ -163,7 +163,7 @@ class BaseTargetValidator(BaseEstimator):
         """
         Revert any encoding transformation done on a target array
 
-        Arguments:
+        Args:
             y (Union[np.ndarray, pd.DataFrame, pd.Series]):
                 Target array to be transformed back to original form before encoding
         Returns:
--- a/autoPyTorch/data/base_validator.py
+++ b/autoPyTorch/data/base_validator.py
@@ -58,7 +58,7 @@ class BaseInputValidator(BaseEstimator):
             + Checks for dimensionality as well as missing values are performed.
             + If performing a classification task, the data is going to be encoded
 
-        Arguments:
+        Args:
             X_train (SUPPORTED_FEAT_TYPES):
                 A set of features that are going to be validated (type and dimensionality
                 checks). If this data contains categorical columns, an encoder is going to
@@ -102,7 +102,7 @@ class BaseInputValidator(BaseEstimator):
         """
         Transform the given target or features to a numpy array
 
-        Arguments:
+        Args:
             X (SUPPORTED_FEAT_TYPES):
                 A set of features to transform
             y (typing.Optional[SUPPORTED_TARGET_TYPES]):
--- a/autoPyTorch/data/tabular_target_validator.py
+++ b/autoPyTorch/data/tabular_target_validator.py
@@ -28,7 +28,7 @@ class TabularTargetValidator(BaseTargetValidator):
         It does so by also using the classes from the test data, to prevent encoding
         errors
 
-        Arguments:
+        Args:
             y_train (SUPPORTED_TARGET_TYPES)
                 The labels of the current task. They are going to be encoded in case
                 of classification
@@ -100,7 +100,7 @@ class TabularTargetValidator(BaseTargetValidator):
         Validates and fit a categorical encoder (if needed) to the features.
         The supported data types are List, numpy arrays and pandas DataFrames.
 
-        Arguments:
+        Args:
             y (SUPPORTED_TARGET_TYPES)
                 A set of targets that are going to be encoded if the current task
                 is classification
@@ -152,7 +152,7 @@ class TabularTargetValidator(BaseTargetValidator):
         """
         Revert any encoding transformation done on a target array
 
-        Arguments:
+        Args:
             y (typing.Union[np.ndarray, pd.DataFrame, pd.Series]):
                 Target array to be transformed back to original form before encoding
         Returns:
@@ -189,7 +189,7 @@ class TabularTargetValidator(BaseTargetValidator):
         """
         Perform dimensionality and data type checks on the targets
 
-        Arguments:
+        Args:
             y (typing.Union[np.ndarray, pd.DataFrame, pd.Series]):
                 A set of features whose dimensionality and data type is going to be checked
         """


autoPyTorch/pipeline/components/training/trainer/RowCutOutTrainer.py
```diff
@@ -37,7 +37,7 @@ def data_preparation(self, X: np.ndarray, y: np.ndarray,
             return X, {'y_a': y_a, 'y_b': y_b, 'lam': lam}
 
         size = X.shape[1]
-        indices = self.random_state.choice(range(1, size), max(1, np.int32(size * self.patch_ratio)),
+        indices = self.random_state.choice(range(size), max(1, np.int32(size * self.patch_ratio)),
```
nabenabe0928, line 40: What happens when we switch `np.int32` to `int`?
I usually get unexpected errors when mixing up the usage of `torch` and `np.intxx`, so I would like to avoid using `np.intxx` for a single variable as much as possible.

autoPyTorch/pipeline/components/training/trainer/AdversarialTrainer.py
```diff
@@ -189,12 +189,17 @@ def get_hyperparameter_search_space(
...
-            default_value=0.2),
+            value_range=(0.001, 0.15),
+            default_value=0.007,
+            log=True),
     ) -> ConfigurationSpace:
         cs = ConfigurationSpace()
 
+        epsilon = HyperparameterSearchSpace(hyperparameter="epsilon",
+                                            value_range=(0.007, 0.007),
+                                            default_value=0.007)
```
nabenabe0928, line 200: Constant value? but then why do we pass epsilon in the argument as well??

autoPyTorch/pipeline/components/training/data_loader/base_data_loader.py
```diff
@@ -112,10 +112,10 @@ def fit(self, X: Dict[str, Any], y: Any = None) -> torch.utils.data.DataLoader:
         self.train_data_loader = torch.utils.data.DataLoader(
             train_dataset,
             batch_size=min(self.batch_size, len(train_dataset)),
-            shuffle=True,
+            shuffle=False,
             num_workers=X.get('num_workers', 0),
             pin_memory=X.get('pin_memory', True),
-            drop_last=X.get('drop_last', True),
+            drop_last=X.get('drop_last', False),
```
nabenabe0928, line 118: Could you explain the intentions behind those changes?

autoPyTorch/pipeline/components/setup/network_embedding/base_network_embedding.py
```diff
@@ -33,19 +33,20 @@ def build_embedding(self, num_input_features: np.ndarray, num_numerical_features
 
     def _get_args(self, X: Dict[str, Any]) -> Tuple[int, np.ndarray]:
```
nabenabe0928, line 34: Should we delete this method at all?

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/TabularColumnTransformer.py
```diff
@@ -47,15 +47,16 @@ def fit(self, X: Dict[str, Any], y: Any = None) -> "TabularColumnTransformer":
...
-        preprocessors = get_tabular_preprocessers(X)
-        if len(X['dataset_properties']['numerical_columns']):
-            numerical_pipeline = make_pipeline(*preprocessors['numerical'])
-        if len(X['dataset_properties']['categorical_columns']):
-            categorical_pipeline = make_pipeline(*preprocessors['categorical'])
+        # preprocessors = get_tabular_preprocessers(X)
+        # if len(X['dataset_properties']['numerical_columns']):
+        #     numerical_pipeline = make_pipeline(*preprocessors['numerical'])
+        # if len(X['dataset_properties']['categorical_columns']):
+        #     categorical_pipeline = make_pipeline(*preprocessors['categorical'])
```
nabenabe0928, line 59: Could you explain the reasons of the changes?

nabenabe0928 (commented):
Thank you for the PR, I added some comments.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -9,14 +9,63 @@
...
 from sklearn.exceptions import NotFittedError
+from sklearn.pipeline import make_pipeline
+from sklearn.preprocessing import OneHotEncoder
+from sklearn.impute import SimpleImputer
+from sklearn.preprocessing import StandardScaler, OrdinalEncoder
 
 from autoPyTorch.data.base_feature_validator import BaseFeatureValidator, SUPPORTED_FEAT_TYPES
 
 
+def _create_column_transformer(preprocessors: typing.Dict, numerical_columns, categorical_columns):
```
nabenabe0928, line 24: typing
I need the intention behind this function to review.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -9,14 +9,63 @@
...
+        categorical_pipeline = make_pipeline(*preprocessors['categorical'])
+
+    return ColumnTransformer([
+        ('categorical_pipeline', categorical_pipeline, categorical_columns),
+        ('numerical_pipeline', numerical_pipeline, numerical_columns)],
+        remainder='passthrough'
+    )
+
+
+def get_tabular_preprocessors():
```
nabenabe0928, line 39: typing
Here as well, I need to know how this function helps.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -43,73 +92,53 @@ def _fit(
 
         if hasattr(X, "iloc") and not scipy.sparse.issparse(X):
             X = typing.cast(pd.DataFrame, X)
-            # Treat a column with all instances a NaN as numerical
```
nabenabe0928, line 46: I have not checked L46 - L112 (in the old version).

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -9,14 +9,63 @@
...
+                  # Train data is numpy
+                  # as of this point, where
+                  # Ordinal Encoding is using
+                  # for categorical. Only
+                  # Numbers are allowed
+                  # fill_value='!missing!',
+                  fill_value=-1,
+                  copy=False))
+
+    preprocessors['categorical'].append(OrdinalEncoder(
```
ArlindKadra, line 58: The ordering seems weird, is it reversed?
OneHotEncoder should only be called in the end.
Additionally, why do we need the ordinal encoder? Does one hot encoder not handle the case of let us say 4 unique values str and -1?  The ordering should be something like Simple imputer, Onehot encoder

autoPyTorch/data/tabular_feature_validator.py
ArlindKadra: As mentioned, it would be good to include the documentation here too.

ArlindKadra (commented):
To be continued.

autoPyTorch/pipeline/components/training/data_loader/base_data_loader.py
ArlindKadra: `shuffle` from `True` to `False` was an ablation since on the previous version of the cocktails it was turned to False by default. The optimal behavior should be true, you do want the data to be shuffled on each epoch. The `drop_last` was also `True` by default, it should be `False`, we do not want to be dropping data especially when dealing with small datasets and a large batch size.

So all in all, the change to `shuffle` will be reverted and the change to` drop_last` should stay.

autoPyTorch/pipeline/components/training/trainer/RowCutOutTrainer.py
ArlindKadra: That is a good question

autoPyTorch/pipeline/components/training/trainer/AdversarialTrainer.py
ArlindKadra: An argument to replicate the results of the regularization cocktail paper, will not stay when merged to development.

autoPyTorch/pipeline/components/setup/network_embedding/base_network_embedding.py
ArlindKadra: Hmm, I will get back on this one, after I understand the rest of the code additions and changes if this is not necessary anymore. If not needed, it can be removed. 
Additionally, this function should have had some documentation and comments before, it is hard to read.

autoPyTorch/api/base_task.py
```diff
@@ -920,6 +941,8 @@ def _search(
                 self._stopwatch.stop_task(traditional_task_name)
 
         # ============> Starting ensemble
+        self.precision = precision
+        self.opt_metric = optimize_metric
```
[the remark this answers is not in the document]
ravinkohli: they are being used when fitting an ensemble post hoc. We require this to allow the user to reuse the values passed for search in fit_ensemble in case both are None.

autoPyTorch/api/base_task.py
```diff
@@ -1333,6 +1336,141 @@ def fit_pipeline(self,
 
         return fitted_pipeline, run_info, run_value, dataset
 
+    def fit_ensemble(
+            self,
+            ensemble_nbest: int = 50,
+            ensemble_size: int = 50,
+            precision: int = 32,
```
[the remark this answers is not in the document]
ravinkohli: I have added checks to ensure they are in this range. @nabenabe0928 Sure, please make another pull request to add the enumerator. Sounds like a good idea

autoPyTorch/api/base_task.py
```diff
@@ -1333,6 +1336,141 @@ def fit_pipeline(self,
...
+        future = manager.futures.pop()
+        result = future.result()
+        if result is None:
+            raise ValueError("Errors occurred while building the ensemble - please"
+                             " check the log file and command line output for error messages.")
+        self.ensemble_performance_history, _, _, _ = result
+
+        if load_models:
+            self._load_models()
+        if self._logger is not None:
```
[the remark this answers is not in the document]
ravinkohli: yes I'll do so

autoPyTorch/utils/backend.py
```diff
@@ -205,12 +205,12 @@ def temporary_directory(self) -> str:
...
     def _make_internals_directory(self) -> None:
         try:
-            os.makedirs(self.internals_directory)
+            os.makedirs(self.internals_directory, exist_ok=True)
         except Exception as e:
             if self._logger is not None:
                 self._logger.debug("_make_internals_directory: %s" % e)
         try:
-            os.makedirs(self.get_runs_directory())
+            os.makedirs(self.get_runs_directory(), exist_ok=True)
```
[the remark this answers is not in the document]
ravinkohli: so, the main idea is to run post hoc ensemble fitting. In case the user wants to fit an ensemble using the results from a previous search, they can simply pass the old directory and we can fit an ensemble over that. @nabenabe0928 the automl_common is being used in the `development` branch. Here we are still using the old code.

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
```diff
@@ -408,8 +407,8 @@ def _fit(self, X: Dict[str, Any], y: Any = None, **kwargs: Any) -> 'TrainerChoic
             # change model
             update_model_state_dict_from_swa(X['network'], self.choice.swa_model.state_dict())
             if self.choice.use_snapshot_ensemble:
-                for model in self.choice.model_snapshots:
-                    swa_utils.update_bn(X['train_data_loader'], model.double())
+                # we update only the last network which pertains to the stochastic weight averaging model
+                swa_utils.update_bn(X['train_data_loader'], self.choice.model_snapshots[-1].double())
```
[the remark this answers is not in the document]
ravinkohli: Yeah, I agree

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -318,8 +320,14 @@ def on_epoch_end(self, X: Dict[str, Any], epoch: int) -> bool:
...
                                                              "none when snapshot ensembling is enabled"
-                    model_copy = deepcopy(self.swa_model) if self.use_stochastic_weight_averaging \
-                        else deepcopy(self.model)
+                    if epoch == self.budget_tracker.max_epochs:
+                        if self.use_stochastic_weight_averaging:
+                            model_copy = deepcopy(self.swa_model)
+                        else:
+                            model_copy = deepcopy(self.model)
+                    else:
+                        model_copy = deepcopy(self.model)
```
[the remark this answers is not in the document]
ravinkohli: Could you explain how this could be beneficial?

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -221,6 +221,8 @@ def __init__(self, weighted_loss: int = 0,
         self.add_fit_requirements([
             FitRequirement("is_cyclic_scheduler", (bool,), user_defined=False, dataset_property=False),
         ])
+        self.batch_fit_times = []
+        self.data_loading_times = []
```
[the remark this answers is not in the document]
ravinkohli: they were temporary debug statements which are now gone

autoPyTorch/pipeline/components/training/trainer/RowCutOutTrainer.py
ravinkohli: okay I'll change this to np.int.

autoPyTorch/pipeline/components/training/trainer/AdversarialTrainer.py
ravinkohli: should I remove it now?

autoPyTorch/pipeline/components/setup/network_embedding/base_network_embedding.py
ravinkohli: Okay, so to resolve this we would have to decide on how to handle the preprocessing pipeline. To save time, we have agreed to run it only once per dataset and without hyperparameters(this includes imputing, scaling, and encoding and most importantly not feature preprocessing which still has a choice like (PCA, etc.)). However, we may want to also have hyperparameters for the preprocessing pipeline but transform the data once before training so instead of preprocessing once per dataset, we'll do it once per configuration. We need to decide on this tradeoff. Once this is resolved, I'll correct this function.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/TabularColumnTransformer.py
ravinkohli: please refer to [this](https://github.com/automl/Auto-PyTorch/pull/286#discussion_r719595620) comment

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -177,10 +189,9 @@ def transform(
         self._check_data(X)
         # We also need to fillna on the transformation
         # in case test data is provided
-        X = self.impute_nan_in_categories(X)
-
-        if self.encoder is not None:
-            X = self.encoder.transform(X)
+        if len(self.categorical_columns) >= 0:
```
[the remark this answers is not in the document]
ravinkohli: for numerical columns, we have the simple imputer. For categorical columns, we only encode and therefore need to handle this before.

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: yeah, this import typing was there from when fransisco coded it. I was too lazy to change these. I'll fix it in this file

requirements.txt
```diff
@@ -1,10 +1,10 @@
 pandas
-torch
-torchvision
+torch<=1.8
+torchvision<=0.9
 tensorboard
 scikit-learn>=0.24.0,<0.25.0
 numpy
-scipy
+scipy==1.6.3
```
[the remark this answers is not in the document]
ravinkohli: I actually don't remember why we did this. Could you remind me? This is also causing issues when trying to run the github workflows.

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: I have fixed this now.

autoPyTorch/pipeline/components/training/data_loader/base_data_loader.py
ravinkohli: done

autoPyTorch/pipeline/components/setup/network_embedding/base_network_embedding.py
ArlindKadra: The behavior for the future will be once per configuration.

autoPyTorch/pipeline/components/training/trainer/AdversarialTrainer.py
ArlindKadra: Not at the moment.

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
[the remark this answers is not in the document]
ArlindKadra: The benefit would be the code cleanup. It does look nicer to me, the only thing I would suggest would be changing the order of`self.use_stochastic_weight_averaging` with `is_last_epoch`

autoPyTorch/utils/backend.py
```diff
@@ -328,6 +328,11 @@ def load_datamanager(self) -> BaseDataset:
             with open(filepath, 'rb') as fh:
                 return pickle.load(fh)
 
+    def replace_datamanager(self, datamanager: BaseDataset) -> None:
```
ArlindKadra, line 331: @ravinkohli There probably should be a condition to call replace manager only when search is done.

autoPyTorch/datasets/base_dataset.py
```diff
@@ -330,13 +330,19 @@ def get_dataset_for_training(self, split_id: int, train: bool, subset: int = 0)
         to provide training data to fit a pipeline
 
         Args:
-            split (int): The desired subset of the dataset to split and use
+            split_id (int): which split id to get from the splits
+            train (bool): whether the train or valid transforms are to be applied
+            subset (int, default=0): 0 is for train_indices, 1 is for valid_indices
```
nabenabe0928, line 335: To me, `train != subset` looks always true.
I mean we do not need to take `subset`.
Or even if we need it, we should rename it such as `which_subset: Literal['train', 'val']`.

autoPyTorch/pipeline/base_pipeline.py
```diff
@@ -451,12 +451,14 @@ def _check_search_space_updates(self, include: Optional[Dict[str, Any]],
...
                         raise ValueError("Unknown hyperparameter for component {}. "
                                          "Expected update hyperparameter "
-                                         "to be in {} got {}".format(node.__class__.__name__,
-                                                                     component.
-                                                                     get_hyperparameter_search_space(
-                                                                         dataset_properties=self.dataset_properties).
-                                                                     get_hyperparameter_names(),
-                                                                     split_hyperparameter[1]))
+                                         "to be in {} got {}."
+                                         " component is {}".format(node.__class__.__name__,
```
nabenabe0928, line 455: What is the difference between `node.__class__.__name__` and `component.__name__`?
Both say `component`, so it would be nicer if users can know the difference from the error message.

autoPyTorch/pipeline/components/training/trainer/RowCutOutTrainer.py
nabenabe0928: Not `np.int`, but `int`!
It is a shame that I cannot provide the example of the error, but I once got an error regarding `np.int32(x)` vs `int(x)`.
And because torch says `expected int, but got int` although it was `np.int32`, it was quite hard time for me to find out why this error comes up again and again.

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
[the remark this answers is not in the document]
nabenabe0928: Yes, as Arlind mentioned, it reduces the nest of the code block.
Based on his suggestion, this will be the new suggestion.
```suggestion
                    is_last_epoch = (epoch == self.budget_tracker.max_epochs)
                    if is_last_epoch and self.use_stochastic_weight_averaging:
                        model_copy = deepcopy(self.swa_model)
                    else:
                        model_copy = deepcopy(self.model)
```

autoPyTorch/pipeline/components/training/trainer/RowCutMixTrainer.py
```diff
@@ -36,7 +36,7 @@ def data_preparation(self, X: np.ndarray, y: np.ndarray,
             return X, {'y_a': y, 'y_b': y[index], 'lam': 1}
 
         size = X.shape[1]
-        indices = torch.tensor(self.random_state.choice(range(1, size), max(1, np.int32(size * lam)),
+        indices = torch.tensor(self.random_state.choice(range(size), max(1, np.int32(size * lam)),
```
nabenabe0928, line 39:
```suggestion
        indices = torch.tensor(self.random_state.choice(range(size), max(1, int(size * lam)),
```

autoPyTorch/pipeline/components/setup/network_embedding/base_network_embedding.py
```diff
@@ -30,22 +30,23 @@ def transform(self, X: Dict[str, Any]) -> Dict[str, Any]:
...
+    #     #     num_numerical_columns = numerical_column_transformer.transform(
+    #     #         X_train[:, X['dataset_properties']['numerical_columns']]).shape[1]
+    #     # num_input_features = np.zeros((num_numerical_columns + len(X['dataset_properties']['categorical_columns'])),
+    #     #                               dtype=int)
+    #     # categories = X['dataset_properties']['categories']
+    #     #
+    #     # for i, category in enumerate(categories):
+    #     #     num_input_features[num_numerical_columns + i, ] = len(category)
+    #     # return num_numerical_columns, num_input_features
+    #     return None, None
```
nabenabe0928, line 52: Can we remove those lines?

nabenabe0928 (commented):
I checked some of the changes and will post the rest later.

autoPyTorch/pipeline/components/training/trainer/RowCutOutTrainer.py
ravinkohli: hmm, weird. Maybe we can look out for the error? I have never had a problem with this and we have not faced any issues so far also related to this so maybe there was some other contributing factor to the error you got.

autoPyTorch/datasets/base_dataset.py
ravinkohli: I agree with your comment. I'll change it to just have train as the argument and use not train for the subset. I don't think we need a literal as it is a bit redundant to pass both train=True and subset=train

autoPyTorch/pipeline/base_pipeline.py
ravinkohli: yeah I'll fix this comment thanks for pointing out.

autoPyTorch/utils/backend.py
ravinkohli: yeah I'm now handling this better.

autoPyTorch/pipeline/components/setup/network_embedding/base_network_embedding.py
ravinkohli: okay so once the experiments are done, I'll adapt this to the implementation for the preprocessor.

autoPyTorch/datasets/base_dataset.py
```diff
@@ -330,13 +330,19 @@ def get_dataset_for_training(self, split_id: int, train: bool, subset: int = 0)
...
 
         Returns:
+
             Dataset: the reduced dataset to be used for testing
         """
         # Subset creates a dataset. Splits is a (train_indices, test_indices) tuple
-        return TransformSubset(self, self.splits[split_id][subset], train=train)
+        assert split_id <= len(self.splits), "Expected split id to be less than length of splits"
```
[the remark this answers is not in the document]
ravinkohli: oh yeah true, thanks for the catch

autoPyTorch/pipeline/components/training/trainer/RowCutMixTrainer.py
```diff
@@ -28,29 +28,28 @@ def data_preparation(self, X: np.ndarray, y: np.ndarray,
...
-        size = X.shape[1]
-        indices = torch.tensor(self.random_state.choice(range(size), max(1, np.int32(size * lam)),
+        indices = torch.tensor(self.random_state.choice(range(batch_size), max(1, np.int32(n_columns * lam)),
                                                         replace=False))
 
         X[:, indices] = X[index, :][:, indices]
 
         # Adjust lam
-        lam = 1 - ((len(indices)) / (X.size()[1]))
+        lam = 1 - ((len(indices)) / (n_columns))
```
nabenabe0928, line 44:
```
lam = 1 - (len(indices) / n_columns)
```

autoPyTorch/pipeline/tabular_classification.py
```diff
@@ -102,7 +102,7 @@ def _predict_proba(self, X: np.ndarray) -> np.ndarray:
                 proba_k /= normalizer
                 all_proba.append(proba_k)
 
-            return all_proba
+            return np.ndarray(all_proba)
```
nabenabe0928, line 105: `np.array`?

autoPyTorch/pipeline/components/training/trainer/RowCutOutTrainer.py
nabenabe0928: Yeah, it was more or less a torch issue, so I will report when I find the same issue in the future.
But, just in case, are there any reasons to use `np.int32(x)`?
```
import numpy as np


a, b = np.int32(3), int(3)
type(a) == type(b)

>>> False
```

autoPyTorch/pipeline/components/training/trainer/RowCutMixTrainer.py
ravinkohli:
```suggestion
        lam = 1 - ((len(indices)) / n_columns)
```

autoPyTorch/datasets/base_dataset.py
```diff
@@ -322,21 +322,30 @@ def create_holdout_val_split(
...
+
             Dataset: the reduced dataset to be used for testing
         """
         # Subset creates a dataset. Splits is a (train_indices, test_indices) tuple
-        return TransformSubset(self, self.splits[split_id][subset], train=train)
+        if split_id >= len(self.splits):  # old version: split_id > len(self.splits)
+            raise IndexError("split_id out of range, got split_id={}"
+                             " (>= num_splits={})".format(split_id, len(self.splits)))
+        subset = int(not train)
+        indices = self.splits[split_id][subset]
```
nabenabe0928, line 345:
```suggestion
        # which subset (either train splits, i.e. 0, or validation split, i.e. 1) to use
        which_subset = int(not train)
        indices = self.splits[split_id][which_subset]
```

autoPyTorch/pipeline/base_pipeline.py
```diff
@@ -425,16 +427,16 @@ def _check_search_space_updates(self, include: Optional[Dict[str, Any]],
                             if choice in exclude[update.node_name]:
                                 raise ValueError("Found {} in exclude".format(choice))
                         if choice not in components.keys():
-                            raise ValueError("Unknown hyperparameter for choice {}. "
+                            raise ValueError("Unknown component choice for node {}. "
                                              "Expected update hyperparameter "
                                              "to be in {} got {}".format(node.__class__.__name__,
```
nabenabe0928, line 432:
```suggestion
                                             "to be in {}, but got {}".format(node.__class__.__name__,
```

autoPyTorch/pipeline/base_pipeline.py
```diff
@@ -425,16 +427,16 @@ def _check_search_space_updates(self, include: Optional[Dict[str, Any]],
...
                                                                          components.keys(), choice))
                 # check if the component whose hyperparameter
                 # needs to be updated is in components of the
                 # choice module
                 elif split_hyperparameter[0] not in components.keys():
-                    raise ValueError("Unknown hyperparameter for choice {}. "
-                                     "Expected update hyperparameter "
+                    raise ValueError("Unknown component choice for node {}. "
+                                     "Expected update component "
                                      "to be in {} got {}".format(node.__class__.__name__,
```
nabenabe0928, line 440:
```suggestion
                                     "to be in {}, but got {}".format(node.__class__.__name__,
```

autoPyTorch/pipeline/base_pipeline.py
```diff
@@ -449,22 +451,24 @@ def _check_search_space_updates(self, include: Optional[Dict[str, Any]],
...
-                                         "Expected update hyperparameter "
-                                         "to be in {} got {}".format(node.__class__.__name__,
-                                                                     component.
-                                                                     get_hyperparameter_search_space(
-                                                                         dataset_properties=self.dataset_properties).
-                                                                     get_hyperparameter_names(),
-                                                                     split_hyperparameter[1]))
+                        raise ValueError("Unknown hyperparameter for  component {} of node {}."
+                                         " Expected update hyperparameter "
+                                         "to be in {} got {}.".format(component.__name__,
```
nabenabe0928, line 456:
```suggestion
                                         "to be in {}, but got {}.".format(component.__name__,
```

autoPyTorch/pipeline/base_pipeline.py
```diff
@@ -449,22 +451,24 @@ def _check_search_space_updates(self, include: Optional[Dict[str, Any]],
...
                 if update.hyperparameter not in node.get_hyperparameter_search_space(
                         dataset_properties=self.dataset_properties):
                     if any([update.hyperparameter in name for name in
                             node.get_hyperparameter_search_space(
                                 dataset_properties=self.dataset_properties).get_hyperparameter_names()]):
                         continue
-                    raise ValueError("Unknown hyperparameter for component {}. "
+                    raise ValueError("Unknown hyperparameter for node {}. "
                                      "Expected update hyperparameter "
                                      "to be in {} got {}".format(node.__class__.__name__,
```
nabenabe0928, line 473:
```suggestion
                                     "to be in {}, but got {}".format(node.__class__.__name__,
```

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
```diff
@@ -92,15 +92,35 @@ def backward(ctx: typing.Any,
...
+        alpha = torch.rand(1)
+
+    if method == 'shake-shake':
+        beta = torch.rand(1)
+    elif method in ['shake-even', 'even-even']:
+        beta = torch.FloatTensor([0.5])
+    elif method == 'M3':
+        beta = torch.FloatTensor(
+            [torch.rand(1) * (0.5 - alpha) * alpha if alpha < 0.5 else torch.rand(1) * (alpha - 0.5) * alpha]
+        )
```
nabenabe0928, line 121: If I understand the paper correctly, the M3 is not correct.
```suggestion
        # Table 4 in the paper `Shake-Shake regularization`
        rnd = torch.rand(1)
        beta = torch.FloatTensor(
            [rnd * (0.5 - alpha) + alpha if alpha < 0.5 else rnd * (alpha - 0.5) + 0.5]
        )
```

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
```diff
@@ -92,15 +92,35 @@ def backward(ctx: typing.Any,
...
+    https://arxiv.org/abs/1705.07485. The names have been taken from the paper as well.
+    """
     if not is_training:
         result = (torch.FloatTensor([0.5]), torch.FloatTensor([0.5]))
         return result if not is_cuda else (result[0].cuda(), result[1].cuda())
 
     # TODO implement other update methods
-    alpha = torch.rand(1)
-    beta = torch.rand(1)
+    if method == 'even-even':
```
nabenabe0928, line 109: Wait. I checked [the paper](https://arxiv.org/pdf/1705.07485.pdf), but `shake-even` in the paper is equivalent to `even-even` here and there is nothing about `shake-even` in the paper.

ravinkohli: shake-even means sampling alpha and keeping beta as 0.5. 
even even is keeping both alpha and beta at 0.5

ArlindKadra: `shake-even` means that you shake `alpha` and you keep `beta` as 0.5.
From the paper:
```
"Shake" means that all scaling coefficients are overwritten
with new random numbers before the pass. "Even" means that all scaling coefficients are set to
0.5 before the pass.
```

ArlindKadra: so if the method is not `even-even` it will shake alpha and beta will be 0.5, if my understanding is correct. `even-even` should be 0.5 both.

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
```diff
@@ -92,15 +92,37 @@ def backward(ctx: typing.Any,
...
+) -> typing.Tuple[torch.tensor, torch.tensor]:
+    """
+    The methods used in this function have been introduced in 'ShakeShake Regularisation'
+    https://arxiv.org/abs/1705.07485. The names have been taken from the paper as well.
+    """
     if not is_training:
         result = (torch.FloatTensor([0.5]), torch.FloatTensor([0.5]))
         return result if not is_cuda else (result[0].cuda(), result[1].cuda())
 
     # TODO implement other update methods
```
nabenabe0928, line 108:
```suggestion
    # TODO implement other update methods
    # alpha is the weight ratio for the forward pass and beta is that for the backward pass
```

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
```diff
@@ -92,15 +92,37 @@ def backward(ctx: typing.Any,
...
-def shake_get_alpha_beta(is_training: bool, is_cuda: bool
-                         ) -> typing.Tuple[torch.tensor, torch.tensor]:
+def shake_get_alpha_beta(
+        is_training: bool,
+        is_cuda: bool,
+        method: str
+) -> typing.Tuple[torch.tensor, torch.tensor]:
+    """
+    The methods used in this function have been introduced in 'ShakeShake Regularisation'
+    https://arxiv.org/abs/1705.07485. The names have been taken from the paper as well.
```
nabenabe0928, line 102:
```suggestion
    https://arxiv.org/abs/1705.07485. The names have been taken from the paper as well.
    Currently, this function supports `even-even`, `shake-even` and `shake-shake`
```

autoPyTorch/pipeline/components/training/trainer/RowCutOutTrainer.py
```diff
@@ -36,35 +32,19 @@ def data_preparation(self, X: np.ndarray, y: np.ndarray,
...
-    
-        # We use an ordinal encoder on the categorical columns of tabular data
-        # -1 is the conceptual equivalent to 0 in a image, that does not
-        # have color as a feature and hence the network has to learn to deal
-        # without this data. For numerical columns we use 0 to cutout the features
-        # similar to the effect that setting 0 as a pixel value in an image.
-        X[:, categorical_indices.long()] = self.CATEGORICAL_VALUE
-        X[:, numerical_indices.long()] = self.NUMERICAL_VALUE
-        """
         X[:, indices] = 0
```
nabenabe0928, line 39: Is it guaranteed that categorical symbols do not have 0 as a choice?

autoPyTorch/pipeline/components/training/trainer/RowCutMixTrainer.py
```diff
@@ -28,29 +28,28 @@ def data_preparation(self, X: np.ndarray, y: np.ndarray,
         """
         beta = 1.0
         lam = self.random_state.beta(beta, beta)
-        batch_size = X.size()[0]
+        batch_size, n_columns = np.shape(X)
         index = torch.randperm(batch_size).cuda() if X.is_cuda else torch.randperm(batch_size)
```
nabenabe0928, line 32:
```suggestion
        # permed_indices: Shuffled version of torch.arange(batch_size) 
        permed_indices = torch.randperm(batch_size).cuda() if X.is_cuda else torch.randperm(batch_size)
```

autoPyTorch/pipeline/components/training/trainer/RowCutMixTrainer.py
```diff
@@ -28,29 +28,28 @@ def data_preparation(self, X: np.ndarray, y: np.ndarray,
...
 
         r = self.random_state.rand(1)
         if beta <= 0 or r > self.alpha:
             return X, {'y_a': y, 'y_b': y[index], 'lam': 1}
 
-        size = X.shape[1]
-        indices = torch.tensor(self.random_state.choice(range(1, size), max(1, np.int32(size * lam)),
+        indices = torch.tensor(self.random_state.choice(range(n_columns), max(1, np.int32(n_columns * lam)),
```
nabenabe0928, line 38:
```suggestion
        cut_indices = torch.tensor(self.random_state.choice(range(n_columns), max(1, np.int32(n_columns * lam)),
```

autoPyTorch/pipeline/components/training/trainer/RowCutMixTrainer.py
```diff
@@ -28,29 +28,28 @@ def data_preparation(self, X: np.ndarray, y: np.ndarray,
...
                                                         replace=False))
 
         X[:, indices] = X[index, :][:, indices]
```
nabenabe0928, line 41:
```suggestion
        # Replace the values in `cut_indices` columns with
        # the values from `permed_indices`
        X[:, cut_indices] = X[permed_indices, :][:, cut_indices]
```

autoPyTorch/pipeline/components/training/trainer/RowCutMixTrainer.py
```diff
@@ -28,29 +28,28 @@ def data_preparation(self, X: np.ndarray, y: np.ndarray,
         """
         beta = 1.0
         lam = self.random_state.beta(beta, beta)
-        batch_size = X.size()[0]
+        batch_size, n_columns = np.shape(X)
         index = torch.randperm(batch_size).cuda() if X.is_cuda else torch.randperm(batch_size)
 
         r = self.random_state.rand(1)
         if beta <= 0 or r > self.alpha:
             return X, {'y_a': y, 'y_b': y[index], 'lam': 1}
```
nabenabe0928, line 36:
```suggestion
            return X, {'y_a': y, 'y_b': y[permed_indices], 'lam': 1}
```

autoPyTorch/pipeline/components/training/trainer/RowCutMixTrainer.py
```diff
@@ -28,29 +28,28 @@ def data_preparation(self, X: np.ndarray, y: np.ndarray,
...
-        size = X.shape[1]
-        indices = torch.tensor(self.random_state.choice(range(1, size), max(1, np.int32(size * lam)),
+        indices = torch.tensor(self.random_state.choice(range(n_columns), max(1, np.int32(n_columns * lam)),
                                                         replace=False))
 
         X[:, indices] = X[index, :][:, indices]
 
         # Adjust lam
-        lam = 1 - ((len(indices)) / (X.size()[1]))
+        lam = 1 - (len(indices) / n_columns)
```
nabenabe0928, line 44:
```suggestion
        lam = 1 - (len(cut_indices) / n_columns)
```

autoPyTorch/pipeline/components/training/trainer/RowCutMixTrainer.py
```diff
@@ -28,29 +28,28 @@ def data_preparation(self, X: np.ndarray, y: np.ndarray,
...
 
         y_a, y_b = y, y[index]
```
nabenabe0928, line 46:
```suggestion
        y_a, y_b = y, y[permed_indices]
```

autoPyTorch/pipeline/components/training/trainer/RowCutMixTrainer.py
```diff
@@ -28,29 +28,28 @@ def data_preparation(self, X: np.ndarray, y: np.ndarray,
...
             return X, {'y_a': y, 'y_b': y[index], 'lam': 1}
 
-        size = X.shape[1]
-        indices = torch.tensor(self.random_state.choice(range(1, size), max(1, np.int32(size * lam)),
+        indices = torch.tensor(self.random_state.choice(range(n_columns), max(1, np.int32(n_columns * lam)),
                                                         replace=False))
 
         X[:, indices] = X[index, :][:, indices]
 
         # Adjust lam
```
nabenabe0928, line 43:
```suggestion
        # Since we cannot cut exactly `lam x 100 %` of rows, we need to adjust the `lam`
```

autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
```diff
@@ -327,11 +338,14 @@ def forward(self, x: torch.FloatTensor) -> torch.FloatTensor:
             if self.config["multi_branch_choice"] == 'shake-shake':
                 x1 = self.layers(x)
                 x2 = self.shake_shake_layers(x)
-                alpha, beta = shake_get_alpha_beta(self.training, x.is_cuda)
+                alpha, beta = shake_get_alpha_beta(is_training=self.training,
+                                                   is_cuda=x.is_cuda,
+                                                   method=self.config['shake_alpha_beta_method'])
```
nabenabe0928, line 343: `function`, but not `method`?
If it is not hard to change, `shake_shake_update_func` is more precise based on the original paper.

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
```diff
@@ -92,15 +92,39 @@ def backward(ctx: typing.Any,
...
+    Currently, this function supports `even-even`, `shake-even` and `shake-shake`
```
nabenabe0928, line 103: Do we add `shake-keep` and `even-shake` in the future?
```suggestion
    Currently, this function supports `even-even`, `shake-even` and `shake-shake`
    TODO: `shake-keep` and `even-shake`
```

requirements.txt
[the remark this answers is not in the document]
nabenabe0928: I think it is about `sobol_seq` deprecation in `scipy>1.7.0`

autoPyTorch/api/base_task.py
```diff
@@ -1333,6 +1340,215 @@ def fit_pipeline(self,
...
+                self._stopwatch.start_task(traditional_task_name)
+                elapsed_time = self._stopwatch.wall_elapsed(ensemble_fit_task_name)
+                time_for_traditional = int(
+                    time_for_task - elapsed_time
+                )
+                self._do_traditional_prediction(
+                    func_eval_time_limit_secs=func_eval_time_limit_secs,
+                    time_left=time_for_traditional,
+                )
+                self._stopwatch.stop_task(traditional_task_name)
```
nabenabe0928, line 1458: Could you separate this process as a function?

```suggestion
        def run_traditional_ml(self, task_name: str, runtime_limit: int) -> None:
            if STRING_TO_TASK_TYPES[self.task_type] in REGRESSION_TASKS:
                self._logger.warning("Traditional Pipeline is not enabled for regression. Skipping...")
            else:
                traditional_task_name = 'runTraditional'
                self._stopwatch.start_task(traditional_task_name)
                elapsed_time = self._stopwatch.wall_elapsed(task_name)
                time_for_traditional = int(runtime_limit)
                self._do_traditional_prediction(
                    func_eval_time_limit_secs=func_eval_time_limit_secs,
                    time_left=time_for_traditional,
                )
                self._stopwatch.stop_task(traditional_task_name)
```

autoPyTorch/api/base_task.py
```diff
@@ -1333,6 +1340,215 @@ def fit_pipeline(self,
...
+                )
+                func_eval_time_limit_secs = time_for_task
+
+            # Make sure that at least 2 models are created for the ensemble process
+            num_models = time_for_task // func_eval_time_limit_secs
+            if num_models < 2:
+                func_eval_time_limit_secs = time_for_task // 2
+                self._logger.warning(
+                    "Capping the func_eval_time_limit_secs to {} to have "
+                    "time for a least 2 models to ensemble.".format(
```
nabenabe0928, line 1432:
```suggestion
                    "time for at least 2 models to ensemble.".format(
```

autoPyTorch/api/base_task.py
```diff
@@ -1333,6 +1340,215 @@ def fit_pipeline(self,
...
+            self
+        """
+        # Make sure that input is valid
+        if self.dataset is None or self.opt_metric is None:
+            raise ValueError("fit_ensemble() can only be called after `search()`. "
+                             "Please call the `search()` method of {} prior to "
+                             "fit_ensemble().".format(self.__class__.__name__))
+
+        if precision not in [16, 32, 64]:
+            raise ValueError("precision must be one of 16, 32, 64. Got {}".format(precision))
```
nabenabe0928, line 1404:
```suggestion
            raise ValueError("precision must be one of 16, 32, 64, but got {}".format(precision))
```

autoPyTorch/api/base_task.py
```diff
@@ -1333,6 +1340,215 @@ def fit_pipeline(self,
...
+                    time_left=time_for_traditional,
+                )
+                self._stopwatch.stop_task(traditional_task_name)
+
+        elapsed_time = self._stopwatch.wall_elapsed(ensemble_fit_task_name)
+        time_left_for_ensemble = int(time_for_task - elapsed_time)
+        manager = self._init_ensemble_builder(
+            time_left_for_ensembles=time_left_for_ensemble,
+            optimize_metric=self.opt_metric if optimize_metric is None else optimize_metric,
+            precision=self.precision if precision is None else precision,
```
nabenabe0928, line 1465: So here are the lines of `self.precision` and `self.opt_metric`.
Since we are taking `self.opt_metric` and `self.precision` from the arguments according to the current implementation, it seems to me that the deprecation of the instance variables does not make much difference.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -42,74 +99,54 @@ def _fit(
...
-            # Clear whatever null column markers we had previously
-            self.null_columns.clear()
-            if np.any(pd.isnull(X)):
-                for column in X.columns:
-                    if X[column].isna().all():
-                        self.null_columns.add(column)
-                        X[column] = pd.to_numeric(X[column])
-                        # Also note this change in self.dtypes
-                        if len(self.dtypes) != 0:
-                            self.dtypes[list(X.columns).index(column)] = X[column].dtype
```
nabenabe0928, line 62: It is also replaced by some lines (then which lines?)?

requirements.txt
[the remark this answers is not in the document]
ravinkohli: True so we can have it as scipy < 1.7.0

autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
ravinkohli: sure

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
ravinkohli: I don't think so. We decided on these as they gave the best results in the paper.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -291,70 +323,66 @@ def _get_columns_to_encode(
...
-                        "before feeding it to AutoPyTorch.".format(
-                            column,
-                            X[column].dtype.name,
-                        )
+            elif is_numeric_dtype(column_dtype):
+                feat_type.append('numerical')
+                numerical_columns.append(column)
+            elif column_dtype == 'object':
+                # TODO verify how would this happen when we always convert the object dtypes to category
+                raise ValueError(
```
nabenabe0928, line 356:
```suggestion
                raise TypeError(
```

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -291,70 +323,66 @@ def _get_columns_to_encode(
...
+                    "You can cast it to a valid dtype using pandas.Series.astype."
+                    "If you are working with string objects, the following "
+                    "tutorial illustrates how to work with text data: "
+                    "https://scikit-learn.org/stable/tutorial/text_analytics/working_with_text_data.html".format(
+                        # noqa: E501
+                        err_msg,
                     )
+                )
+            elif pd.core.dtypes.common.is_datetime_or_timedelta_dtype(column_dtype):
+                raise ValueError(
```
nabenabe0928, line 367:
```suggestion
                raise TypeError(
```

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -291,70 +323,66 @@ def _get_columns_to_encode(
...
+                    "{} Convert the time information to a numerical value"
+                    " before feeding it to AutoPyTorch. "
+                    "One example of the conversion can be found on "
+                    "https://stats.stackexchange.com/questions/311494/".format(err_msg)
+                )
             else:
-                feat_type.append('numerical')
-        return enc_columns, feat_type
+                raise ValueError(
```
nabenabe0928, line 374:
```suggestion
                raise TypeError(
```

autoPyTorch/api/base_task.py
```diff
@@ -1333,6 +1340,215 @@ def fit_pipeline(self,
...
+                self._stopwatch.start_task(traditional_task_name)
+                elapsed_time = self._stopwatch.wall_elapsed(ensemble_fit_task_name)
+                time_for_traditional = int(
+                    time_for_task - elapsed_time
+                )
+                self._do_traditional_prediction(
+                    func_eval_time_limit_secs=func_eval_time_limit_secs,
+                    time_left=time_for_traditional,
+                )
+                self._stopwatch.stop_task(traditional_task_name)
```
nabenabe0928, line 1458: Could you separate this process as a function?

```suggestion
        def run_traditional_ml(self, task_name: str, runtime_limit: int) -> None:
            if STRING_TO_TASK_TYPES[self.task_type] in REGRESSION_TASKS:
                self._logger.warning("Traditional Pipeline is not enabled for regression. Skipping...")
            else:
                traditional_task_name = 'runTraditional'
                self._stopwatch.start_task(traditional_task_name)
                elapsed_time = self._stopwatch.wall_elapsed(task_name)
                time_for_traditional = int(runtime_limit)
                self._do_traditional_prediction(
                    func_eval_time_limit_secs=func_eval_time_limit_secs,
                    time_left=time_for_traditional,
                )
                self._stopwatch.stop_task(traditional_task_name)
```

nabenabe0928 (changes requested):

autoPyTorch/api/base_task.py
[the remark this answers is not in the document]
nabenabe0928: This comment also should be addressed.

autoPyTorch/api/base_task.py
nabenabe0928: I think this comment was addressed in the PR#291, but it got back when merging?

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
```diff
@@ -92,15 +92,39 @@ def backward(ctx: typing.Any,
...
+    if method == 'shake-shake':
+        beta = torch.rand(1)
+    elif method in ['shake-even', 'even-even']:
+        beta = torch.FloatTensor([0.5])
+    elif method == 'M3':
+        # Table 4 in the paper `Shake-Shake regularization`
+        rnd = torch.rand(1)
+        beta = torch.FloatTensor(
+            [rnd * (0.5 - alpha) + alpha if alpha < 0.5 else rnd * (alpha - 0.5) + 0.5]
+        )
```
nabenabe0928, line 125:
```suggestion
    alpha = torch.FloatTensor([0.5]) if method.startswith('even') else torch.rand(1)

    if method.endswith('even'):
        beta = torch.FloatTensor([0.5])
    elif method.endswith('shake'):
        beta = torch.rand(1)
    elif method == 'M3':
        # Table 4 in the paper `Shake-Shake regularization`
        rnd = torch.rand(1)
        beta = torch.FloatTensor(
            [rnd * (0.5 - alpha) + alpha if alpha < 0.5 else rnd * (alpha - 0.5) + 0.5]
        )
```

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
```diff
@@ -92,15 +92,39 @@ def backward(ctx: typing.Any,
 shake_drop = ShakeDropFunction.apply
 
 
-def shake_get_alpha_beta(is_training: bool, is_cuda: bool
-                         ) -> typing.Tuple[torch.tensor, torch.tensor]:
+def shake_get_alpha_beta(
+        is_training: bool,
+        is_cuda: bool,
+        method: str
```
nabenabe0928, line 98: The indents

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
```diff
@@ -115,14 +139,14 @@ def shake_drop_get_bl(
...
         is_training: bool,
         is_cuda: bool
-) -> torch.tensor:
+) -> torch.Tensor:
     pl = 1 - ((block_index + 1) / num_blocks) * (1 - min_prob_no_shake)
 
     if is_training:
         # Move to torch.randn(1) for reproducibility
-        bl = torch.tensor(1.0) if torch.randn(1) <= pl else torch.tensor(0.0)
+        bl = torch.Tensor(1.0) if torch.randn(1) <= pl else torch.Tensor(0.0)
```
nabenabe0928, line 147: Why normal distribution?
Could you give me the reference of the implementation or the paper?
I think we need the doc-string.

autoPyTorch/pipeline/components/training/trainer/RowCutMixTrainer.py
```diff
@@ -28,29 +28,31 @@ def data_preparation(self, X: np.ndarray, y: np.ndarray,
...
 
-        size = X.shape[1]
-        indices = torch.tensor(self.random_state.choice(range(1, size), max(1, np.int32(size * lam)),
+        cut_column_indices = torch.tensor(self.random_state.choice(range(n_columns), max(1, np.int32(n_columns * lam)),
                                                         replace=False))
 
-        X[:, indices] = X[index, :][:, indices]
+        # Replace the values in `cut_indices` columns with
+        # the values from `permed_indices`
+        X[:, cut_indices] = X[shuffled_indices, :][:, cut_column_indices]
```
nabenabe0928, line 44:
```suggestion
        # Replace the values in `cut_column_indices` columns with
        # the values from `permed_indices`
        X[:, cut_column_indices] = X[shuffled_indices, :][:, cut_column_indices]
```

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
ravinkohli: You can check the implementation and the paper from [here](https://github.com/imenurok/ShakeDrop). The exactly lines from the paper are `bl ∈ {0, 1} is a Bernoulli random variable with the probability P(bl = 1) = E[bl] = pl`. I think from this we can see that the current code is doing exactly as the paper says. 

P.S. I am not the original author of this function.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -9,15 +9,75 @@
...
+    Given a dictionary of preprocessors, this function
+    creates a sklearn column transformer with appropriate
+    columns associated with their preprocessors.
+    Args:
+        preprocessors (Dict[str, List]):
+            Dictionary containing list of numerical and categorical preprocessors.
+        numerical_columns (List[int]):
+            List of names of numerical columns
+        categorical_columns (List[int]):
+            List of names of categorical columns
```
nabenabe0928, line 37:
```suggestion
        preprocessors (Dict[str, List[BaseEstimator]]):
            Dictionary containing list of numerical and categorical preprocessors.
        numerical_columns (List[str]):
            List of names of numerical columns
        categorical_columns (List[str]):
            List of names of categorical columns
```

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -39,87 +99,63 @@ def _fit(
...
+
+            # The column transformer reorders the feature types
+            # therefore, we need to change the order of columns as well
+            # This means categorical columns are shifted to the right
+            def comparator(cmp1: str, cmp2: str) -> int:
+                """ Order so that categorical columns come right and numerical columns come left """
+                choices = ['categorical', 'numerical']
+                if cmp1 not in choices or cmp2 not in choices:
+                    raise ValueError('The comparator for the column order only accepts {}, '
+                                                'but got {} and {}'.format(choices, cmp1, cmp2))
```
nabenabe0928, line 138: flake8? (indent?)

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -246,43 +267,51 @@ def _check_data(
...
-            if not X.select_dtypes(include='object').empty:
+            exist_object_columns = has_object_columns(X.dtypes.values)
+            if exist_object_columns:
                 X = self.infer_objects(X)
 
             # Define the column to be encoded here as the feature validator is fitted once
             # per estimator
-            enc_columns, _ = self._get_columns_to_encode(X)
-
+            # enc_columns, _ = self._get_columns_to_encode(X)
```
nabenabe0928, line 279: Can we remove it?

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -246,43 +267,51 @@ def _check_data(
...
-                    raise ValueError("Changing the dtype of the features after fit() is "
-                                     "not supported. Fit() method was called with "
-                                     "{} whereas the new features have {} as type".format(self.dtypes,
-                                                                                          dtypes,
-                                                                                          )
-                                     )
+                dtypes_diff = [s_dtype != dtype for s_dtype, dtype in zip(self.dtypes, dtypes)]
+                if any(dtypes_diff):
+                    if self.all_nan_columns is not None and len(self.all_nan_columns) > 0:
+                        if len(set(X.columns[dtypes_diff]).difference(self.all_nan_columns)) != 0:
```
nabenabe0928, line 296: If I understand the logic correctly, this condition `self.all_nan_columns is not None and len(self.all_nan_columns) > 0` does not matter?
Because if we remove by this condition, we end up with allowing some cases where dtypes are different in the training and test.

For example, when we have column `A (int)`, `B (int)`  in training and `A (int)`, `B (float)` in test where `B` is not `all_nan` in both training and test.
Then we cannot filter such cases.

autoPyTorch/pipeline/components/training/trainer/RowCutMixTrainer.py
```diff
@@ -28,29 +28,31 @@ def data_preparation(self, X: np.ndarray, y: np.ndarray,
...
 
         r = self.random_state.rand(1)
         if beta <= 0 or r > self.alpha:
-            return X, {'y_a': y, 'y_b': y[index], 'lam': 1}
+            return X, {'y_a': y, 'y_b': y[shuffled_indices], 'lam': 1}
 
-        size = X.shape[1]
-        indices = torch.tensor(self.random_state.choice(range(1, size), max(1, np.int32(size * lam)),
+        cut_column_indices = torch.tensor(self.random_state.choice(range(n_columns), max(1, np.int32(n_columns * lam)),
                                                         replace=False))
```
nabenabe0928, line 40: indent

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
```diff
@@ -110,19 +130,19 @@ def shake_get_alpha_beta(is_training: bool, is_cuda: bool
...
-        is_training: bool,
-        is_cuda: bool
-) -> torch.tensor:
+    block_index: int,
+    min_prob_no_shake: float,
+    num_blocks: int,
+    is_training: bool,
+    is_cuda: bool
+) -> torch.Tensor:
     pl = 1 - ((block_index + 1) / num_blocks) * (1 - min_prob_no_shake)
```
nabenabe0928, line 139:
```suggestion
    """
    The sampling of Bernoulli random variable
    based on Eq. (4) in the paper

    Args:
        block_index (int): The index of the block from the input layer
        min_prob_no_shake (float): The initial shake probability
        num_blocks (int): The total number of building blocks
        is_training (bool): Whether it is training
        is_cuda (bool): Whether the tensor is on CUDA

    Returns:
        bl (torch.Tensor): a Bernoulli random variable in {0, 1}

    Reference:
        ShakeDrop Regularization for Deep Residual Learning
        Yoshihiro Yamada et. al. (2020)
        paper: https://arxiv.org/pdf/1802.02375.pdf
        implementation: https://github.com/imenurok/ShakeDrop
    """
    # Linear decay of the shake drop based on Eq. (5) in the reference
    pl = 1 - ((block_index + 1) / num_blocks) * (1 - min_prob_no_shake)
```

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
nabenabe0928: Uum, if I understand the paper correctly, this should be `torch.rand(1)`, but not `torch.randn(1)`...
Basically, this is the definition the Bernouilli random variable as well.

nabenabe0928 (changes requested):

autoPyTorch/data/tabular_feature_validator.py
nabenabe0928: I mean we may have a case of which `len(self.all_nan_columns) == 0` but we should remove.
Let me make sure my point.
First, there are three conditions:
`C1`. `any(dtypes_diff)`: At least one of the columns has different data type between training and test datasets
`C2`. `len(self.all_nan_columns) > 0`: There is at least one column in the dataset that only has nan in the column
`C3`. `len(set(X.columns[dtypes_diff]).difference(self.all_nan_columns)) != 0`: A set of columns that have different data type in training and test and a set of columns that have only nan in test are identical

We raise the error if and only `C1`, `C2` and `C3` are satisfied.
But I am saying the condition `C1 and C3` also should be failed.

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
```diff
@@ -92,15 +92,35 @@ def backward(ctx: typing.Any,
...
-                         ) -> typing.Tuple[torch.tensor, torch.tensor]:
+def shake_get_alpha_beta(
+    is_training: bool,
+    is_cuda: bool,
+    method: str
+) -> typing.Tuple[torch.Tensor, torch.Tensor]:
+    """
+    The methods used in this function have been introduced in 'ShakeShake Regularisation'
+    https://arxiv.org/abs/1705.07485. The names have been taken from the paper as well.
+    Currently, this function supports `even-even`, `shake-even` and `shake-shake`
```
nabenabe0928, line 103:
```suggestion
    Currently, this function supports `even-even`, `shake-even`, `shake-shake` and `M3`.
```

test/test_data/test_feature_validator.py
```diff
@@ -317,6 +317,93 @@ def test_featurevalidator_get_columns_to_encode():
...
+        ],
+        dtype='category',
+    )
+
+    validator.fit(df_train)
+    transformed_df_train = validator.transform(df_train)
+    transformed_df_test = validator.transform(df_test)
+
+    assert np.array_equal(transformed_df_train, np.array([[0, 1], [1, 0], [0, 1]], dtype=float))
+    assert np.array_equal(transformed_df_test, np.array([[1, 0], [1, 0], [0, 1]], dtype=float))
```
nabenabe0928, line 350: Could you add the meaning of the answer sets?

test/test_data/test_feature_validator.py
```diff
@@ -317,6 +317,93 @@ def test_featurevalidator_get_columns_to_encode():
     assert feature_types == ['numerical', 'numerical', 'categorical', 'categorical']
 
 
+def test_featurevalidator_remove_nan_catcolumns():
+    """
+    Make sure categorical columns that have only nan values are removed.
+    """
+    # First case, there exist null columns in the train set
+    # and the same columns are not all null for the test set.
```
nabenabe0928, line 325: Only one of the columns are not all nan?

test/test_data/test_feature_validator.py
```diff
@@ -311,10 +312,98 @@ def test_featurevalidator_get_columns_to_encode():
...
+
+
+def test_feature_validator_remove_nan_catcolumns():
+    """
+    Make sure categorical columns that have only nan values are removed.
+    The ans arrays contain the final output after calling transform on
+    datasets, this includes fitting and transforming a column transformer
+    containing simple imputation for both categorical and numerical
+    columns, scaling for numerical columns and one hot encoding for
+    categorical columns.
```
nabenabe0928, line 342: Thanks for the description.
Could you tell me when we will get `0` and when we will get `1` as outputs?

autoPyTorch/data/tabular_feature_validator.py
nabenabe0928: Yes, I got your point.
But I am saying your condition does not cover everything, does it?
Yours will miss the following case:
```
train_data = {'A': 1}
test_data = {'A': 1.0}
```

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -9,15 +9,98 @@
...
+    preprocessors['numerical'] = [numerical_imputer, standard_scaler]
+
+    return preprocessors
+
+
 class TabularFeatureValidator(BaseFeatureValidator):
+
+    @staticmethod
+    def _comparator(cmp1: str, cmp2: str) -> int:
+        """Order so that categorical columns come right and numerical columns come left
```
ArlindKadra, line 83: Is it not the exact opposite ?

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -39,87 +122,54 @@ def _fit(
...
+                categorical_columns=categorical_columns,
+            )
+
+            # Mypy redefinition
+            assert self.column_transformer is not None
+            self.column_transformer.fit(X)
+
+            # The column transformer reorders the feature types
+            # therefore, we need to change the order of columns as well
+            # This means categorical columns are shifted to the right
```
ArlindKadra, line 155: to the left right ?

autoPyTorch/api/base_task.py
```diff
@@ -895,31 +950,25 @@ def _search(
             )
 
         # ============> Run dummy predictions
-        dummy_task_name = 'runDummy'
-        self._stopwatch.start_task(dummy_task_name)
-        self._do_dummy_prediction()
-        self._stopwatch.stop_task(dummy_task_name)
+        # We only want to run dummy predictions in case we want to build an ensemble
+        if self.ensemble_size > 0:
```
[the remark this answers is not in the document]
ArlindKadra: I see that in the advanced example that you made you left it to 0.

autoPyTorch/api/base_task.py
```diff
@@ -866,7 +921,7 @@ def _search(
         # If no dask client was provided, we create one, so that we can
         # start a ensemble process in parallel to smbo optimize
         if (
```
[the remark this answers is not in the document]
ravinkohli: I think pep8 recommends using paranthesis when we have multiple lines. See [this](https://stackoverflow.com/questions/5253348/very-long-if-statement-in-python)

autoPyTorch/api/base_task.py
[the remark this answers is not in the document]
ravinkohli: yeah, this is in the case where we don't want an ensemble at all. As we already have `SingleBest` ([here](https://github.com/automl/Auto-PyTorch/blob/baa7ab87fe64cf6d8e1f9638ff029147e6b2a310/autoPyTorch/ensemble/singlebest_ensemble.py#L14), I think we should not use ensemble_size=1 for fitting a single configuration as we did during the cocktail experiments.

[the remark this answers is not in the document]
ArlindKadra: okay, so then we can also update the code there and probably save some time, nothing would break right ?

autoPyTorch/api/base_task.py
```diff
@@ -936,26 +985,12 @@ def _search(
...
-                seed=self.seed,
-                max_iterations=None,
-                read_at_most=sys.maxsize,
-                ensemble_memory_limit=self._memory_limit,
-                random_state=self.seed,
-                precision=precision,
-                logger_port=self._logger_port,
-            )
+            proc_ensemble = self._init_ensemble_builder(time_left_for_ensembles=time_left_for_ensembles,
+                                                        ensemble_size=self.ensemble_size,
```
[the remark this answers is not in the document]
ArlindKadra: Yes, this is also what I meant with my second comment. it makes sense, I will resolve this one.

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
```diff
@@ -110,19 +130,36 @@ def shake_get_alpha_beta(is_training: bool, is_cuda: bool
...
+        Yoshihiro Yamada et. al. (2020)
+        paper: https://arxiv.org/pdf/1802.02375.pdf
+        implementation: https://github.com/imenurok/ShakeDrop
+    """
     pl = 1 - ((block_index + 1) / num_blocks) * (1 - min_prob_no_shake)
 
     if is_training:
         # Move to torch.randn(1) for reproducibility
-        bl = torch.tensor(1.0) if torch.randn(1) <= pl else torch.tensor(0.0)
+        bl = torch.Tensor(1.0) if torch.rand(1) <= pl else torch.Tensor(0.0)
```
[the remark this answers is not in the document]
ravinkohli: yeah there was an error with `torch.tensor` in the github actions and `torch.Tensor` fixed it. But now I cant reproduce the error and both of them work for torch 1.7, 1.8 and 1.9.1. Do you think we should use something more explicit like `torch.as_tensor`, see [here](https://pytorch.org/docs/stable/generated/torch.as_tensor.html#torch.as_tensor)

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -9,15 +9,98 @@
...
+    return ColumnTransformer([
+        ('categorical_pipeline', categorical_pipeline, categorical_columns),
+        ('numerical_pipeline', numerical_pipeline, numerical_columns)],
+        remainder='drop'
+    )
+
+
+def get_tabular_preprocessors() -> Dict[str, List[BaseEstimator]]:
+    """
+    This function creates a Dictionary containing list
```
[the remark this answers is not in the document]
ravinkohli: isnt there a line break before the return? Do you want 2? I'll fix the documentation though.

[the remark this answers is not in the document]
ravinkohli: yup, sure

PR #287, opened 2021-09-02
Shake Shake updates

ravinkohli:


examples/tabular/40_advanced/example_custom_configuration_space.py
```diff
@@ -54,6 +54,15 @@ def get_search_space_updates():
                    hyperparameter='ResNetBackbone:dropout',
                    value_range=[0, 0.5],
                    default_value=0.2)
+    updates.append(node_name='network_backbone',
+                   hyperparameter='ResNetBackbone:multi_branch_choice',
+                   value_range=['shake-shake'],
+                   default_value='shake-shake')
+    updates.append(node_name='network_backbone',
```
nabenabe0928, line 61: Is it ok to have the same `node_name`?

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
```diff
@@ -92,15 +92,31 @@ def backward(ctx: typing.Any,
...
         return result if not is_cuda else (result[0].cuda(), result[1].cuda())
 
     # TODO implement other update methods
-    alpha = torch.rand(1)
-    beta = torch.rand(1)
+    if method != 'even-even':
+        alpha = torch.rand(1)
+    else:
+        alpha = torch.FloatTensor([0.5])
+    if method in ['shake-shake', 'shake-drop']:
```
nabenabe0928, line 110: space between `if` statements
```suggestion
    else:
        alpha = torch.FloatTensor([0.5])

    if method in ['shake-shake', 'shake-drop']:
```

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
```diff
@@ -92,15 +92,31 @@ def backward(ctx: typing.Any,
...
+        beta = torch.rand(1)
+    elif 'even' in method:  # 'shake-even' or 'even-even'
+        beta = torch.FloatTensor([0.5])
+    elif method == 'M3':
+        beta = torch.FloatTensor(
+            [torch.rand(1)*(0.5 - alpha)*alpha if alpha < 0.5 else torch.rand(1)*(alpha - 0.5)*alpha]
+        )
+    else:
```
nabenabe0928, line 118: What is M3?
Are there any references?

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
```diff
@@ -92,15 +92,31 @@ def backward(ctx: typing.Any,
...
     # TODO implement other update methods
-    alpha = torch.rand(1)
-    beta = torch.rand(1)
+    if method != 'even-even':
+        alpha = torch.rand(1)
+    else:
+        alpha = torch.FloatTensor([0.5])
+    if method in ['shake-shake', 'shake-drop']:
+        beta = torch.rand(1)
+    elif 'even' in method:  # 'shake-even' or 'even-even'
```
nabenabe0928, line 112: Then shouldn't it be `method in ['shake-even', 'even-even']`?
Probably we should make an enumerator for the method.

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
```diff
@@ -92,15 +92,31 @@ def backward(ctx: typing.Any,
...
+        is_cuda: bool,
+        method: str
+) -> typing.Tuple[torch.tensor, torch.tensor]:
+
     if not is_training:
         result = (torch.FloatTensor([0.5]), torch.FloatTensor([0.5]))
         return result if not is_cuda else (result[0].cuda(), result[1].cuda())
 
     # TODO implement other update methods
-    alpha = torch.rand(1)
```
nabenabe0928, line 102: Overall, I am not sure if it is ok to have `alpha + beta != 1`

autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
```diff
@@ -180,9 +187,12 @@ def get_hyperparameter_search_space(
 
         if skip_connection_flag:
 
+            shake_shake_flag = False
             shake_drop_prob_flag = False
             if 'shake-drop' in multi_branch_choice.value_range:
                 shake_drop_prob_flag = True
+            elif 'shake-shake' in multi_branch_choice.value_range:
+                shake_shake_flag = True
```
nabenabe0928, line 195: Is the condition for `shake_shake_flag` correct?
I found the hierarchy of those two a bit weird.
It makes sense `shake_drop_prob_flag=False` if `not shake_shake_flag`,
but the other way seems strange.
```suggestion
            shake_drop_prob_flag = 'shake-drop' in multi_branch_choice.value_range
            shake_shake_flag = not shake_drop_prob_flag and 'shake-shake' in multi_branch_choice.value_range
```

autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
```diff
@@ -188,9 +196,12 @@ def get_hyperparameter_search_space(  # type: ignore[override]
 
         if skip_connection_flag:
 
+            shake_shake_flag = False
             shake_drop_prob_flag = False
             if 'shake-drop' in multi_branch_choice.value_range:
                 shake_drop_prob_flag = True
+            elif 'shake-shake' in multi_branch_choice.value_range:
```
nabenabe0928, line 203: Same in the L190--195 of `autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py`

autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
ravinkohli: no actually both shake-shake and shake-drop can be in the value range for the multi branch choice. and I think you are right. There is a bug here it should not be elif but just if 'shake-shake' in multi_branch_choice.value_range.

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
ravinkohli: it was mentioned in the paper this way

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
ravinkohli: sure, we can improve this. Let's do it after we have meaningful results. For now, it's logically correct.

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
ravinkohli: yes, we'll add them and probably even change the name. It is mentioned in the ShakeShake regularisation paper.

examples/tabular/40_advanced/example_custom_configuration_space.py
ravinkohli: what do you mean?

ArlindKadra: I guess he means having `network_backbone` for both updates as node names although the hyperparameter name changes.

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
ArlindKadra: on that matter, `shake-drop` will only have random alpha and beta, if we really want to have the search space of the paper, we should pass even, even also for shake-drop.

ArlindKadra (commented):
Looks good, I have a suggestion and a few questions.

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
ravinkohli: okay, I can make shake_shake_methods to be dependent on both shake-shake and shake-drop. and add even-even, shake-shake, shake-even, m3 as the methods for alpha and beta. Does that make sense?

autoPyTorch/api/base_task.py
```diff
@@ -1024,18 +1044,12 @@ def _search(
                 pd.DataFrame(self.ensemble_performance_history).to_json(
                     os.path.join(self._backend.internals_directory, 'ensemble_history.json'))
 
-        self._logger.info("Closing the dask infrastructure")
-        self._close_dask_client()
-        self._logger.info("Finished closing the dask infrastructure")
-
         if load_models:
             self._logger.info("Loading models...")
             self._load_models()
```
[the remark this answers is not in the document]
ravinkohli: I dont think so, As it only involves reading from disk on the main process. It is not interacting with the dask infrastructure here.

autoPyTorch/api/base_task.py
```diff
@@ -1506,7 +1520,7 @@ def predict(
 
         predictions = self.ensemble_.predict(all_predictions)
 
-        self._clean_logger()
+        self._cleanup()
```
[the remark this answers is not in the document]
ravinkohli: by the time that the code reaches fit pipeline. The Dask infrastructure is usually already closed. So, trying to close it would make no difference as we are checking if it is open or not. Also, this function does not need the dask infrastructure so if it is called in isolation we would not have the dask server open. I think it's safer to always close the servers when we are exiting a function just to be safe. I dont think we would run into any issues because of this

examples/tabular/40_advanced/example_custom_configuration_space.py
ravinkohli: yeah, that's totally fine. Do you see a problem with it @nabenabe0928 ?

autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py
```diff
@@ -180,9 +187,12 @@ def get_hyperparameter_search_space(
 
         if skip_connection_flag:
 
+            shake_shake_flag = False
             shake_drop_prob_flag = False
             if 'shake-drop' in multi_branch_choice.value_range:
                 shake_drop_prob_flag = True
+            if 'shake-shake' in multi_branch_choice.value_range:
```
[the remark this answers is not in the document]
ravinkohli: done

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
ArlindKadra: Yes it does for the moment.

autoPyTorch/api/base_task.py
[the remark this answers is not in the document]
ArlindKadra: Okay, that would not hurt I guess. As long as there is no error being thrown if it is already closed.

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
ravinkohli: yeah I added them

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
```diff
@@ -103,13 +103,14 @@ def shake_get_alpha_beta(
...
-    if method != 'even-even':
-        alpha = torch.rand(1)
-    else:
+    if method == 'even-even':
         alpha = torch.FloatTensor([0.5])
-    if method in ['shake-shake', 'shake-drop']:
+    else:
+        alpha = torch.rand(1)
+
+    if method == 'shake-shake':
```
ArlindKadra, line 111: https://github.com/automl/Auto-PyTorch/blob/292178197068a59a6a9c2ec869349dfec0ffd96f/autoPyTorch/pipeline/components/setup/network_backbone/ResNetBackbone.py#L347, you should change the following here too for 'shake_drop' and pass the config.

ArlindKadra (changes requested):

autoPyTorch/pipeline/components/setup/network_backbone/utils.py
ravinkohli: oh yeah sorry. Thanks for the catch

ArlindKadra (approved):
Looks better now.

PR #291, opened 2021-10-01
Enhancement for the tabular validator.

ArlindKadra:
- Refactoring the tabular validator parts.


autoPyTorch/data/base_feature_validator.py
```diff
@@ -152,4 +154,30 @@ def transform(
             np.ndarray:
                 The transformed array
         """
+
+        raise NotImplementedError()
+
+    def list_to_dataframe(
```
ravinkohli, line 160: we dont need to have this on the base class as it will only be used for the tabular data right?

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -63,18 +65,25 @@ def get_tabular_preprocessors() -> Dict[str, List[BaseEstimator]]:
...
-                                                    copy=False))
-    preprocessors['numerical'].append(StandardScaler(with_mean=True, with_std=True, copy=False))
+    preprocessors['categorical'].append(
+        OneHotEncoder(
+            categories='auto',
+            sparse=False,
+            handle_unknown='ignore',
+        )
+    )
+    preprocessors['numerical'].append(
```
ravinkohli, line 75: are we going to scale inside the pipeline?

autoPyTorch/data/base_feature_validator.py
ArlindKadra: It is referenced in ` fit`. It does not look nice when you reference a method that a subclass has and not the class itself.

autoPyTorch/data/tabular_feature_validator.py
ArlindKadra: Yes, in the future yes.

autoPyTorch/data/base_feature_validator.py
ravinkohli: oh I see, thanks for this then

ArlindKadra: I did think about the fact that `list_to_dataframe ` and where it should be placed in the future, but I thought that belongs better in another PR that does consider more than the tabular domain.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -268,13 +279,13 @@ def _check_data(
             X = cast(pd.DataFrame, X)
 
             # Handle objects if possible
-            if not X.select_dtypes(include='object').empty:
+            object_columns_indicator = has_object_columns(X.dtypes)
```
ravinkohli, line 282: is this a faster way to test for column with 'object' type?

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -96,24 +105,27 @@ def _fit(
...
+            # save all the information about the column order and data types
+            self._check_data(X)
 
         if hasattr(X, "iloc") and not scipy.sparse.issparse(X):
-            X = cast(pd.DataFrame, X)
-
-            if not X.select_dtypes(include='object').empty:
-                X = self.infer_objects(X)
 
-            self._check_data(X)
```
ravinkohli, line 107: should we not have a check data for pandas data frame?

autoPyTorch/data/tabular_feature_validator.py
ArlindKadra: Hmm, I tried to do it in a faster way with the function I added, `has_object_columns` which does not return a ` DataFrame`.

autoPyTorch/data/tabular_feature_validator.py
ArlindKadra: To my understanding no, because it is already checked in `fit` at the `base_feature_validator`, here:

https://github.com/automl/Auto-PyTorch/blob/217c38dd43b012d3839f757cce7877299aac74cc/autoPyTorch/data/base_feature_validator.py#L93

ravinkohli: also set the self.data_type for both pandas?

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -189,10 +204,6 @@ def transform(
         if hasattr(X, "iloc") and not scipy.sparse.issparse(X):
             X = cast(pd.DataFrame, X)
 
-            # Also remove the object dtype for new data
-            if not X.select_dtypes(include='object').empty:
-                X = self.infer_objects(X)
```
ravinkohli, line 194: why do we not want to infer objects here? I think its needed to check if the datatype for a column has changed between train and test. See [this](https://github.com/automl/Auto-PyTorch/blob/217c38dd43b012d3839f757cce7877299aac74cc/autoPyTorch/data/tabular_feature_validator.py#L458)

autoPyTorch/data/tabular_feature_validator.py
ArlindKadra: With what I saw no, because it is already set once when calling `_check_data` from fit in the `base_feature_validator`. The type does not change, it is still a DataFrame.

autoPyTorch/data/tabular_feature_validator.py
ArlindKadra: To my understanding it is done a few lines of code later in `_check_data`, here: 

https://github.com/automl/Auto-PyTorch/blob/217c38dd43b012d3839f757cce7877299aac74cc/autoPyTorch/data/tabular_feature_validator.py#L284

autoPyTorch/data/base_feature_validator.py
ravinkohli: yeah sure

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: but do we not ned it for our experiments?

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: oh ohkay I forgot about that. Thanks

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: okay so there were duplications.

autoPyTorch/data/tabular_feature_validator.py
ArlindKadra: Ups, now I see what you mean. It was wrongly deleted during rebase, I just added it.

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: no I think its fine for now

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -481,14 +507,16 @@ def impute_nan_in_categories(self,
...
                     except Exception:
                         can_cast_as_number = False
                     if can_cast_as_number:
                         # In this case, we expect to have a number as category
                         # it might be string, but its value represent a number
-                        missing_value: Union[str, int] = '-1' if isinstance(X[column].dropna().values[0], str) else -1
+
+                        missing_value: Union[str, int] = '-1' if isinstance(first_value, str) else -1
                     else:
                         missing_value = 'Missing!'
```
nabenabe0928, line 522: L495--L504 (L523--L532)
dict_missing_value_per_col --> missing_value_symbols
missing_value --> missing_value_symbol (because padded value for missing_value must be a symbol)

Separate the process as a function named `get_unused_category_symbol`.
Plus, this function is easily performed by the following:
```
# Let's assume missing_value is already replaced by missing_value_symbol
if missing_value not in X[column].cat.categories:
    pass
elif isinstance(missing_value, str):
    max_length = max(len(c) for c in X[column].cat.categories)
    missing_value_symbol += '0' * max_length
else:
    # min_value is guaranteed to be negative since there exists -1 in categories
    # and min_value must be smaller than -1. So the symbol is always negative.
    min_value = min(c for c in X[column].cat.categories)
    missing_value_symbol = missing_value + min_value
```

In this case, the time complexity is `O(N)` in the worst case where N is the length of the categories array or the summation of the lengths of categories.
On the other hand, the previous implementation requires `O(N^2)` in the worst case (I do not believe it happens in practice, but the readability is not good in this implementation anyways, so I prefer this implementation).

nabenabe0928 (changes requested):
Hi thanks for the PR.
I checked the implementation and suggested some changes.

autoPyTorch/pipeline/components/setup/network_embedding/base_network_embedding.py
```diff
@@ -49,4 +49,4 @@ def build_embedding(self, num_input_features: np.ndarray, num_numerical_features
     #     # for i, category in enumerate(categories):
     #     #     num_input_features[num_numerical_columns + i, ] = len(category)
     #     # return num_numerical_columns, num_input_features
-    #     return None, None
+        return None, None
```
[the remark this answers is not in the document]
ArlindKadra: @ravinkohli can take over this question since it was his commit. I would need to double-check the implementation again for that.

autoPyTorch/data/tabular_feature_validator.py
ArlindKadra: @nabenabe0928 Thanks for the code, it pointed out the fact that my code was going through all features and not just return if it found the first object one.

How about this implementation

```
    def arlind_v2(X) -> bool:
        feature_types = X.dtypes
        for feature_type in feature_types:
            if pd.api.types.is_object_dtype(feature_type):
                return True
            else:
                return False
```

```
Old version
594 µs ± 14.6 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
1.35 ms ± 9.81 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
Arlind version
914 µs ± 11.5 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
917 µs ± 17.8 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
Shuhei version
107 µs ± 1.03 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
83 µs ± 695 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
Arlind v2 version
57.7 µs ± 684 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
57.9 µs ± 730 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

autoPyTorch/data/base_feature_validator.py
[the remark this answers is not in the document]
ArlindKadra: That is a good question, in my pov I do not think it needs to be overall, however, if you check the implementation:
https://github.com/automl/Auto-PyTorch/blob/5f672b550d2fdc25c2a3865c20a78e44f04e796a/autoPyTorch/data/tabular_feature_validator.py#L394

it needs access to `self` to use the logger. 

I do not think it fits at the base_feature_validator, especially when dealing with non-tabular data, but I thought that would be out of the scope of this PR. We can raise an issue and get back to this in the future. I am sure it can be done way better.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -63,18 +65,32 @@ def get_tabular_preprocessors() -> Dict[str, List[BaseEstimator]]:
...
+            copy=False,
+        )
+    )
+    preprocessors['numerical'].append(
+        StandardScaler(
+            with_mean=True,
+            with_std=True,
+            copy=False,
+        )
+    )
```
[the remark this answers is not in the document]
ArlindKadra: I like the suggestion, it looks pretty.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -142,21 +161,24 @@ def comparator(cmp1: str, cmp2: str) -> int:
 
             if len(categorical_columns) > 0:
                 self.categories = [
-                    # We fit an ordinal encoder, where all categorical
+                    # We fit an one-hot encoder, where all categorical
```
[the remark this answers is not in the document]
ArlindKadra: True.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -481,14 +507,16 @@ def impute_nan_in_categories(self,
             if X[column].isna().any():
                 if column not in self.dict_missing_value_per_col:
                     try:
-                        float(X[column].dropna().values[0])
+                        first_value = X[column].dropna().values[0]
+                        float(first_value)
                         can_cast_as_number = True
                     except Exception:
```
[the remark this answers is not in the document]
ArlindKadra: Why not, since we already worked on the module, we can also do the right error handling instead of a general one.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -446,17 +468,21 @@ def infer_objects(self, X: pd.DataFrame) -> pd.DataFrame:
...
-                if not is_numeric_dtype(X[column]):
+            # initial data types
+            data_types = X.dtypes
+            for index, column in enumerate(X.columns):
+                if not is_numeric_dtype(data_types[index]):
                     X[column] = X[column].astype('category')
-            self.object_dtype_mapping = {column: X[column].dtype for column in X.columns}
+            # only numerical attributes and categories
+            data_types = X.dtypes
+            self.object_dtype_mapping = {column: data_types[index] for index, column in enumerate(X.columns)}
```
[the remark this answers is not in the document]
ArlindKadra: Nice suggestion, I like it.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -481,14 +507,16 @@ def impute_nan_in_categories(self,
             if X[column].isna().any():
                 if column not in self.dict_missing_value_per_col:
                     try:
-                        float(X[column].dropna().values[0])
+                        first_value = X[column].dropna().values[0]
+                        float(first_value)
                         can_cast_as_number = True
                     except Exception:
                         can_cast_as_number = False
```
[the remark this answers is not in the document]
ArlindKadra: we do not want to save ` first_value` as float, or did I miss something.

autoPyTorch/data/tabular_feature_validator.py
ArlindKadra: @ravinkohli yes there were duplicates. @nabenabe0928 the ` infer_objects` is already called in a later function named _check_data, so it did not make sense to have it called before too.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -96,24 +112,27 @@ def _fit(
         # The final output of a validator is a numpy array. But pandas
         # gives us information about the column dtype
         if isinstance(X, np.ndarray):
+
             X = self.numpy_array_to_pandas(X)
+            # Replace the data type from the previously saved type.
+            self.data_type = type(X)
+            # save all the information about the column order and data types
+            self._check_data(X)
```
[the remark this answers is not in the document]
ArlindKadra: Hmm, well this is a bit complex. `_check_data ` is already called in `base_feature_validator`. If you pass a numpy array there, it will have a numpy data type, while above, it is converted to DataFrame, when you call` _check_data` again you will have a warning in the logger that the data type was changed, a warning that we trigger. I do not think it fits `_check_data` to return a type, the idea is to just verify that the data type is the same and the column order and types are the same. The problem here is that we convert it ourselves.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -481,14 +507,16 @@ def impute_nan_in_categories(self,
             if X[column].isna().any():
```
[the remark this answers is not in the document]
ArlindKadra: Matter of preference, but I am thinking of adding the else too for the rest of the process and that should in the future increase code coverage from different tools.

autoPyTorch/data/tabular_feature_validator.py
ArlindKadra: Answer below.

autoPyTorch/data/tabular_feature_validator.py
nabenabe0928: @ArlindKadra Thanks for the suggestion:)
I also thought in the same way because early-return is a good idea, but this implementation only looks at the first element, so you need to do the following:
```
    def arlind_v2(X) -> bool:
        feature_types = X.dtypes
        for feature_type in feature_types:
            if pd.api.types.is_object_dtype(feature_type):
                return True

        return False
```

autoPyTorch/data/tabular_feature_validator.py
nabenabe0928: Ah ok
So this function comes first then

autoPyTorch/data/tabular_feature_validator.py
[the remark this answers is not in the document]
nabenabe0928: I was thinking of something like the following:
```
def can_cast_as_number(value: Union[int, float, str]) -> bool:
    try:
        float(first_value)
        can_cast_as_number = True
    except ValueError:
        can_cast_as_number = False

first_value = X[column].dropna().values[0]
if can_cast_as_number(first_value):
    # process
```

Just in case, we have not done `first_value = float(first_value)`, so the type of `first_value` does not change.
```
def try_cast(value):
    try:
        float(value)
    except ValueError:
        pass


val = 1
print(type(val))
try_cast(val)
print(type(val))

val = 1.0
print(type(val))
try_cast(val)
print(type(val))

val = '1.0'
print(type(val))
try_cast(val)
print(type(val))

[Output]
<class 'int'>
<class 'int'>
<class 'float'>
<class 'float'>
<class 'str'>
<class 'str'>
```

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -504,13 +489,16 @@ def impute_nan_in_categories(self, X: pd.DataFrame) -> pd.DataFrame:
         # TypeError: '<' not supported between instances of 'int' and 'str'
         # in the encoding
         for column in self.enc_columns:
-            if X[column].isna().any():
+            # no missing values for categorical column
+            if not X[column].isna().any():
+                continue
+            else:
```
nabenabe0928, line 495: Ah thanks for addressing my comment:)

If you need to use `else` later, you can go back to the original version.
Just in case, what I was thinking was to remove the `else` completely, because we do not nest for this `else` after the line continuation.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -555,10 +543,8 @@ def has_object_columns(
...
-
-    if True in object_columns_indicator:
-        return True
-    else:
-        return False
+    for feature_type in feature_types:
+        if pd.api.types.is_object_dtype(feature_type):
+            return True
+        else:
+            return False
```
nabenabe0928, line 550: Probably you already read my comment beforehand, but you cannot detect the case like `feature_types = [int, object]`.

nabenabe0928 (commented):
I left just quick comments!

autoPyTorch/data/tabular_feature_validator.py
ArlindKadra: ups, very right, a typo and bug right there.

autoPyTorch/data/tabular_feature_validator.py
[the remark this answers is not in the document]
ArlindKadra: ah yeah yeah, I do know the value would not change, since you wrote not a method in parentheses I did not get the main idea. Sure, looks like it makes the code easier to read.

autoPyTorch/data/tabular_feature_validator.py
ArlindKadra: Yes I understood it, I have a bit of time without checking code coverage tools, but if I remember correctly when you only have an if and not the else, it counts your code as not fully covered, so for the sake of that and a nice metric value in the future, I was thinking of adding the else, as it is pretty redundant. If it is not the case now, we can remove it.

autoPyTorch/data/tabular_feature_validator.py
ArlindKadra: Yes, as soon as you pointed it out I noticed the bug that I introduced with the typo. On another note, the code now without that typo does not seem faster than your version, I guess numpy does have its perks.

autoPyTorch/data/tabular_feature_validator.py
ArlindKadra:
```
Old version
626 µs ± 26.7 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
1.42 ms ± 23.6 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
Arlind version
974 µs ± 5.84 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
966 µs ± 9.93 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
Shuhei version
109 µs ± 2.58 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
84.4 µs ± 980 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
Arlind v2 version
932 µs ± 10.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
59.5 µs ± 1.45 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

Does not seem faster anymore, awful when working with numerical values.

autoPyTorch/data/tabular_feature_validator.py
nabenabe0928: Sure, I did not know [the bug](https://github.com/nedbat/coveragepy/issues/198).
Then it is better like this based on the report?:
```
if skip:
    _dummy = True  # Heuristics to avoid a non-coverage issue
    continue

func1()
func2()
...
```

autoPyTorch/pipeline/components/setup/network_embedding/base_network_embedding.py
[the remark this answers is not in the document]
ravinkohli: hey, @nabenabe0928, As I previously explained in a previous PR, I'll fix this once our experiments are done and we have decided on our approach to preprocessing. @ArlindKadra has told us to have the preprocessing in the pipeline with hyperparameters. I can do this right now but it will hamper the current experiments we are doing so I have decided to wait.

autoPyTorch/data/tabular_feature_validator.py
ArlindKadra: They say they addressed it so it should be fine for the future.

autoPyTorch/data/tabular_feature_validator.py
ArlindKadra: The change in performance is insignificant, however, the code is better.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -478,33 +495,84 @@ def impute_nan_in_categories(self,
...
+def get_unused_category_symbol(
+    frame_column: pd.Series,
+    missing_value_symbol: Union[int, str],
+) -> Union[int, str]:
+    """
+    Select the appropriate missing value symbol for a column.
+
+    Giving a column from a DataFrame and an initial missing value symbol,
+    check if the missing_value is contained in the column, f it is, make
+    the necessary changes for a unique missing value symbol.
```
nabenabe0928, line 554:
```suggestion
    Given a column from a DataFrame and an initial missing value symbol,
    check if the missing_value is contained in the column.
    If it is, make a simple change to obtain a unique missing value symbol.
```

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -478,33 +495,84 @@ def impute_nan_in_categories(self,
...
+
+    Returns:
+        missing_value_symbol (Union[int, str]):
+            The unique missing value symbol.
+    """
+
+    if missing_value_symbol not in frame_column.cat.categories:
+        pass
+    elif isinstance(missing_value_symbol, str):
+        max_length = max(len(c) for c in frame_column.cat.categories)
```
nabenabe0928, line 570:
```suggestion
    elif isinstance(missing_value_symbol, str):
        max_length = max(len(c) for c in frame_column.cat.categories)
        # There are no categories that are longer than `max_length`
```

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -478,33 +495,84 @@ def impute_nan_in_categories(self,
...
+        missing_value_symbol += '0' * max_length
+    else:
+        # min_value is guaranteed to be negative since there exists -1 in categories
+        # and min_value must be smaller than -1. So the symbol is always negative.
+        min_value = min(c for c in frame_column.cat.categories)
+        missing_value_symbol = missing_value_symbol + min_value
```
nabenabe0928, line 576:
```suggestion
        # always missing_value_symbol + min_value < min_value < 0
        missing_value_symbol = missing_value_symbol + min_value
```

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -478,33 +495,84 @@ def impute_nan_in_categories(self,
...
+
+    Giving a column from a DataFrame and an initial missing value symbol,
+    check if the missing_value is contained in the column, f it is, make
+    the necessary changes for a unique missing value symbol.
+
+    Arguments:
+        frame_column (pd.Series):
+            The DataFrame column.
+        missing_value_symbol (Union[int, str]):
+            The initial symbol for the missing value.
```
nabenabe0928, line 560:
```suggestion
            The initial symbol for the missing value.
            -1 for int and '-1' for str. 
```

nabenabe0928 (commented):
@ArlindKadra Thanks for the PR. I added some comments as suggestions to help better understanding of the codes:)

nabenabe0928 (changes requested):
Hi, I found some bugs, so I will send `Request changes`.
There are two points that I addressed:
1. We cannot judge whether a whole column can be casted to number by just looking only the first value
2. The simple changes for the `get_unused_category_symbol` may suffer from overflow issues

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -478,33 +495,87 @@ def impute_nan_in_categories(self,
...
-                    self.dict_missing_value_per_col[column] = missing_value
+            # no missing values for categorical column
+            if not X[column].isna().any():
+                continue
+
+            if column not in self.dict_missing_value_per_col:
+
+                first_value = X[column].dropna().values[0]
+
+                if can_cast_as_number(first_value):
```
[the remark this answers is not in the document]
ArlindKadra: This is an additional edge case. The whole idea is for values to be either ints represented as string, or strings, this would be a mix of both. Going through the unique values and seeing if they can all be converted to floats is a smart idea. 

On an additional note, all of these values will be encoded, so, although I think imputing -1 and a string is nice, I also wonder to what use, when in the end it will be an additional column at least for the case of one-hot encoding.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -267,6 +269,15 @@ def _check_data(
         if hasattr(X, "iloc"):
             # If entered here, we have a pandas dataframe
             X = cast(pd.DataFrame, X)
+    
+            if hasattr(self, 'all_nan_columns') and set(self.all_nan_columns).issubset(X.columns):
```
ArlindKadra, line 273: Do we need to check if it is a subset?

ArlindKadra:
The code for deleting null columns will fail, since on the first call to `fit` you will have a list of nan_columns, however, on the call to transform for train, you will not have those columns there anymore and you will make the list empty. Then, while transforming the test dataset, you will get an error because of the DataFrames with different columns.

ravinkohli:
> The code for deleting null columns will fail, since on the first call to `fit` you will have a list of nan_columns, however, on the call to transform for train, you will not have those columns there anymore and you will make the list empty. Then, while transforming the test dataset, you will get an error because of the DataFrames with different columns.

Actually what is happening now is that we store the column order after we drop the all nan columns. So the column order is already only set after we have dropped the columns in fit. Therefore I don't think it will fail.

ArlindKadra:
> > The code for deleting null columns will fail, since on the first call to `fit` you will have a list of nan_columns, however, on the call to transform for train, you will not have those columns there anymore and you will make the list empty. Then, while transforming the test dataset, you will get an error because of the DataFrames with different columns.
> 
> Actually what is happening now is that we store the column order after we drop the all nan columns. So the column order is already only set after we have dropped the columns in fit. Therefore I don't think it will fail.

This is not an assumption, I noticed it failing and I debugged the trace.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -96,24 +99,31 @@ def _fit(
...
         if hasattr(X, "iloc") and not scipy.sparse.issparse(X):
-            X = cast(pd.DataFrame, X)
 
-            if not X.select_dtypes(include='object').empty:
-                X = self.infer_objects(X)
+            X = cast(pd.DataFrame, X)
+            self.all_nan_columns = set()
+            for column in X.columns:
+                if X[column].isna().all():
+                    self.all_nan_columns.add(column)
```
nabenabe0928, line 115:
```suggestion
             self.all_nan_columns = set([column for column in X.columns if X[column].isna().all()])
```

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -96,24 +99,31 @@ def _fit(
...
 
-            self._check_data(X)
             categorical_columns, numerical_columns, feat_type = self._get_columns_info(X)
 
             self.enc_columns = categorical_columns
-            if len(categorical_columns) >= 0:
-                X = self.impute_nan_in_categories(X)
```
nabenabe0928, line 112: Where are we imputing now?

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -142,21 +152,24 @@ def comparator(cmp1: str, cmp2: str) -> int:
 
             if len(categorical_columns) > 0:
                 self.categories = [
-                    # We fit an ordinal encoder, where all categorical
+                    # We fit a one-hot encoder, where all categorical
```
nabenabe0928, line 155: Uum, what do you mean by one-hot encoder here?
The corresponding index for each categorical in onehot encoding?
As far as I can see from this line, `list(range(len(cat)))` is not a one-hot encoder...

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -268,13 +284,13 @@ def _check_data(
             X = cast(pd.DataFrame, X)
 
             # Handle objects if possible
-            if not X.select_dtypes(include='object').empty:
+            object_columns_indicator = has_object_columns(X.dtypes.values)
+            if object_columns_indicator:
```
nabenabe0928, line 288:
```suggestion
            exist_object_columns = has_object_columns(X.dtypes.values)
            if exist_object_columns:
```

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -288,13 +304,20 @@ def _check_data(
...
+                            # we expect the dtypes to only be different if the column belongs
+                            # to all_nan_columns as these columns would be imputed. if there is
+                            # a value in the test set for a column in all_nan_columns, pandas
+                            # does not recognise the dtype of the test column properly
+                            raise ValueError("Changing the dtype of the features after fit() is "
+                                             "not supported. Fit() method was called with "
+                                             "{} whereas the new features have {} as type".format(self.dtypes,
+                                                                                                  dtypes,
+                                                                                                  )
+                                             )
```
nabenabe0928, line 320: What the comment out is saying is more informative to me.
At least, the error message should say `The dtype of some columns are different between training and test datasets.`.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -323,14 +348,17 @@ def _get_columns_info(
 
         # Make sure each column is a valid type
         for i, column in enumerate(X.columns):
```
nabenabe0928, line 350: Could you change this for loop block to the following?
The if statement should be coded in this way and I made the error message shorter by sharing a common part.
```suggestion
        for i, column in enumerate(X.columns):
            if self.all_nan_columns is not None and column in self.all_nan_columns:
                continue

            column_dtype = self.dtypes[i]
            err_msg = "Valid types are `numerical`, `categorical` or `boolean`, " \
                      "but input Column {} has an invalid type `{}`.".format(column, column_dtype)

            if column_dtype in ['category', 'bool']:
                categorical_columns.append(column)
                feat_type.append('categorical')
            # Move away from np.issubdtype as it causes
            # TypeError: data type not understood in certain pandas types
            elif is_numeric_dtype(column_dtype):
                feat_type.append('numerical')
                numerical_columns.append(column)
            elif column_dtype == 'object':
                # TODO verify how would this happen when we always convert the object dtypes to category
                raise ValueError(
                    "{} Cast it to a valid dtype before feeding it to AutoPyTorch. "
                    "You can cast it to a valid dtype using pandas.Series.astype."
                    "If you are working with string objects, the following "
                    "tutorial illustrates how to work with text data: "
                    "https://scikit-learn.org/stable/tutorial/text_analytics/working_with_text_data.html".format(
                        # noqa: E501
                        err_msg,
                    )
                )
            elif pd.core.dtypes.common.is_datetime_or_timedelta_dtype(column_dtype):
                raise ValueError(
                    "{} Convert the time information to a numerical value"
                    " before feeding it to AutoPyTorch. "
                    "One example of the conversion can be found on "
                    "https://stats.stackexchange.com/questions/311494/".format(err_msg)
                )
            else:
                raise ValueError(
                    "{} Make sure your data is formatted in a correct way"
                    "before feeding it to AutoPyTorch.".format(err_msg)
                )
```

nabenabe0928 (changes requested):
Hi, thanks for the commits. I was a bit surprised because the files changed a lot from the last time I saw!
I put some comments and modifications, so please address them:)

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: Adjusted

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: I have done so

PR #293, opened 2021-10-12
[refactor] Reduce code block sizes to avoid bugs

nabenabe0928:
Since there were too many unneeded nests, I resolved some of them to increase the readability.


ravinkohli:
I dont think we will be merging this to the intended branch.

PR #298, opened 2021-10-22
[FIX] Passing checks

ravinkohli:
This PR fixes the following 

-  tests which were failing due to the updated implementation in the branch.
-  In addition as certain components (preprocessors and network embedding) require a fix before tests can be run on them, I have currently skipped those tests.
- mypy and flake checks

The goal of this PR is to pass all checks for the reg_cocktails branch.

### DIFF
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -1429,6 +1429,7 @@ class BaseTask:
                              "Please call the `search()` method of {} prior to "
                              "fit_ensemble().".format(self.__class__.__name__))
 
+        precision = precision if precision is not None else self.precision
         if precision not in [16, 32, 64]:
             raise ValueError("precision must be one of 16, 32, 64 but got {}".format(precision))
 
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -1386,7 +1386,7 @@ class BaseTask:
         Args:
             optimize_metric (str): name of the metric that is used to
                 evaluate a pipeline. if not specified, value passed to search will be used
-            precision (int), (default=32): Numeric precision used when loading
+            precision (Optional[int]): Numeric precision used when loading
                 ensemble data. Can be either 16, 32 or 64.
             ensemble_nbest (Optional[int]):
                 only consider the ensemble_nbest models to build the ensemble.
@@ -1480,7 +1480,7 @@ class BaseTask:
         manager = self._init_ensemble_builder(
             time_left_for_ensembles=time_left_for_ensemble,
             optimize_metric=self.opt_metric if optimize_metric is None else optimize_metric,
-            precision=self.precision if precision is None else precision,
+            precision=precision,
             ensemble_size=ensemble_size,
             ensemble_nbest=ensemble_nbest,
         )
--- a/autoPyTorch/data/tabular_feature_validator.py
+++ b/autoPyTorch/data/tabular_feature_validator.py
@@ -345,7 +345,7 @@ class TabularFeatureValidator(BaseFeatureValidator):
                 and self.all_nan_columns is not None  # Ignore all_nan_columns is None
                 and len(set(X.columns[dtypes_diff]).difference(self.all_nan_columns)) != 0
             ):
-                diff_indices = np.where(np.array(dtypes_diff) is True)[0]
+                diff_indices = np.where(np.array(dtypes_diff))[0]
 
                 if not all([X[X.columns[i]].isna().all() for i in diff_indices]):
                     # The dtypes can be different if the column belongs
--- a/autoPyTorch/data/tabular_feature_validator.py
+++ b/autoPyTorch/data/tabular_feature_validator.py
@@ -247,7 +247,7 @@ class TabularFeatureValidator(BaseFeatureValidator):
         # having a value for a categorical column.
         # We need to convert the column in test data to
         # object otherwise the test column is interpreted as float
-        if self.column_transformer is not None: 
+        if self.column_transformer is not None:
             if len(self.categorical_columns) > 0:
                 categorical_columns = self.column_transformer.transformers_[0][-1]
                 for column in categorical_columns:
--- a/autoPyTorch/data/tabular_feature_validator.py
+++ b/autoPyTorch/data/tabular_feature_validator.py
@@ -247,11 +247,12 @@ class TabularFeatureValidator(BaseFeatureValidator):
         # having a value for a categorical column.
         # We need to convert the column in test data to
         # object otherwise the test column is interpreted as float
-        if self.column_transformer is not None and len(self.categorical_columns) > 0:
-            categorical_columns = self.column_transformer.transformers_[0][-1]
-            for column in categorical_columns:
-                if X[column].isna().all():
-                    X[column] = X[column].astype('object')
+        if self.column_transformer is not None: 
+            if len(self.categorical_columns) > 0:
+                categorical_columns = self.column_transformer.transformers_[0][-1]
+                for column in categorical_columns:
+                    if X[column].isna().all():
+                        X[column] = X[column].astype('object')
             X = self.column_transformer.transform(X)
 
         # Sparse related transformations
--- a/autoPyTorch/data/tabular_feature_validator.py
+++ b/autoPyTorch/data/tabular_feature_validator.py
@@ -518,7 +518,10 @@ class TabularFeatureValidator(BaseFeatureValidator):
             _ (bool): Whether the training and test datasets are consistent.
         """
         if self.all_nan_columns is None:
-            return True
+            if len(diff_cols) == 0:
+                return True
+            else:
+                return False
 
         # dtype is different ==> the column in at least either of train or test datasets must be all NaN
         # inconsistent <==> dtype is different and the col in both train and test is not all NaN
--- a/autoPyTorch/data/tabular_feature_validator.py
+++ b/autoPyTorch/data/tabular_feature_validator.py
@@ -100,6 +100,7 @@ class TabularFeatureValidator(BaseFeatureValidator):
         if cmp1 not in choices or cmp2 not in choices:
             raise ValueError('The comparator for the column order only accepts {}, '
                              'but got {} and {}'.format(choices, cmp1, cmp2))
+
         idx1, idx2 = choices.index(cmp1), choices.index(cmp2)
         return idx1 - idx2
 
@@ -246,13 +247,12 @@ class TabularFeatureValidator(BaseFeatureValidator):
         # having a value for a categorical column.
         # We need to convert the column in test data to
         # object otherwise the test column is interpreted as float
-        if len(self.categorical_columns) > 0:
-            categorical_columns = self.column_transformer.transformers_[0][-1]
-            for column in categorical_columns:
-                if X[column].isna().all():
-                    X[column] = X[column].astype('object')
-
         if self.column_transformer is not None:
+            if len(self.categorical_columns) > 0:
+                categorical_columns = self.column_transformer.transformers_[0][-1]
+                for column in categorical_columns:
+                    if X[column].isna().all():
+                        X[column] = X[column].astype('object')
             X = self.column_transformer.transform(X)
 
         # Sparse related transformations
@@ -345,11 +345,16 @@ class TabularFeatureValidator(BaseFeatureValidator):
                 and self.all_nan_columns is not None  # Ignore all_nan_columns is None
                 and len(set(X.columns[dtypes_diff]).difference(self.all_nan_columns)) != 0
             ):
-                # The dtypes can be different if and only if the column belongs
-                # to all_nan_columns as these columns would be imputed.
-                raise ValueError("The dtype of the features must not be changed after fit(), but"
-                                 " the dtypes of some columns are different between training ({}) and"
-                                 " test ({}) datasets.".format(self.dtypes, dtypes))
+                diff_indices = np.where(np.array(dtypes_diff) is True)[0]
+
+                if not all([X[X.columns[i]].isna().all() for i in diff_indices]):
+                    # The dtypes can be different if the column belongs
+                    # to all_nan_columns if the dataset is in training
+                    # set or if the column is all nan if the dataset
+                    # is for testing as these columns would be imputed.
+                    raise ValueError("The dtype of the features must not be changed after fit(), but"
+                                     " the dtypes of some columns are different between training ({}) and"
+                                     " test ({}) datasets.".format(self.dtypes, dtypes))
 
     def _get_columns_info(
         self,
--- a/autoPyTorch/data/tabular_feature_validator.py
+++ b/autoPyTorch/data/tabular_feature_validator.py
@@ -521,7 +521,7 @@ class TabularFeatureValidator(BaseFeatureValidator):
             if len(diff_cols) == 0:
                 return True
             else:
-                return False
+                return all(X[diff_cols].isna().all())
 
         # dtype is different ==> the column in at least either of train or test datasets must be all NaN
         # inconsistent <==> dtype is different and the col in both train and test is not all NaN
--- a/autoPyTorch/data/tabular_feature_validator.py
+++ b/autoPyTorch/data/tabular_feature_validator.py
@@ -247,12 +247,11 @@ class TabularFeatureValidator(BaseFeatureValidator):
         # having a value for a categorical column.
         # We need to convert the column in test data to
         # object otherwise the test column is interpreted as float
-        if self.column_transformer is not None:
-            if len(self.categorical_columns) > 0:
-                categorical_columns = self.column_transformer.transformers_[0][-1]
-                for column in categorical_columns:
-                    if X[column].isna().all():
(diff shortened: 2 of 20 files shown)


autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -246,13 +247,12 @@ def transform(
...
-            for column in categorical_columns:
-                if X[column].isna().all():
-                    X[column] = X[column].astype('object')
-
         if self.column_transformer is not None:
+            if len(self.categorical_columns) > 0:
+                categorical_columns = self.column_transformer.transformers_[0][-1]
+                for column in categorical_columns:
+                    if X[column].isna().all():
+                        X[column] = X[column].astype('object')
```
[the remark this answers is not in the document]
ravinkohli: no as `all_nan_columns` are the nan columns of the `X_train`, `transform` can be called also on test or val data

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -345,11 +345,16 @@ def _check_data(
...
                 and len(set(X.columns[dtypes_diff]).difference(self.all_nan_columns)) != 0
             ):
-                # The dtypes can be different if and only if the column belongs
-                # to all_nan_columns as these columns would be imputed.
-                raise ValueError("The dtype of the features must not be changed after fit(), but"
-                                 " the dtypes of some columns are different between training ({}) and"
-                                 " test ({}) datasets.".format(self.dtypes, dtypes))
+                diff_indices = np.where(np.array(dtypes_diff))[0]
+
+                if not all([X[X.columns[i]].isna().all() for i in diff_indices]):
```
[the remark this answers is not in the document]
ravinkohli: no as mentioned in the previous comment `all_nan_columns` are the nan columns of the `X_train`, `transform` can be called also on test or val data

autoPyTorch/pipeline/components/training/trainer/AdversarialTrainer.py
```diff
@@ -109,11 +109,7 @@ def train_step(self, data: np.ndarray, targets: np.ndarray) -> Tuple[float, torc
         loss = loss_func(self.criterion, original_outputs, adversarial_outputs)
         loss.backward()
         self.optimizer.step()
-        if self.scheduler:
-            if 'ReduceLROnPlateau' in self.scheduler.__class__.__name__:
-                self.scheduler.step(loss)
-            else:
-                self.scheduler.step()
```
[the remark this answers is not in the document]
ravinkohli: because for now we only update the step in `train_epoch` of the `BaseTrainerComponent`. These lines were a bug as for `AdversarialTrainer` the scheduler is updated every batch as well as epoch

autoPyTorch/data/tabular_feature_validator.py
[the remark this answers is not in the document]
nabenabe0928: Wait, I would like to check a few points:

First, the following understanding is correct?
Suppose some columns in train and test datasets have different dtypes,
1. if both train and test datasets have all NaN, it does not matter ==> both float
2. If one of train or test datasets have all NaN, it does not matter ==> one float and the other any
3. If both train and test datasets do not have all NaN, it might matter ==> for example, float in train and str in test because a user wrongly mapped training dataset, which does not include leading zero, and test dataset has some numbers that are leading zero.

Then, we should raise an error only when the point 3 happens, right?
I think we wrongly allow some of point 3 conditions and filter some of point 2 condition.

The precise condition here should be something like this, isn't it?:
```
    def is_datasets_consistent(self, diff_cols: List[str]) -> bool:
        """
        Check the consistency of dtypes between training and test datasets.

        Args:
            diff_cols (List[bool]): The column labels that have different dtypes.

        Returns:
            _ (bool): Whether the training and test datasets are consistent.
        """

        if self.all_nan_columns is None:
            return True

        # dtype is different ==> the column in at least either of train or test datasets must be all NaN
        # inconsistent ==> dtype is different and the col in both train and test is not all NaN
        inconsistent_cols = list(set(diff_cols) - self.all_nan_columns)
        
        return len(inconsistent_cols) == 0 or all(X[inconsistent_cols].isna().all())
```

nabenabe0928 (changes requested):

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -345,11 +344,16 @@ def _check_data(
...
+                    # The dtypes can be different if the column belongs
+                    # to all_nan_columns if the dataset is in training
+                    # set or if the column is all nan if the dataset
+                    # is for testing as these columns would be imputed.
+                    raise ValueError("The dtype of the features must not be changed after fit(), but"
+                                     " the dtypes of some columns are different between training ({}) and"
+                                     " test ({}) datasets.".format(self.dtypes, dtypes))
```
nabenabe0928, line 357:
```suggestion

    def _is_datasets_consistent(self, diff_cols: List[str]) -> bool:
        """
        Check the consistency of dtypes between training and test datasets.

        Args:
            diff_cols (List[bool]): The column labels that have different dtypes.

        Returns:
            _ (bool): Whether the training and test datasets are consistent.
        """

        if self.all_nan_columns is None:
            return True

        # dtype is different ==> the column in at least either of train or test datasets must be all NaN
        # inconsistent ==> dtype is different and the col in both train and test is not all NaN
        inconsistent_cols = list(set(diff_cols) - self.all_nan_columns)
        
        return len(inconsistent_cols) == 0 or all(X[inconsistent_cols].isna().all())
        
```

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -345,11 +344,16 @@ def _check_data(
                 and self.all_nan_columns is not None  # Ignore all_nan_columns is None
```
nabenabe0928, line 344: So the new version will be this:
```
            diff_cols = X.columns[[s_dtype != dtype for s_dtype, dtype in zip(self.dtypes, dtypes)]]
            if len(self.dtypes) == 0:
                self.dtypes = dtypes
            elif not self._is_datasets_consistent(diff_cols):
                # The dtypes can be different if the column belongs
                # to all_nan_columns if the dataset is in training
                # set or if the column is all nan if the dataset
                # is for testing as these columns would be imputed.
                raise ValueError("The dtype of the features must not be changed after fit(), but"
                                 " the dtypes of some columns are different between training ({}) and"
                                 " test ({}) datasets.".format(self.dtypes, dtypes))
```

test/test_data/test_feature_validator.py
```diff
@@ -237,9 +236,10 @@ def test_featurevalidator_categorical_nan(input_data_featuretest):
     validator.fit(input_data_featuretest)
     transformed_X = validator.transform(input_data_featuretest)
     assert any(pd.isna(input_data_featuretest))
-    assert any((-1 in categories) or ('-1' in categories) or ('Missing!' in categories) for categories in
-               validator.encoder.named_transformers_['encoder'].categories_)
-    assert np.shape(input_data_featuretest) == np.shape(transformed_X)
```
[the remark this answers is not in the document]
ravinkohli: because we are using a one hot encoder

autoPyTorch/data/tabular_feature_validator.py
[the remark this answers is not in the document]
ravinkohli: yes you are right. I'll change this.

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: okay it also makes it easier to understand. I will commit this.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -508,6 +506,23 @@ def infer_objects(self, X: pd.DataFrame) -> pd.DataFrame:
 
         return X
 
+    def _is_datasets_consistent(self, diff_cols: List[str], X: pd.DataFrame) -> bool:
+        """
+        Check the consistency of dtypes between training and test datasets.
+        Args:
+            diff_cols (List[bool]): The column labels that have different dtypes.
+        Returns:
+            _ (bool): Whether the training and test datasets are consistent.
```
nabenabe0928, line 515: Sorry I forgot adding `X`

```suggestion
        Check the consistency of dtypes between training and test datasets.

        Args:
            diff_cols (List[bool]): The column labels that have different types.
            X (pd.DataFrame): A validation or test dataset to be compared with the training dataset

        Returns:
            _ (bool): Whether the training and test datasets are consistent.
```

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -508,6 +506,23 @@ def infer_objects(self, X: pd.DataFrame) -> pd.DataFrame:
...
+        """
+        if self.all_nan_columns is None:
+            return True
+
+        # dtype is different ==> the column in at least either of train or test datasets must be all NaN
+        # inconsistent <==> dtype is different and the col in both train and test is not all NaN
+        inconsistent_cols = list(set(diff_cols) - self.all_nan_columns)
+
+        return len(inconsistent_cols) == 0 or all(X[inconsistent_cols].isna().all())
```
nabenabe0928, line 524: Which is better? You can take according to your preference.
```suggestion
        return len(inconsistent_cols) == 0 or X[inconsistent_cols].isna().all().all()
```

nabenabe0928 (commented):
Thanks for addressing my comments.
I noticed that my comments was not completed properly, so I added comments.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -337,16 +337,14 @@ def _check_data(
...
-                and self.all_nan_columns is not None  # Ignore all_nan_columns is None
-                and len(set(X.columns[dtypes_diff]).difference(self.all_nan_columns)) != 0
-            ):
-                # The dtypes can be different if and only if the column belongs
-                # to all_nan_columns as these columns would be imputed.
+            elif not self._is_datasets_consistent(diff_cols, X):
+                # The dtypes can be different if the column belongs
+                # to all_nan_columns if the dataset is in training
+                # set or if the column is all nan if the dataset
+                # is for testing as these columns would be imputed.
```
nabenabe0928, line 347: Probably, we can delete those comments or move them to the doc-string of `is_datasets_consistent`?

ravinkohli: yes I was thinking the same thing. But I wanted to wait for the tests to pass. I'll change this if they pass now.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -246,13 +247,12 @@ def transform(
         # having a value for a categorical column.
         # We need to convert the column in test data to
         # object otherwise the test column is interpreted as float
-        if len(self.categorical_columns) > 0:
-            categorical_columns = self.column_transformer.transformers_[0][-1]
-            for column in categorical_columns:
-                if X[column].isna().all():
-                    X[column] = X[column].astype('object')
-
         if self.column_transformer is not None:
```
ArlindKadra, line 250: Is this just to pass the tests, theoretically this is always instantiated during `_fit` right

ravinkohli: So, the lines moved into this condition that is to make columns 'object' dtype, require that the data `X` is a dataframe. If there is a column transformer,  we are sure that the data is a dataframe.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -508,6 +502,30 @@ def infer_objects(self, X: pd.DataFrame) -> pd.DataFrame:
...
+
+        Args:
+            diff_cols (List[bool]):
+                The column labels that have different dtypes.
+            X (pd.DataFrame):
+                A validation or test dataset to be compared with the training dataset
+        Returns:
+            _ (bool): Whether the training and test datasets are consistent.
+        """
+        if self.all_nan_columns is None:
```
ArlindKadra, line 520: when `self.all_nan_columns`, should you not check only for the second part at line 527?
it does not seem like it should return true, it seems like `inconsistent_cols` should be `diff_cols`.

ArlindKadra (commented):
Looks good, although I have a few uncertainties :)

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: we technically also support sparse data. So the test for sparse data were failing

autoPyTorch/pipeline/components/setup/network_backbone/InceptionTimeBackbone.py
```diff
@@ -78,7 +78,7 @@ def __init__(self, n_res_inputs: int, n_outputs: int):
     def forward(self, x: torch.Tensor, res: torch.Tensor) -> torch.Tensor:
         shortcut = self.shortcut(res)
         shortcut = self.bn(shortcut)
-        x += shortcut
+        x = x + shortcut
```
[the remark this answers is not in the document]
ravinkohli: no it was causing an error

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: oh thats true, let me fix that

autoPyTorch/api/base_task.py
```diff
@@ -1429,6 +1429,7 @@ def fit_ensemble(
                              "Please call the `search()` method of {} prior to "
                              "fit_ensemble().".format(self.__class__.__name__))
 
+        precision = precision if precision is not None else self.precision
```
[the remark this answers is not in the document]
ravinkohli: yeah true, I'll change this as well

autoPyTorch/pipeline/components/setup/network_backbone/InceptionTimeBackbone.py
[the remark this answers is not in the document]
ArlindKadra: surprising, why is there a different behavior.

autoPyTorch/data/tabular_feature_validator.py
ArlindKadra: Ah, then it is understandable.

autoPyTorch/data/tabular_feature_validator.py
ArlindKadra: I think this still does not fix it because it can be that `diff_cols` is not empty, but they are all `None`. So, the second `or `part of the return.

autoPyTorch/pipeline/components/setup/network_backbone/InceptionTimeBackbone.py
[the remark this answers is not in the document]
ravinkohli: I think its due to a new torch version

autoPyTorch/data/tabular_feature_validator.py
ArlindKadra:
```
 if self.all_nan_columns is None:
     if len(diff_cols) == 0:
         return True
     else:
         return False
```
If `self.nan_columns` is `None`, you check if there is a difference in the types by `len(diff_cols)`, if there is, you return `False`. Should you not also consider the case that there might be a difference in the types, however all of those columns might be NA fully for test?

`return len(inconsistent_cols) == 0 or all(X[inconsistent_cols].isna().all())` this is done by the second part of this or with 2 clauses. You are not reaching here and it might be that it is this case.

Maybe I am missunderstanding something?

test/test_data/test_feature_validator.py
```diff
@@ -237,9 +236,10 @@ def test_featurevalidator_categorical_nan(input_data_featuretest):
...
+    categories_ = validator.column_transformer.\
+        named_transformers_['categorical_pipeline'].named_steps['onehotencoder'].categories_
+    assert any(('0' in categories) or (0 in categories) or ('missing_value' in categories) for categories in
```
ArlindKadra, line 241: This are the unique category values for the one hot encoder right, what would be the case where they are 0, is it not 'missing_value' for categorical columns?

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: okay I got it, yeah it makes sense

test/test_data/test_feature_validator.py
ArlindKadra: according to this https://github.com/automl/Auto-PyTorch/blob/10aea660389c899e9ab16410e95371c7c97f2436/autoPyTorch/data/tabular_feature_validator.py#L380-L381

It will be a categorical column only when it has a `string` or `bool` dtype.
And if between test and train there are differences, it will either throw an error or convert a null column to object for test if not empty in train.

ArlindKadra: Does it fail if you run it without the `0` checks?

ravinkohli: > It will be a categorical column only when it has a `string` or `bool` dtype

not string, but category. And category can be string or int.

ravinkohli: It works fine without `'0'` check but removing `0` gives an error

Issue #301, opened 2021-10-25
[FIX] List of minor fixes to be made to the development branch


ravinkohli:
These are some of the minor fixes to be made to the development branch. 

- [ ] N_threads in API module. 
- [ ] Dropout Shape in ShapedResNetBackbone. [these](https://github.com/automl/Auto-PyTorch/blob/43d4639dc84e2657662586996f0002c98a331766/autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py#L48-L55) lines. 
- [ ] Fixes in shake drop and shake-shake utils. See [this](https://github.com/automl/Auto-PyTorch/blob/43d4639dc84e2657662586996f0002c98a331766/autoPyTorch/pipeline/components/setup/network_backbone/utils.py#L143) and [this](https://github.com/automl/Auto-PyTorch/blob/43d4639dc84e2657662586996f0002c98a331766/autoPyTorch/pipeline/components/setup/network_backbone/utils.py#L197-L201)
- [ ] Fixes in fully connected head config space. See [this](https://github.com/automl/Auto-PyTorch/blob/43d4639dc84e2657662586996f0002c98a331766/autoPyTorch/pipeline/components/setup/network_head/fully_connected.py#L62-L94) 
- [ ] Xaviar_normal in Xavier. [Here](https://github.com/automl/Auto-PyTorch/blob/43d4639dc84e2657662586996f0002c98a331766/autoPyTorch/pipeline/components/setup/network_initializer/XavierInit.py#L27)
- [ ] Add test loader in base_data_loader and add random state. [Here](https://github.com/automl/Auto-PyTorch/blob/43d4639dc84e2657662586996f0002c98a331766/autoPyTorch/pipeline/components/training/data_loader/base_data_loader.py#L134-L152)
- [ ] Predict_proba in pipeline tabular classification. [Here](https://github.com/automl/Auto-PyTorch/blob/43d4639dc84e2657662586996f0002c98a331766/autoPyTorch/pipeline/tabular_classification.py#L93-L105)

Issue #283, opened 2021-07-15
Wrong implementation of shake-drop


ArlindKadra:
https://github.com/automl/Auto-PyTorch/blob/900293795cd0f3766c39c87dcdb8c72a23d90659/autoPyTorch/pipeline/components/setup/network_backbone/utils.py#L121-L125

We noticed with @ravinkohli that the implementation is not right for shake-drop. The if conditions should be reversed, only during test it should be fixed.

PR #306, opened 2021-10-26
[FIX] Minor Fixes

ravinkohli:
Addresses issue #301 , and issue #283

### DIFF
--- a/autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
+++ b/autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
@@ -48,9 +48,13 @@ class ShapedResNetBackbone(ResNetBackbone):
                 self.config['resnet_shape'], 0, 0, 1000, self.config['num_groups'] + 1
             )[:-1]
 
-            dropout_shape = [
-                dropout / 1000 * self.config["max_dropout"] for dropout in dropout_shape
-            ]
+            dropout_shape = get_shaped_neuron_counts(
+                shape=self.config['dropout_shape'],
+                in_feat=0,
+                out_feat=0,
+                max_neurons=self.config["max_dropout"],
+                layer_count=self.config['num_groups'] + 1,
+            )[:-1]
 
             self.config.update(
                 {"dropout_%d" % (i + 1): dropout for i, dropout in enumerate(dropout_shape)}
--- a/autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
+++ b/autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
@@ -42,7 +42,7 @@ class ShapedResNetBackbone(ResNetBackbone):
         if self.config['use_dropout']:
             # the last dropout ("neuron") value is skipped since it will be equal
             # to output_feat, which is 0. This is also skipped when getting the
-            # nr of units for the architecture, since, it is mostly implemented for the
+            # n_units for the architecture, since, it is mostly implemented for the
             # output layer, which is part of the head and not of the backbone.
             dropout_shape = get_shaped_neuron_counts(
                 shape=self.config['resnet_shape'],
--- a/autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
+++ b/autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
@@ -39,17 +39,13 @@ class ShapedResNetBackbone(ResNetBackbone):
         self.config.update(
             {"num_units_%d" % (i): num for i, num in enumerate(neuron_counts)}
         )
-        # we are skipping the last layer, as the function get_shaped_neuron_counts
-        # is built for getting neuron counts, so it will add the out_features to
-        # the last layer. However, in dropout we dont want to have that, we just
-        # want to use the shape and not worry about the output.
         if self.config['use_dropout']:
+            # the last dropout ("neuron") value is skipped since it will be equal
+            # to output_feat, which is 0. This is also skipped when getting the
+            # nr of units for the architecture, since, it is mostly implemented for the
+            # output layer, which is part of the head and not of the backbone.
             dropout_shape = get_shaped_neuron_counts(
-                self.config['resnet_shape'], 0, 0, 1000, self.config['num_groups'] + 1
-            )[:-1]
-
-            dropout_shape = get_shaped_neuron_counts(
-                shape=self.config['dropout_shape'],
+                shape=self.config['resnet_shape'],
                 in_feat=0,
                 out_feat=0,
                 max_neurons=self.config["max_dropout"],
--- a/autoPyTorch/pipeline/components/setup/network_head/fully_connected.py
+++ b/autoPyTorch/pipeline/components/setup/network_head/fully_connected.py
@@ -59,23 +59,22 @@ class FullyConnectedHead(NetworkHeadComponent):
     ) -> ConfigurationSpace:
         cs = ConfigurationSpace()
 
-        num_layers_is_constant = True if len(num_layers.value_range) == 1 else False
         min_num_layers: int = num_layers.value_range[0]  # type: ignore
         max_num_layers: int = num_layers.value_range[-1]  # type: ignore
+        num_layers_is_constant = (min_num_layers == max_num_layers)
 
         num_layers_hp = get_hyperparameter(num_layers, UniformIntegerHyperparameter)
         activation_hp = get_hyperparameter(activation, CategoricalHyperparameter)
         cs.add_hyperparameter(num_layers_hp)
 
-        if num_layers_is_constant:
-            # only add activation if we have more than 1 layer
-            if max_num_layers > 1:
-                cs.add_hyperparameter(activation_hp)
-        else:
+        if not num_layers_is_constant:
             cs.add_hyperparameter(activation_hp)
             cs.add_condition(CS.GreaterThanCondition(activation_hp, num_layers_hp, 1))
+        elif max_num_layers > 1:
+            # only add activation if we have more than 1 layer
+            cs.add_hyperparameter(activation_hp)
 
-        for i in range(1, max_num_layers + 1):
+        for i in range(min_num_layers, max_num_layers + 1):
             num_units_search_space = HyperparameterSearchSpace(
                 hyperparameter=f"units_layer_{i}",
                 value_range=units_layer.value_range,
--- a/autoPyTorch/pipeline/components/setup/network_head/fully_connected.py
+++ b/autoPyTorch/pipeline/components/setup/network_head/fully_connected.py
@@ -59,23 +59,38 @@ class FullyConnectedHead(NetworkHeadComponent):
     ) -> ConfigurationSpace:
         cs = ConfigurationSpace()
 
-        min_num_layers, max_num_layers = num_layers.value_range
+        num_layers_is_constant = True if len(num_layers.value_range) == 1 else False
+        min_num_layers: int = num_layers.value_range[0]  # type: ignore
+        max_num_layers: int = num_layers.value_range[-1]  # type: ignore
+
         num_layers_hp = get_hyperparameter(num_layers, UniformIntegerHyperparameter)
-
         activation_hp = get_hyperparameter(activation, CategoricalHyperparameter)
+        cs.add_hyperparameter(num_layers_hp)
 
-        cs.add_hyperparameters([num_layers_hp, activation_hp])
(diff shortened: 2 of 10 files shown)


autoPyTorch/pipeline/tabular_classification.py
```diff
@@ -101,13 +100,8 @@ def _predict_proba(self, X: np.ndarray) -> np.ndarray:
...
-            proba = pred[:, :self.dataset_properties['output_shape']]
-            normalizer = proba.sum(axis=1)[:, np.newaxis]
-            normalizer[normalizer == 0.0] = 1.0
-            proba /= normalizer
-
-            return proba
-
+            # The final layer is always softmax now (`pred` already gives pseudo proba)
+            return pred
         else:
```
nabenabe0928, line 105: We do not change this else statement in this PR or did you forget?

nabenabe0928 (commented):
Thanks for the PR. I will add other comments soon.

autoPyTorch/pipeline/tabular_classification.py
ravinkohli: why dont we need this else? I know it was not there in the reg cocktails but here we are raising an error for the user to notify that we don't support other output shapes. I think its fine or even necessary

autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
```diff
@@ -39,19 +39,19 @@ def build_backbone(self, input_shape: Tuple[int, ...]) -> torch.nn.Sequential:
...
+            # the last dropout ("neuron") value is skipped since it will be equal
+            # to output_feat, which is 0. This is also skipped when getting the
+            # nr of units for the architecture, since, it is mostly implemented for the
+            # output layer, which is part of the head and not of the backbone.
             dropout_shape = get_shaped_neuron_counts(
-                self.config['resnet_shape'], 0, 0, 1000, self.config['num_groups'] + 1
+                shape=self.config['resnet_shape'],
+                in_feat=0,
+                out_feat=0,
+                max_neurons=self.config["max_dropout"],
```
nabenabe0928, line 51: `int` or `float`?

autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
```diff
@@ -39,19 +39,19 @@ def build_backbone(self, input_shape: Tuple[int, ...]) -> torch.nn.Sequential:
...
             {"num_units_%d" % (i): num for i, num in enumerate(neuron_counts)}
         )
-        # we are skipping the last layer, as the function get_shaped_neuron_counts
-        # is built for getting neuron counts, so it will add the out_features to
-        # the last layer. However, in dropout we dont want to have that, we just
-        # want to use the shape and not worry about the output.
         if self.config['use_dropout']:
+            # the last dropout ("neuron") value is skipped since it will be equal
+            # to output_feat, which is 0. This is also skipped when getting the
+            # nr of units for the architecture, since, it is mostly implemented for the
```
nabenabe0928, line 45:
```suggestion
            # n_units for the architecture, since, it is mostly implemented for the
```

autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
```diff
@@ -39,19 +39,19 @@ def build_backbone(self, input_shape: Tuple[int, ...]) -> torch.nn.Sequential:
         self.config.update(
             {"num_units_%d" % (i): num for i, num in enumerate(neuron_counts)}
         )
-        # we are skipping the last layer, as the function get_shaped_neuron_counts
-        # is built for getting neuron counts, so it will add the out_features to
-        # the last layer. However, in dropout we dont want to have that, we just
-        # want to use the shape and not worry about the output.
         if self.config['use_dropout']:
+            # the last dropout ("neuron") value is skipped since it will be equal
+            # to output_feat, which is 0. This is also skipped when getting the
```
nabenabe0928, line 44: The dropout rate at the last layer is skipped because it must be 0?
I did not get the comment here.

autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
```diff
@@ -39,19 +39,19 @@ def build_backbone(self, input_shape: Tuple[int, ...]) -> torch.nn.Sequential:
...
+            # nr of units for the architecture, since, it is mostly implemented for the
+            # output layer, which is part of the head and not of the backbone.
```
nabenabe0928, line 46: What does the "head" in "output layer is part of the head" mean?

autoPyTorch/pipeline/components/setup/network_head/fully_connected.py
```diff
@@ -59,23 +59,38 @@ def get_hyperparameter_search_space(
     ) -> ConfigurationSpace:
         cs = ConfigurationSpace()
 
-        min_num_layers, max_num_layers = num_layers.value_range
-        num_layers_hp = get_hyperparameter(num_layers, UniformIntegerHyperparameter)
+        num_layers_is_constant = True if len(num_layers.value_range) == 1 else False
+        min_num_layers: int = num_layers.value_range[0]  # type: ignore
+        max_num_layers: int = num_layers.value_range[-1]  # type: ignore
```
nabenabe0928, line 64:
```suggestion
        min_num_layers: int = num_layers.value_range[0]  # type: ignore
        max_num_layers: int = num_layers.value_range[-1]  # type: ignore
        num_layers_is_constant = (min_num_layers == max_num_layers)
```

autoPyTorch/pipeline/components/setup/network_head/fully_connected.py
```diff
@@ -59,23 +59,38 @@ def get_hyperparameter_search_space(
...
 
-        cs.add_hyperparameters([num_layers_hp, activation_hp])
-        cs.add_condition(CS.GreaterThanCondition(activation_hp, num_layers_hp, 1))
+        if num_layers_is_constant:
+            # only add activation if we have more than 1 layer
+            if max_num_layers > 1:
+                cs.add_hyperparameter(activation_hp)
+        else:
+            cs.add_hyperparameter(activation_hp)
+            cs.add_condition(CS.GreaterThanCondition(activation_hp, num_layers_hp, 1))
```
nabenabe0928, line 76:
```suggestion
        if not num_layers_is_constant:
            cs.add_hyperparameter(activation_hp)
            cs.add_condition(CS.GreaterThanCondition(activation_hp, num_layers_hp, 1))
        elif max_num_layers > 1:
            # only add activation if we have more than 1 layer
            cs.add_hyperparameter(activation_hp)
```

autoPyTorch/pipeline/components/setup/network_head/fully_connected.py
```diff
@@ -59,23 +59,38 @@ def get_hyperparameter_search_space(
...
 
-        for i in range(1, int(max_num_layers)):
-            num_units_search_space = HyperparameterSearchSpace(hyperparameter=f"units_layer_{i}",
-                                                               value_range=units_layer.value_range,
-                                                               default_value=units_layer.default_value,
-                                                               log=units_layer.log)
+        for i in range(1, max_num_layers + 1):
```
nabenabe0928, line 78:
```suggestion
        for i in range(min_num_layers, max_num_layers + 1):
```

autoPyTorch/pipeline/components/setup/network_head/fully_connected.py
```diff
@@ -59,23 +59,38 @@ def get_hyperparameter_search_space(
...
-            if i >= int(min_num_layers):
-                cs.add_condition(CS.GreaterThanCondition(num_units_hp, num_layers_hp, i))
+            if i >= min_num_layers:
+                cs.add_hyperparameter(num_units_hp)
+                # In the case of a constant, the max and min number of layers are the same.
+                # So no condition is needed. If it is not a constant but a hyperparameter,
+                # then a condition has to be made so that it accounts for the value of the
+                # hyperparameter.
+                if not num_layers_is_constant:
+                    cs.add_condition(CS.GreaterThanCondition(num_units_hp, num_layers_hp, i))
```
nabenabe0928, line 94:
```suggestion
            cs.add_hyperparameter(num_units_hp)
            # (1) n_layers is const ==> n_units for each layer is always required (not conditional)
            # (2) n_layers is not const ==> n_units for deeper layers is required only when
            #     the corresponding layers are appended in the architecture. (conditional)
            if not num_layers_is_constant:
                cs.add_condition(CS.GreaterThanCondition(num_units_hp, num_layers_hp, i))
```

nabenabe0928 (changes requested):
I left some comments.

autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
ravinkohli: both

autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
ravinkohli: network_head

autoPyTorch/pipeline/components/setup/network_head/fully_connected.py
ravinkohli: why would you like to remove this `if i >= min_num_layers:` condition?

nabenabe0928: Good point!
Thanks to your comment, I checked this point again.
So the last version before addressing my comment is [here](https://github.com/ravinkohli/Auto-PyTorch/blob/040d26214bdd907b8ea2db3603df19574b51d2a0/autoPyTorch/pipeline/components/setup/network_head/fully_connected.py#L78-L94).

The question is:
1. Do we set `num_units_hp` for all the layers or just from `min_num_layers` to `max_num_layers`?
2. If we do not set `num_units_hp` from `1` to `min_num_layers - 1`, why do we iterate from `1`?

I thought that we do not set this hyperparameter for lower layers, but if I think through more, it is a bit strange to not set this hyperparameter from the `1` to `min_num_layers - 1`.
Means shouldn't we have something like this?
```
        for i in range(1, max_num_layers + 1):
            num_units_search_space = HyperparameterSearchSpace(
                hyperparameter=f"units_layer_{i}",
                value_range=units_layer.value_range,
                default_value=units_layer.default_value,
                log=units_layer.log,
            )
            num_units_hp = get_hyperparameter(num_units_search_space, UniformIntegerHyperparameter)
            cs.add_hyperparameter(num_units_hp)

            if i >= min_num_layers and not num_layers_is_constant:
                # In the case of a constant, the max and min number of layers are the same.
                # So no condition is needed. If it is not a constant but a hyperparameter,
                # then a condition has to be made so that it accounts for the value of the
                # hyperparameter.
                cs.add_condition(CS.GreaterThanCondition(num_units_hp, num_layers_hp, i))
```

autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
nabenabe0928: `max_neurons` relates to `max_droput`?

ravinkohli: Its a hacky way of using the same function to get shaped neurons and shaped dropout. They do not relate to max_neurons but we use the same logic for getting a number of neurons or the value for dropout for a shape. It's not ideal, and we should fix it later.

autoPyTorch/pipeline/components/setup/network_backbone/ShapedResNetBackbone.py
ravinkohli: yeah so this relates to the point I made in the previous comment, for a shape of neurons we also get the output layer, here the out_feat=0 and we ignore that.

autoPyTorch/pipeline/components/setup/network_head/fully_connected.py
ravinkohli: we set num_units for all the layers, 
we iterate from 1 as we count from 1 for the layers so the first layer has num_units_1 rather than num_units_0.

Yeah I agree with the code you have presented. Thanks for pointing this out, I think this was not correct before, we should search for all the units and actually I think the current code will also fail if we have min_num_layers >1 as the `self.config` wont have units for the layers less than the min. I'll fix this.

nabenabe0928 (approved):

PR #308, opened 2021-10-26
[style] Modify tabular_target_validator.py

nabenabe0928:
Modified typing.


ravinkohli:
I dont think we will be getting back to this PR as we are no longer using `refactor_development_regularization_cocktails`.

PR #310, opened 2021-11-04
[version] Update the automl_common to the latest

nabenabe0928:
Update the automl_common to the latest version in the `autoPytorch` branch of `automl_common` repository.


ravinkohli:
I think we should wait for all the PRs to be merged to automl_common and then update automl_common

PR #311, opened 2021-11-04
[refactor] Update the all nan searching by a quicker method

nabenabe0928:
Since the `isna().all()` can be re-write in a concise way, I made a change and created this PR.

### DIFF
--- a/autoPyTorch/data/tabular_feature_validator.py
+++ b/autoPyTorch/data/tabular_feature_validator.py
@@ -249,11 +249,11 @@ class TabularFeatureValidator(BaseFeatureValidator):
         if len(self.categorical_columns) > 0:
             categorical_columns = self.column_transformer.transformers_[0][-1]
 
-            all_nan_cat_cols = X[categorical_columns].isna().all()
-            # filter all nan is True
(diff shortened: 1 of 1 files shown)


autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -247,10 +256,12 @@ def transform(
         # We need to convert the column in test data to
         # object otherwise the test column is interpreted as float
         if len(self.categorical_columns) > 0:
-            categorical_columns = self.column_transformer.transformers_[0][-1]
-            for column in categorical_columns:
-                if X[column].isna().all():
-                    X[column] = X[column].astype('object')
+            assert hasattr(self, 'enc_columns') and self.enc_columns is not None  # MyPy
```
ravinkohli, line 259: we have changed this to `transformed_columns`, I think we should wait for PR #306 to be merged before merging this one as well.

ravinkohli:
Thanks for this PR there are some great changes suggested. But I think we have already merged some of the changes suggested in this PR. Also, we are now using the `reg_cocktails` branch. So, I think we won't be proceeding with this pull request. But if you think any important change is still missing from that branch, please create a new PR.

Issue #231, opened 2021-05-21
Docker image


ravinkohli:
We should make a docker image for running autopytorch on github

PR #314, opened 2021-11-10
[Add] Dockerfile

ravinkohli:
This resolves issue #231. But it is an incomplete implementation. The rest has been raised as another issue #315.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,40 @@
+FROM ubuntu:20.04
+
+WORKDIR /auto-pytorch
+
+# install linux packages
+RUN apt-get update
+
+# Set the locale
+# workaround for https://github.com/automl/auto-sklearn/issues/867
+RUN apt-get -y install locales
+RUN touch /usr/share/locale/locale.alias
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM ubuntu:20.04
```
nabenabe0928, line 1: I think we should use any one of the docker images from [here](https://hub.docker.com/r/nvidia/cuda) because some codes are assuming that we can use torch on CUDA. FYI, basically people use `xxx-devel-xxx`.

[Additional reference here](https://stackoverflow.com/questions/65492490/how-to-install-cuda-enabled-pytorch-in-a-docker-container)

ravinkohli: It's fine to use this ubuntu container for the dockerfile that we upload to the docker hub. I think the most general use case is to use CPU. If someone would like to use it with CUDA, it is quite easy for them to edit this file.

nabenabe0928: Btw, i don't think pynisher works properly when we use CUDA, but have you ever tried any runs on CUDA?

ravinkohli: Actually thats a good point. I'll create an issue and a subsequent PR where we can use docker to run all our tests on cuda. What do you think?

nabenabe0928: Sounds great:)

PR #317, opened 2021-11-11
[ADD] Missing Batchnorm

### LINKED ISSUE #297: Batch normalization missing
I noticed that batch normalization is missing in the `MLP` and `ShapedMLP` backbone, we should probably add it to both implementations.

ravinkohli:
This PR fixes #297


nabenabe0928 (approved):
I do not see problems now.

PR #321, opened 2021-11-11
[FIX] remove .pth for early stopping

### LINKED ISSUE #110: Remove the temporal .pth files left by early stopping

ravinkohli:
Fixes #110

### DIFF
--- a/autoPyTorch/pipeline/components/training/trainer/__init__.py
+++ b/autoPyTorch/pipeline/components/training/trainer/__init__.py
@@ -354,13 +354,7 @@ class TrainerChoice(autoPyTorchChoice):
 
         # As training have finished, load the best weight
         if self.checkpoint_dir is not None:
-            best_path = os.path.join(self.checkpoint_dir, 'best.pth')
-            self.logger.debug(f" Early stopped model {X['num_run']} on epoch {self.run_summary.get_best_epoch()}")
-            # We will stop the training. Load the last best performing weights
-            X['network'].load_state_dict(torch.load(best_path))
-
-            # Clean the temp dir
-            shutil.rmtree(self.checkpoint_dir)
+            self._load_best_weights_and_clean_checkpoints(X)
 
         self.logger.info(f"Finished training with {self.run_summary.repr_last_epoch()}")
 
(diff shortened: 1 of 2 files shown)


test/test_pipeline/components/training/test_training.py
```diff
@@ -13,6 +16,7 @@
 import torch
 
 from autoPyTorch import constants
+from autoPyTorch.pipeline.components.setup.lr_scheduler.constants import StepIntervalUnit
```
ravinkohli, line 19:
```suggestion
```

autoPyTorch/pipeline/components/training/trainer/__init__.py
```diff
@@ -351,6 +352,16 @@ def _fit(self, X: Dict[str, Any], y: Any = None, **kwargs: Any) -> 'TrainerChoic
...
 
+        # As training have finished, load the best weight
+        if self.checkpoint_dir is not None:
+            best_path = os.path.join(self.checkpoint_dir, 'best.pth')
+            self.logger.debug(f" Early stopped model {X['num_run']} on epoch {self.run_summary.get_best_epoch()}")
+            # We will stop the training. Load the last best performing weights
+            X['network'].load_state_dict(torch.load(best_path))
+
+            # Clean the temp dir
+            shutil.rmtree(self.checkpoint_dir)
```
nabenabe0928, line 363: Probably we need to refactor this file in the future.
But just for the start, could you make this method?
Just in case, Dict is passed by reference, so it will not be a problem. 
```
        # As training have finished, load the best weight
        def _load_best_weights_and_clean_checkpoints(self, X: Dict[str, Any]) -> None:
            """
            Load the best model until the last epoch and delete all the files for checkpoints.

            Args:
                X (Dict[str, Any]): Dependencies needed by current component to perform fit
           """
            best_path = os.path.join(self.checkpoint_dir, 'best.pth')
            self.logger.debug(f" Early stopped model {X['num_run']} on epoch {self.run_summary.get_best_epoch()}")
            # We will stop the training. Load the last best performing weights
            X['network'].load_state_dict(torch.load(best_path))

            # Clean the temp dir
            shutil.rmtree(self.checkpoint_dir)
```

Issue #318, opened 2021-11-11
[Add] Add information about parallel computation


ravinkohli:
We should also add information about parallel computation in the manual as auto-sklearn has done [here](https://automl.github.io/auto-sklearn/master/manual.html#parallel-computation)

PR #322, opened 2021-11-11
[Add] documentation and example for parallel computation

ravinkohli:
Addresses #318


examples/40_advanced/example_parallel_n_jobs.py
```diff
@@ -0,0 +1,68 @@
+"""
+======================
+Tabular Classification
+======================
+
+The following example shows how to fit a sample classification model
```
ravinkohli, line 6:
```suggestion
The following example shows how to fit a sample classification model parallely on 2 cores
```

Issue #330, opened 2021-11-16
[feat] Add functionality to show incumbent performance over time


nabenabe0928:
Since the current auto pytorch does not support the functionality to visualize the performance over time, this issue is a request to add such functionality.

This issue is inspired by the auto sklearn ([doc](https://automl.github.io/auto-sklearn/master/examples/40_advanced/example_pandas_train_test.html#sphx-glr-examples-40-advanced-example-pandas-train-test-py), [code](https://github.com/automl/Auto-PyTorch/blob/development/examples/40_advanced/example_visualization.py)).

The basic strategy is to encapsulate an object that has functionalities to plot in the `BaseTask`.

PR #331, opened 2021-11-16
[feat] Add an object that realizes the perf over time viz

nabenabe0928:
This PR addresses the [issue#330](https://github.com/automl/Auto-PyTorch/issues/330).

Milestones
1. Add tests/debug
2. Encapsulate in BaseTask


nabenabe0928:
A toy running example:

```
import warnings

import numpy as np
import pandas as pd

from sklearn import model_selection

from autoPyTorch.api.tabular_classification import TabularClassificationTask


warnings.simplefilter(action='ignore', category=UserWarning)
warnings.simplefilter(action='ignore', category=FutureWarning)


n_samples, dim = 50, 2
X = np.random.random((n_samples, dim)) * 2 - 1
y = ((X ** 2).sum(axis=-1) < 2 / np.pi).astype(np.int32)
print(y)

X, y = pd.DataFrame(X), pd.DataFrame(y)
X_train, X_test, y_train, y_test = model_selection.train_test_split(X, y)

api = TabularClassificationTask(seed=42)

api.search(X_train=X_train, y_train=y_train, X_test=X_test, y_test=y_test, optimize_metric='accuracy',
           total_walltime_limit=300, func_eval_time_limit_secs=20)

metric_name = 'accuracy'

api.plot_perf_over_time(
    metric_name=metric_name,
    show=True,
    marker='*',
    markersize=10
)
```

nabenabe0928:
@ravinkohli 
I would like you to check if it works on a cluster.
Because matplotlib usually causes some problems on the usage of clusters.

In such case, we need to make sure that it works universally using `Agg` or `TkAgg` in the matplotlib setting.

eddiebergman (approved):
I'm now getting updates for AutoPytorch and I'll review them just to get a better idea of what's going on inside them.

... but yes this commit changing status type to fixed constants seems good

autoPyTorch/api/base_task.py
```diff
@@ -190,6 +194,8 @@ def __init__(
         self.pipeline_options = replace_string_bool_to_bool(json.load(open(
             os.path.join(os.path.dirname(__file__), '../configs/default_pipeline_options.json'))))
 
+        self._visualizer = ResultsVisualizer()
```
ravinkohli, line 197: Why do we need it to be an attribute of self? Do you expect to reuse an already created instance of this class?  Unlike `results_manager`, this class does not hold any data. Also, I don't expect a user to call `plot_perf_over_time` more than once with the same parameters. So I suggest you recreate an instance whenever the `plot_perf_over_time` function is called.

autoPyTorch/api/base_task.py
```diff
@@ -1479,3 +1485,65 @@ def sprint_statistics(self) -> str:
...
+        metric_name: str,
+        ax: Optional[plt.Axes] = None,
+        plot_setting_params: PlotSettingParams = PlotSettingParams(),
+        color_label_settings: ColorLabelSettings = ColorLabelSettings(),
+        *args: Any,
+        **kwargs: Any
+    ) -> None:
+        """
+        Visualize the performance over time using matplotlib.
+        The plot relate arguments are based on matplotlib.
```
ravinkohli, line 1500:
```suggestion
        The plot related arguments are based on matplotlib.
```

autoPyTorch/api/base_task.py
```diff
@@ -1479,3 +1485,65 @@ def sprint_statistics(self) -> str:
...
+        for key, color_label in vars(color_label_settings).items():
+            if color_label is None:
+                continue
+
+            prefix = '::'.join(key.split('_'))
+            try:
+                new_key = [key for key in results.data.keys() if key.startswith(prefix)][0]
+                colors[new_key], labels[new_key] = color_label
+            except IndexError:  # ensemble does not always have results
+                pass
```
ravinkohli, line 1543: Do you think these lines can also be handled inside the visualiser? It would be consistent with the use of `plot_setting_params`.

autoPyTorch/api/results_visualizer.py
```diff
@@ -0,0 +1,273 @@
...
+    worst_val: float
+) -> Tuple[np.ndarray, np.ndarray]:
+    """
+    Get the performance and time step to plot.
+
+    Args:
+        results (np.ndarray):
+            The cumulated performance per evaluation.
+        cum_times (np.ndarray):
+            The cumulated runtime at each end of evaluations.
```
ravinkohli, line 104: Do you mean this? 
```suggestion
            The cumulated runtime at the end of each evaluation.
```

autoPyTorch/api/results_visualizer.py
```diff
@@ -0,0 +1,273 @@
...
+        results: np.ndarray,
+        worst_val: float,
+        plot_setting_params: PlotSettingParams,
+        label: Optional[str] = None,
+        color: Optional[str] = None,
+        *args: Any,
+        **kwargs: Any
+    ) -> None:
+        """
+        Plot the performance of the AutoPytorch over time.
```
ravinkohli, line 195:
```suggestion
        Plot the incumbent performance of AutoPytorch over time.
```

autoPyTorch/api/results_visualizer.py
```diff
@@ -0,0 +1,273 @@
...
+        results: MetricResults,
+        plot_setting_params: PlotSettingParams,
+        colors: Dict[str, Optional[str]],
+        labels: Dict[str, Optional[str]],
+        ax: Optional[plt.Axes] = None,
+        *args: Any,
+        **kwargs: Any
+    ) -> None:
+        """
+        Plot the performance of the AutoPytorch over time.
```
ravinkohli, line 235:
```suggestion
        Plot the incumbent performance of AutoPytorch over time.
```

autoPyTorch/api/results_visualizer.py
ravinkohli: Actually, could you also add details on the difference between this and the `plot_perf_over_time`?

ravinkohli (changes requested):
Thanks for this PR. I have left some comments. I think we should move `results_manager` and `results_visualizer` to the utils submodule. It's better to keep the API folder only for the APIs for the tasks. Overall, I think there is a need for tests to improve coverage but I have already seen this as a TODO so just to remind you about it. Also, we should update the `example_visualisation.py` with how to use this functionality.

ravinkohli:
> @ravinkohli I would like you to check if it works on a cluster. Because matplotlib usually causes some problems on the usage of clusters.
> 
> In such case, we need to make sure that it works universally using `Agg` or `TkAgg` in the matplotlib setting.



> @ravinkohli I would like you to check if it works on a cluster. Because matplotlib usually causes some problems on the usage of clusters.
> 
> In such case, we need to make sure that it works universally using `Agg` or `TkAgg` in the matplotlib setting.

Sure, I'll do so today or tomorrow.

autoPyTorch/api/base_task.py
eddiebergman: +1

eddiebergman:
@ravinkohli Just a question for later considerations, nothing of the `ResultsVisualizer` seems to be `auto-pytorch` specific, could we eventually factor this out into `automl_common`?

ravinkohli:
> @ravinkohli Just a question for later considerations, nothing of the `ResultsVisualizer` seems to be `auto-pytorch` specific, could we eventually factor this out into `automl_common`? Not to be done now, just that `automl_commons` having visualization features would be nice and I could look at this for implementation.

Hey, thanks for your comment. Yes, I think this is something we should aim to have in `automl_common`.

autoPyTorch/api/results_visualizer.py
nabenabe0928: exactly, thanks for pointing out:)

autoPyTorch/api/results_visualizer.py
nabenabe0928: This is an implicit method, so it is supposed to be referred only inside the class. In other words, whenever people see methods with preceding `_`, such methods should not be called from outside.

It is also in the autopep8 guideline:)

_single_leading_underscore: weak "internal use" indicator. E.g. from M import * does not import objects whose names start with an underscore.

Ref: [autopep8 Descriptive Naming Styles](https://www.python.org/dev/peps/pep-0008/#id34)

nabenabe0928:
> > @ravinkohli Just a question for later considerations, nothing of the `ResultsVisualizer` seems to be `auto-pytorch` specific, could we eventually factor this out into `automl_common`? Not to be done now, just that `automl_commons` having visualization features would be nice and I could look at this for implementation.
> 
> Hey, thanks for your comment. Yes, I think this is something we should aim to have in `automl_common`.

Yeah, since it is separated objects, we can easily encapsulate in each package:)

autoPyTorch/api/results_visualizer.py
ravinkohli: Sure, but I meant for better documentation for the users, we can let them know that this is plotting performance for an individual step.

nabenabe0928:
> Thanks for this PR. I have left some comments. I think we should move `results_manager` and `results_visualizer` to the utils submodule. It's better to keep the API folder only for the APIs for the tasks. Overall, I think there is a need for tests to improve coverage but I have already seen this as a TODO so just to remind you about it. Also, we should update the `example_visualisation.py` with how to use this functionality.

So I addressed your comments and since changing file names loses the change history in the latest commit (of course you can see the diff in the older commits though), I will change the file names after all the reviews finish.

autoPyTorch/api/results_manager.py
```diff
@@ -55,42 +82,127 @@ def _extract_metrics_info(
...
     return {
         metric.name: cost2metric(cost=cost_info[metric.name], metric=metric)
-        if metric.name in avail_metrics else np.nan
+        if metric.name in avail_metrics else metric._worst_possible_result
         for metric in scoring_functions
     }
 
 
+class EnsembleResults:
+    def __init__(
```
ravinkohli, line 98: Could you add documentation for this class?

autoPyTorch/api/results_manager.py
```diff
@@ -55,42 +82,127 @@ def _extract_metrics_info(
...
+        self._train_scores = self.train_scores[order].tolist()
+        self._test_scores = self.test_scores[order].tolist()
+        self._end_times = self.end_times[order].tolist()
+
+    def _extract_results_from_ensemble_performance_history(
+        self,
+        ensemble_performance_history: List[Dict[str, Any]]
+    ) -> None:
+        """
+        Extract the information to match this class format.
```
ravinkohli, line 153:
```suggestion
        Extract information from `ensemble_performance_history` to match the format of this class.
```

autoPyTorch/api/results_manager.py
```diff
@@ -55,42 +82,127 @@ def _extract_metrics_info(
...
+
+        Args:
+            ensemble_performance_history (List[Dict[str, Any]]):
+                The history of the ensemble optimization from SMAC.
```
ravinkohli, line 157: Ensemble optimization is not happening in SMAC. This information comes from [here](https://github.com/automl/Auto-PyTorch/blob/ef8d21ae982883548a8e77a3690eefa810d3ae07/autoPyTorch/ensemble/ensemble_builder.py#L856) in `EnsembleBuilder`.

autoPyTorch/api/results_manager.py
```diff
@@ -55,42 +82,127 @@ def _extract_metrics_info(
...
+        for data in ensemble_performance_history:
+            self.update(
+                train_score=data[f'train_{self._metric.name}'],
+                test_score=data[f'test_{self._metric.name}'],
+                end_time=datetime.timestamp(data['Timestamp'])
+            )
+
+
 class SearchResults:
     def __init__(
```
ravinkohli, line 179: can you add documentation for this class?

autoPyTorch/api/results_manager.py
```diff
@@ -179,9 +322,113 @@ def _extract_results_from_run_history(self, run_history: RunHistory) -> None:
         )
 
 
+class MetricResults:
+    def __init__(
```
ravinkohli, line 326: Could you add documentation for this class?

test/test_api/test_results_manager.py
```diff
@@ -88,8 +113,43 @@ def _check_metric_dict(metric_dict, status_types):
...
         # ^ is a XOR operator
         # True and False / False and True must be fulfilled
-        assert all([(s == STATUS2MSG[StatusType.SUCCESS]) ^ isnan
-                    for s, isnan in zip(status_types, np.isnan(vals))])
+        assert all([(s == StatusType.SUCCESS) ^ np.isclose([val], [worst_val])
+                    for s, val in zip(status_types, vals)])
+
+
+def _check_metric_results(scores, metric, run_history, ensemble_performance_history):
+    if metric.name == 'accuracy':
```
ravinkohli, line 121: could you explain the need for this condition?

autoPyTorch/api/results_manager.py
nabenabe0928: Oh, thanks for the catch.

test/test_api/test_results_manager.py
nabenabe0928: I added the doc string.
But just in case I will explain.
In this case, we are interested in `accuracy`, but ensemble history has only `log_loss` as hard-coded.
In this case, ensemble results should be None and merged data also should include the same data as the data obtained from `run_history`.

ravinkohli: okay. Thanks

ravinkohli (approved):
Thanks for your changes, I think this PR can be merged now.

ravinkohli (commented):
Sorry for the previous comment. I'll wait for the example and then this PR can be merged.

examples/40_advanced/example_plot_over_time.py
```diff
@@ -0,0 +1,75 @@
...
+
+from autoPyTorch.api.results_visualizer import PlotSettingParams
+from autoPyTorch.api.tabular_classification import TabularClassificationTask
+
+
+warnings.simplefilter(action='ignore', category=UserWarning)
+warnings.simplefilter(action='ignore', category=FutureWarning)
+
+
+if __name__ == '__main__':
```
ravinkohli, line 34: I think you can remove this `__main__` check as you are not using multiple jobs.

PR #332, opened 2021-11-16
[ADD] Search statistics

ravinkohli:
Addresses #326.

Issue #326, opened 2021-11-15
[ADD] Functionality to display run stats like crashed runs, successful runs etc.


ravinkohli:
Currently, after the search is completed we can only show the whole runhistory data dict, however its a lot of information to parse. It would be useful to show the statistics of the search for example number of successful runs, number of crashed runs, etc. This functionality can be based on auto-sklearn's [`sprint_statistics`](https://github.com/automl/auto-sklearn/blob/e21047aa7b52e762a58992e33ffcebb420586e67/autosklearn/automl.py#L1795).

PR #334, opened 2021-11-19
[feat] Support statistics print by adding results manager object

nabenabe0928:
This PR addresses the [ISSUE#326](https://github.com/automl/Auto-PyTorch/issues/326) and took over the ownership from the [PR#332](https://github.com/automl/Auto-PyTorch/pull/332) authored by @ravinkohli .

ToDo
- [x] Make test more comprehensive
- [x] Make test stronger


nabenabe0928:
Toy example is the following:
```
import warnings

import numpy as np
import pandas as pd

from sklearn import model_selection

from autoPyTorch.api.tabular_classification import TabularClassificationTask


warnings.simplefilter(action='ignore', category=UserWarning)
warnings.simplefilter(action='ignore', category=FutureWarning)


n_samples, dim = 50, 2
X = np.random.random((n_samples, dim)) * 2 - 1
y = ((X ** 2).sum(axis=-1) < 2 / np.pi).astype(np.int32)
print(y)

X, y = pd.DataFrame(X), pd.DataFrame(y)
X_train, X_test, y_train, y_test = model_selection.train_test_split(X, y)

api = TabularClassificationTask(seed=42)

api.search(X_train=X_train, y_train=y_train, X_test=X_test, y_test=y_test, optimize_metric='accuracy',
           total_walltime_limit=300, func_eval_time_limit_secs=10)

# Check the statistics (, which has been added in this PR)
print(api.sprint_statistics())
# Check the incumbent (, which has been modified in this PR)
api.get_incumbent_results(include_traditional=True)
```

autoPyTorch/api/results_manager.py
```diff
@@ -0,0 +1,318 @@
...
+        if self.run_history is None:
+            raise RuntimeError("No Run History found, search has not been called.")
+
+        if self.run_history.empty():
+            raise RuntimeError("Run History is empty. Something went wrong, "
+                               "SMAC was not able to fit any model?")
+
+    def get_incumbent_results(
+        self,
+        metric: autoPyTorchMetric,
```
ravinkohli, line 201: can we have this metric argument as`Optional` ? I think the default behaviour of this function should be to return the incumbent result according to the `self._metric`.

autoPyTorch/api/results_manager.py
```diff
@@ -0,0 +1,318 @@
...
+        include_traditional: bool = False
+    ) -> Tuple[Configuration, Dict[str, Union[int, str, float]]]:
+        """
+        Get Incumbent config and the corresponding results
+
+        Args:
+            metric (autoPyTorchMetric):
+                A metric that is used to fit AutoPytorch.
```
ravinkohli, line 209:
```suggestion
                A metric that is evaluated when searching with AutoPytorch.
```

ravinkohli: I want to avoid using fit as we don't have a `fit` function.

autoPyTorch/api/results_manager.py
```diff
@@ -0,0 +1,318 @@
...
+        6. Total number of target algorithm runs that exceeded the time limit
+        7. Total number of successful target algorithm runs that exceeded the memory limit
+
+        Args:
+            dataset_name (str):
+                The dataset name that was used in the run.
+            scoring_functions (List[autoPyTorchMetric]):
+                Metrics to show in the results.
+            metric (autoPyTorchMetric):
+                A metric that is used to fit AutoPytorch.
```
ravinkohli, line 287: same here

ravinkohli (changes requested):
Overall I think the PR can be merged with a few changes. I have suggested minor changes to the docstring. Also as we are only testing `get_incumbent_results` with the accuracy we should also test for a different metric for example 'logloss' for classification where greater is not better. Could you add it in [`test_get_incumbent_results`](https://github.com/automl/Auto-PyTorch/blob/a1512d56d4db89133e895e85765e3b72afbfe157/test/test_api/test_api.py#L580)?

nabenabe0928:
> Overall I think the PR can be merged with a few changes. I have suggested minor changes to the docstring. Also as we are only testing `get_incumbent_results` with the accuracy we should also test for a different metric for example 'logloss' for classification where greater is not better. Could you add it in [`test_get_incumbent_results`](https://github.com/automl/Auto-PyTorch/blob/a1512d56d4db89133e895e85765e3b72afbfe157/test/test_api/test_api.py#L580)?

I addressed it:)

test/test_api/test_results_manager.py
```diff
@@ -0,0 +1,216 @@
...
+        # ^ is a XOR operator
+        # True and False / False and True must be fulfilled
+        assert all([(s == STATUS2MSG[StatusType.SUCCESS]) ^ isnan
+                    for s, isnan in zip(status_types, np.isnan(vals))])
+
+
+def test_search_results_sprint_statistics():
+    api = BaseTask()
+    for method in ['get_search_results', 'sprint_statistics', 'get_incumbent_results']:
+        try:
```
ravinkohli, line 98: you can use pytest.raises(RuntimeError) instead of try and except.

test/test_api/test_results_manager.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+@pytest.mark.parametrize('run_history', (None, RunHistory()))
+def test_check_run_history(run_history):
+    manager = ResultsManager()
+    manager.run_history = run_history
+
+    try:
+        manager._check_run_history()
+    except RuntimeError:
```
ravinkohli, line 156: same here

test/test_api/test_results_manager.py
```diff
@@ -0,0 +1,216 @@
...
+
+    incumbent_config, incumbent_results = manager.get_incumbent_results(
+        metric=metric,
+        include_traditional=include_traditional
+    )
+
+    assert isinstance(incumbent_config, Configuration)
+    assert isinstance(incumbent_results, dict)
+    best_score, best_a = scores[best_idx], configs[best_idx]
+    assert abs(best_score - cost2metric(best_cost, metric)) < 1e-5
```
ravinkohli, line 211: do you think we can use [`np.isclose`](https://numpy.org/doc/stable/reference/generated/numpy.isclose.html`)

ravinkohli (commented):
I have suggested some minor changes in the tests. After they are resolved, I think this PR can be merged.

PR #338, opened 2021-11-22
[FIX] Silence catboost

ravinkohli:
This PR silences the output of the catboost traditional learner.


nabenabe0928 (approved):
We can merge this

Issue #339, opened 2021-11-22
[BUG] Incorrect cost for crash, r2 metric. 


ravinkohli:
The cost of an unsuccessful configuration such as a crash, time out or mem out is not the lowest in the run history. I think the issue is due to the worst_possible_result of r2 to be 0 instead of a large negative value.

Currently, we assume that the worst r2 value is 0, however, it can be worse as well. So, for [this](https://github.com/automl/Auto-PyTorch/runs/4287084282?check_suite_focus=true) test, `get_incumbent_results` is giving a failed configuration. 

`run_history.data`
```
[...]

PR #340, opened 2021-11-22
[FIX] Incorrect cost for crash, r2

ravinkohli:
Addresses #339


nabenabe0928 (approved):
Thanks for the fix
We can merge this!

Issue #345, opened 2021-11-30
Weird if statements in `SimpleImputer.py`


eddiebergman:
I was browsing some code in `SimpleImputer.py` and I really have no idea how this multi-line double if statement with a `0 != 0` in it works? Little cleaning job I guess

* https://github.com/automl/Auto-PyTorch/blob/master/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py#L91

```python
if len(dataset_properties['numerical_columns']) \
        if isinstance(dataset_properties['numerical_columns'], List) else 0 != 0:
    add_hyperparameter(cs, numerical_strategy, CategoricalHyperparameter)
```

nabenabe0928:
solved

PR #346, opened 2021-11-30
Cleanup of simple_imputer

eddiebergman:
Does a clean-up of `SimpleImputer` stemming from the weirdness of the `if` statements, mentioned in #345. Does not change functionality.

* Simplifies the `fit` method a bit
* Add docstrings
* Fixes some typing
* Document unexpected functionality of passing `'constant'` for `categorical_strategy` instead of `'constant_!missing!'`.
    * `'constant'` - fill_value = 0  *(sklearn default)*
    * `'constant_!missing!'` - fill_value = -1

### DIFF
--- a/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
+++ b/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
@@ -26,7 +26,7 @@ class SimpleImputer(BaseImputer):
             The random state to use for the imputer.
         numerical_strategy (str: default='mean'):
             The strategy to use for imputing numerical columns.
-            Can be one of ['mean', 'median', 'most_frequent', 'constant', 'constant_!missing!']
+            Can be one of ['most_frequent', 'constant_!missing!']
         categorical_strategy (str: default='most_frequent')
             The strategy to use for imputing categorical columns.
             Can be one of ['mean', 'median', 'most_frequent', 'constant_zero']
--- a/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
+++ b/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
@@ -33,7 +33,7 @@ class SimpleImputer(BaseImputer):
     """
 
     def __init__(
-        random_state: Optional[Union[np.random.RandomState, int]] = None,
+        self,
         random_state: Optional[np.random.RandomState] = None,
         numerical_strategy: str = 'mean',
         categorical_strategy: str = 'most_frequent'
--- a/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
+++ b/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
@@ -58,7 +58,8 @@ class SimpleImputer(BaseImputer):
                 The labels for the input features `X`
 
         Returns:
-            SimpleImputer: returns the object itself
+            SimpleImputer:
+                returns self
         """
         self.check_requirements(X, y)
 
@@ -145,7 +146,8 @@ class SimpleImputer(BaseImputer):
         """Get the properties of the SimpleImputer class and what it can handle
 
         Returns:
-            Dict[str, Union[str, bool]]: A dict from property names to values
+            Dict[str, Union[str, bool]]:
+                A dict from property names to values
         """
         return {
             'shortname': 'SimpleImputer',
--- a/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
+++ b/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
@@ -1,9 +1,7 @@
 from typing import Any, Dict, List, Optional, Union
 
 from ConfigSpace.configuration_space import ConfigurationSpace
-from ConfigSpace.hyperparameters import (
-    CategoricalHyperparameter
-)
+from ConfigSpace.hyperparameters import CategoricalHyperparameter
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
```diff
@@ -15,92 +13,154 @@
...
-                 numerical_strategy: str = 'mean',
-                 categorical_strategy: str = 'most_frequent'):
+    def __init__(
+        self,
+        random_state: Optional[Union[np.random.RandomState, int]] = None,
+        numerical_strategy: str = 'mean',
+        categorical_strategy: str = 'most_frequent'
+    ):
+        """
+        Parameters
```
ravinkohli, line 32: Actually, we follow google style docstring. You can find more information about it [here](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html). Also, you can check out the docstrings as they are used in AutoPyTorch [here](https://github.com/automl/Auto-PyTorch/blob/ef8d21ae982883548a8e77a3690eefa810d3ae07/autoPyTorch/api/base_task.py#L107).

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
```diff
@@ -29,45 +29,36 @@ def __init__(
...
 
-        y: Optional[np.ndarray]
-            The labels for the input features `X`
-
-        Returns
-        -------
-        SimpleImputer
-            returns self
+        Returns:
+            SimpleImputer: returns the object itself
```
ravinkohli, line 61: Can you add a new line after `SimpleImputer:`? Otherwise, the formatting of documentation generated by Sphinx causes issues.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
```diff
@@ -156,10 +144,8 @@ def get_properties(
     ) -> Dict[str, Union[str, bool]]:
         """Get the properties of the SimpleImputer class and what it can handle
 
-        Returns
-        -------
-        Dict[str, Union[str, bool]]
-            A dict from property names to values
+        Returns:
+            Dict[str, Union[str, bool]]: A dict from property names to values
```
ravinkohli, line 148: Same here

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
```diff
@@ -15,92 +13,142 @@
...
+        that categorical data is fit with an Ordinal Scaler.
     """
 
-    def __init__(self,
-                 random_state: Optional[Union[np.random.RandomState, int]] = None,
-                 numerical_strategy: str = 'mean',
-                 categorical_strategy: str = 'most_frequent'):
+    def __init__(
+        self,
+        random_state: Optional[Union[np.random.RandomState, int]] = None,
```
nabenabe0928, line 27: We removed integer type from `random_state`, so it must be `Optional[np.random.RandomState]`

nabenabe0928 (changes requested):
Hello thanks for the PR.
Since we had some changes (probably those are not in ASK), so I will leave some comments.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
```diff
@@ -15,92 +13,142 @@
...
         dataset_properties: Optional[Dict[str, BaseDatasetPropertiesType]] = None,
-        numerical_strategy: HyperparameterSearchSpace = HyperparameterSearchSpace(hyperparameter='numerical_strategy',
-                                                                                  value_range=("mean", "median",
-                                                                                               "most_frequent",
-                                                                                               "constant_zero"),
-                                                                                  default_value="mean",
-                                                                                  ),
+        numerical_strategy: HyperparameterSearchSpace = HyperparameterSearchSpace(
+            hyperparameter='numerical_strategy',
+            value_range=("mean", "median", "most_frequent", "constant_zero"),
```
nabenabe0928, line 97: It would be better if we can use `Enum`.
We did not have it so much just because I did not join the refactoring phase.
At least, I would like to reduce so many dependencies on hard-coded strings as much as possible.
Because this file also uses so many hard-coded strings, which we can avoid by enum.

```
from enum import Enum
from typing import List


class NumericalImputerChoices(Enum):
    mean = "mean"
    median = "median"
    most_frequent = "most_frequent"
    constant_zero = "constant_zero"

    @classmethod
    def get_choices(cls) -> List[str]:
        return [c.value for c in cls]


class CategoricalImputerChoices(Enum):
    most_frequent = "most_frequent"
    constant_missing = "constant_!missing!"

    @classmethod
    def get_choices(cls) -> List[str]:
        return [c.value for c in cls]
```

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
```diff
@@ -15,92 +13,142 @@
...
+    def __init__(
+        self,
+        random_state: Optional[Union[np.random.RandomState, int]] = None,
+        numerical_strategy: str = 'mean',
+        categorical_strategy: str = 'most_frequent'
+    ):
+        """
+        Note:
+            Using 'constant' defaults to fill_value of 0 where 'constant_!missing!'
+            uses a fill_value of -1. This behaviour should probably be fixed.
```
nabenabe0928, line 34: Is this what you mean?
```suggestion
            'constant' uses fill_value of 0 by default while
            'constant_!missing!' uses fill_value of -1. This behaviour should probably be fixed.
```

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
```diff
@@ -15,92 +13,142 @@
...
+
+        Args:
+            random_state (Optional[Union[np.random.RandomState, int]]):
+                The random state to use for the imputer.
+            numerical_strategy (str: default='mean'):
+                The strategy to use for imputing numerical columns.
+                Can be one of ['mean', 'median', 'most_frequent', 'constant', 'constant_!missing!']
+            categorical_strategy (str: default='most_frequent')
+                The strategy to use for imputing categorical columns.
+                Can be one of ['mean', 'median', 'most_frequent', 'constant_zero']
```
nabenabe0928, line 44: We can move this doc-string to the enums once we make them. 

Btw, choices of `categorical_strategy` is incompatible with the hyperparamer search space.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
```diff
@@ -15,92 +13,142 @@
...
+                `dataset_properties`
+        """
         cs = ConfigurationSpace()
-        assert dataset_properties is not None, "To create hyperparameter search space" \
-                                               ", dataset_properties should not be None"
-        if len(dataset_properties['numerical_columns']) \
-                if isinstance(dataset_properties['numerical_columns'], List) else 0 != 0:
+
+        if dataset_properties is None:
+            raise ValueError("SimpleImputer requires `dataset_properties` for generating"
```
nabenabe0928, line 125: Could you add a test here?

nabenabe0928 (commented):
Hi thanks for cleaning up this file.
It is much better than the previous state:)
I left some comments as well.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
eddiebergman: Okay, will do. As an fyi, this can cause different output if you run the same function twice on the same object. The random state produces a sequence of numbers.

For example, if you create a single RandomState object and pass it to every object that requires a random_state, you will get different output depending on the order in which objects use that random_state. On the flip-side, if you use an int, they are independant of each other and so it doesn't matter which order objects use it.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
eddiebergman: I can do that, you do however require to update these whenever sklearn updates them as well as that is where they are forwarded to.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
eddiebergman: Sure, I'm just not sure what you mean by the second point, should I do something with respect to this?

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
eddiebergman: Sure, it wasn't tested before but I can add one.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
eddiebergman: I can see why you want to do that but it makes the code that uses it a little less pretty. For example, it's hard to see where the error with the follow code is:

```python
        numerical_strategy: HyperparameterSearchSpace = HyperparameterSearchSpace(
            hyperparameter='numerical_strategy',
            value_range=NumericalImputerChoice.get_choices(),
            default_value=NumericalImputerChoice.mean,
        ),
```

It should be `default_value=NumericalImputerChoice.mean.value`. The problem here is that Enum's are namespaces `NumericalImputerChoice.mean == "NumericalImputerChoice.mean"` where as we would like `NumericalImputerChoice.mean == "mean"`. This is more readily achievable with a `NamedTuple`. I will implement the Enum version and you can change it if you like or leave it as is

eddiebergman: Another clean'ish solution is just ditch the enum part. 

```python
class Choices:
    x: str = "x"
    y: str = "y"
    
assert Choices.x = "x"
```

The type is still a string, meaning it's easy to use, people don't need to know about the existence of this class to use it, they can still pass a string. It also allows for the internal code to use the class, where we do know about it.

eddiebergman: As another note, it also makes parameters extremely long, it's 101 characters which is over the character count limit of the checker:
```python
    def __init__(
        self,
        random_state: Optional[np.random.RandomState] = None,
        numerical_strategy: NumericalImputerChoice = NumericalImputerChoice.mean.value,
        categorical_strategy: CategoricalImputerChoice = CategoricalImputerChoice.most_frequent.value
    ):
```

eddiebergman: I've not changed it for now, there's too many decisions to make that I think can be addressed based on how you would like it done yourself. I've changed it back to strings.

ravinkohli: Hey, thanks for the insights. Yeah, I think such changes should be a part of a separate PR as this PR is meant to clean up the messy statements in this file.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
nabenabe0928: Yeah, that is also true. My explanation was not sufficient, but we, in reality, decided to use `seed` for `int` and `random_state` for `np.random.RandomState`.
So it is a very good decision if we switch to `seed` instead of `random_state` in the future.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
nabenabe0928: we have the following as the default searching space, right?
So I think either this value_range or your doc-string should follow the other.
I am assuming that you take the doc-string from AutoSKLearn, so probably this `value_range` should be replaced, don't you think?
```
        categorical_strategy: HyperparameterSearchSpace = HyperparameterSearchSpace(
            hyperparameter='categorical_strategy',
            value_range=("most_frequent", "constant_!missing!"),
            default_value="most_frequent"
        )
```

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
eddiebergman: We kept `random_state` as the arg name and allow for both, internally we just pass the `seed` argument given at construction of an `AutoSklearnClassifier` so that internally it's an int passed throughout.

We follow sklearn in principle so we copy [their expected behaviour](https://scikit-learn.org/stable/common_pitfalls.html#using-none-or-randomstate-instances-and-repeated-calls-to-fit-and-split).

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
nabenabe0928: All the problems come from the fact that we still need to follow python3.7.
Once we switched to python3.8, we can use `Literal`...

eddiebergman: We have the same thing, there is `Literal` at the moment but the type checking for it doesn't work as expected.

```python
def f(s: Literal['yes', 'no'])
   ...
   
f("yes") # Type Error

param: Literal = "yes"
f(param) # Okay
```

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
eddiebergman: Ahh my bad, I see your point. I've updated the docstring rather than update the search space, just to be inline with what's there so the actual functionality doesn't change.

I actually looked at the `sklearn` docs for possible values

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
nabenabe0928: All the problems come from the fact that we still need to follow python3.7.
Once we switched to python3.8, we can use `Literal`...

> in general, enum values should never be directly used, such as their string value

Yeah, I agree with you. I am just uncomfortable using software that assumes that users google string choices (typically sklearn).
And I found it really useful to be able to use `enum` or `class` such as in numpy, facebook ax..
But I got your point, let's stay for now.
I did not know some points you mentioned, thanks for raising them. @eddiebergman .

(just a question, but) the better solution will be something like this:
```
class NumericalImputerChoices(NamedTuple):
    mean = "mean"
    median = "median"

num_imputer_choices = NumericalImputerChoices()
```
and use `str` for all the typings.

Btw, you do not need to put `.value` (FYI).

eddiebergman: also, you can add `python` after the backticks ``` of a code block to get python highlighting ;)

nabenabe0928 (approved):

PR #348, opened 2021-11-30
[ADD] fit pipeline honoring API constraints with tests

### LINKED ISSUE #149: Improve fitting a pipeline to the dataset

ravinkohli:
## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

## Description

This PR fixes the functionality to fit a single configuration using the API. In its current implementation, `api.fit()` does not honour the constraints passed through the `pipeline_options` and as parameters to the function correctly. To enable this, I have created a `get_dataset` which encapsulates the code needed to create a new `Dataset` object. In addition, I have renamed `fit` to `fit_pipeline` to avoid confusion with the use of `fit` which is generally used to fit automl (search in our case) in other automl libraries. 

Also, `build_pipeline()` did not use the include and exclude components.

## Motivation and Context

This is needed in case someone wants to fit a single configuration using the constraints correctly.. 

Fixes #149. 
## How has this been tested?

I have made a test that checks if the default configuration of a pipeline can be fitted properly. It also checks other objects returned by the function `fit_pipeline`. I also check if the accuracy is greater than 70%. To ensure that `disable_file_output` is working as expected, the test also verifies if the pipeline is stored in the correct folder and it can be retrieved from there when `disable_file_output` is None. When it is `['all']`, the test ensures that no files are stored.

### DIFF
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -314,8 +314,7 @@ class BaseTask(ABC):
             BaseInputValidator:
                 fitted input validator
         """
-        raise NotImplementedError("Function called on BaseTask, this can only be called by "
-                                  "specific task which is a child of the BaseTask")
+        raise NotImplementedError
 
     def get_dataset(
         self,
@@ -1495,8 +1494,7 @@ class BaseTask(ABC):
         # TAE expects each configuration to have a config_id.
         # For fitting a pipeline as it is not part of the
         # search process, it makes sense to set it to 0
-        if hasattr(configuration, 'config_id') or configuration.config_id is None:
-            configuration.__setattr__('config_id', 0)
+        configuration.__setattr__('config_id', 0)
 
         # get dataset properties
         dataset_requirements = get_dataset_requirements(
@@ -1582,27 +1580,42 @@ class BaseTask(ABC):
                     instance=None)
         )
 
-        fitted_pipeline: Optional[BasePipeline] = None
+        fitted_pipeline = self._get_fitted_pipeline(
+            pipeline_idx=run_info.config.config_id + tae.initial_num_run,
+            run_info=run_info,
+            run_value=run_value,
+            disable_file_output=disable_file_output
+        )
 
-        if run_value.status == StatusType.SUCCESS:
-            if 'all' in disable_file_output or 'pipeline' in disable_file_output:
-                self._logger.warning("File output is disabled. No pipeline can returned")
-            else:
-                if self.resampling_strategy in CrossValTypes:
-                    load_function = self._backend.load_cv_model_by_seed_and_id_and_budget
-                else:
-                    load_function = self._backend.load_model_by_seed_and_id_and_budget
-                fitted_pipeline = load_function(
-                    seed=self.seed,
-                    idx=run_info.config.config_id + tae.initial_num_run,
-                    budget=float(run_info.budget),
-                )
-        else:
-            warnings.warn(f"Fitting pipeline failed with status: {run_value.status}"
-                          f", additional_info: {run_value.additional_info}")
         self._clean_logger()
 
         return fitted_pipeline, run_info, run_value, dataset
+
+    def _get_fitted_pipeline(
+        self,
+        pipeline_idx: int,
+        run_info: RunInfo,
+        run_value: RunValue,
+        disable_file_output: List[Union[str, DisableFileOutputParameters]]
+    ) -> Optional[BasePipeline]:
+        if run_value.status != StatusType.SUCCESS:
+            warnings.warn(f"Fitting pipeline failed with status: {run_value.status}"
+                          f", additional_info: {run_value.additional_info}")
+            return None
+        elif any(disable_file_output for c in ['all', 'pipeline']):
+            self._logger.warning("File output is disabled. No pipeline can returned")
+            return None
+
+        if self.resampling_strategy in CrossValTypes:
+            load_function = self._backend.load_cv_model_by_seed_and_id_and_budget
+        else:
+            load_function = self._backend.load_model_by_seed_and_id_and_budget
+
+        return load_function(
+            seed=self.seed,
+            idx=pipeline_idx,
+            budget=float(run_info.budget),
+        )
 
     def predict(
         self,
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -1491,6 +1491,10 @@ class BaseTask(ABC):
                                        dataset_name=dataset_name
                                        )
 
+        # dataset_name is created inside the constructor of BaseDataset
+        # we expect it to be not None. This is for mypy
+        assert dataset.dataset_name is not None
+
         # TAE expects each configuration to have a config_id.
         # For fitting a pipeline as it is not part of the
         # search process, it makes sense to set it to 0
@@ -1506,9 +1510,6 @@ class BaseTask(ABC):
         self._backend.save_datamanager(dataset)
 
         if self._logger is None:
-            # dataset_name is created inside the constructor of BaseDataset
-            # we expect it to be not None. This is for mypy
-            assert dataset.dataset_name is not None
             self._logger = self._get_logger(dataset.dataset_name)
 
         include_components = self.include_components if include_components is None else include_components
@@ -1576,6 +1577,7 @@ class BaseTask(ABC):
         )
 
         fitted_pipeline = self._get_fitted_pipeline(
+            dataset_name=dataset.dataset_name,
             pipeline_idx=run_info.config.config_id + tae.initial_num_run,
             run_info=run_info,
             run_value=run_value,
@@ -1588,11 +1590,16 @@ class BaseTask(ABC):
 
     def _get_fitted_pipeline(
         self,
+        dataset_name: str,
         pipeline_idx: int,
         run_info: RunInfo,
         run_value: RunValue,
         disable_file_output: List[Union[str, DisableFileOutputParameters]]
     ) -> Optional[BasePipeline]:
+
+        if self._logger is None:
+            self._logger = self._get_logger(str(dataset_name))
+
         if run_value.status != StatusType.SUCCESS:
             warnings.warn(f"Fitting pipeline failed with status: {run_value.status}"
                           f", additional_info: {run_value.additional_info}")
@@ -1606,7 +1613,7 @@ class BaseTask(ABC):
         else:
             load_function = self._backend.load_model_by_seed_and_id_and_budget
 
-        return load_function(
+        return load_function(  # type: ignore[no-any-return]
             seed=self.seed,
             idx=pipeline_idx,
             budget=float(run_info.budget),
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -27,7 +27,7 @@ import numpy as np
 
 import pandas as pd
 
-from smac.runhistory.runhistory import DataOrigin, RunHistory
+from smac.runhistory.runhistory import DataOrigin, RunHistory, RunInfo, RunValue
 from smac.stats.stats import Stats
 from smac.tae import StatusType
 
@@ -233,7 +233,11 @@ class BaseTask:
                                  " HyperparameterSearchSpaceUpdates got {}".format(type(self.search_space_updates)))
 
     @abstractmethod
-    def build_pipeline(self, dataset_properties: Dict[str, Any]) -> BasePipeline:
+    def build_pipeline(self, dataset_properties: Dict[str, Any],
+                       include_components: Optional[Dict] = None,
+                       exclude_components: Optional[Dict] = None,
+                       search_space_updates: Optional[HyperparameterSearchSpaceUpdates] = None
+                       ) -> BasePipeline:
         """
         Build pipeline according to current task
         and for the passed dataset properties
@@ -244,7 +248,21 @@ class BaseTask:
         Returns:
 
         """
-        raise NotImplementedError
+        raise NotImplementedError("Function called on BaseTask, this can only be called by "
+                                  "specific task which is a child of the BaseTask")
+
+    @abstractmethod
+    def get_dataset(self,
+                    X_train: Union[List, pd.DataFrame, np.ndarray],
+                    y_train: Union[List, pd.DataFrame, np.ndarray],
+                    X_test: Union[List, pd.DataFrame, np.ndarray],
+                    y_test: Union[List, pd.DataFrame, np.ndarray],
+                    resampling_strategy: Optional[Union[CrossValTypes, HoldoutValTypes]] = None,
+                    resampling_strategy_args: Optional[Dict[str, Any]] = None,
+                    dataset_name: Optional[str] = None,
+                    ) -> BaseDataset:
+        raise NotImplementedError("Function called on BaseTask, this can only be called by "
+                                  "specific task which is a child of the BaseTask")
 
     @property
     def run_history(self) -> RunHistory:
@@ -563,7 +581,7 @@ class BaseTask:
             initial_num_run=num_run,
             stats=stats,
             memory_limit=memory_limit,
-            disable_file_output=True if len(self._disable_file_output) > 0 else False,
+            disable_file_output=self._disable_file_output,
             all_supported_metrics=self._all_supported_metrics
         )
 
@@ -647,7 +665,7 @@ class BaseTask:
                     initial_num_run=self._backend.get_next_num_run(),
                     stats=stats,
                     memory_limit=memory_limit,
-                    disable_file_output=True if len(self._disable_file_output) > 0 else False,
+                    disable_file_output=self._disable_file_output,
                     all_supported_metrics=self._all_supported_metrics
                 )
                 dask_futures.append([
@@ -743,7 +761,7 @@ class BaseTask:
         tae_func: Optional[Callable] = None,
         all_supported_metrics: bool = True,
         precision: int = 32,
-        disable_file_output: List = [],
+        disable_file_output: Optional[List[str]] = None,
         load_models: bool = True,
         portfolio_selection: Optional[str] = None,
         dask_client: Optional[dask.distributed.Client] = None
@@ -844,9 +862,8 @@ class BaseTask:
             precision (int: default=32):
                 Numeric precision used when loading ensemble data.
                 Can be either '16', '32' or '64'.
-            disable_file_output (Union[bool, List]):
-                If True, disable model and prediction output.
-                Can also be used as a list to pass more fine-grained
+            disable_file_output (Optional[List]):
+                Used as a list to pass more fine-grained
                 information on what to save. Allowed elements in the list are:
 
                 + `y_optimization`:
@@ -860,6 +877,8 @@ class BaseTask:
                     pipelines fit on each fold.
                 + `y_test`:
                     do not save the predictions for the test set.
+                + `all`:
+                    do not save any of the above.
             load_models (bool: default=True):
                 Whether to load the models after fitting AutoPyTorch.
             portfolio_selection (Optional[str]):
@@ -901,7 +920,7 @@ class BaseTask:
         self._backend.setup_logger(port=self._logger_port)
 
         self._all_supported_metrics = all_supported_metrics
-        self._disable_file_output = disable_file_output
+        self._disable_file_output = disable_file_output if disable_file_output is not None else []
         self._memory_limit = memory_limit
         self._time_for_task = total_walltime_limit
         # Save start time to backend
@@ -1223,10 +1242,29 @@ class BaseTask:
 
         return self
 
-    def fit(self,
-            dataset: BaseDataset,
-            pipeline_config: Optional[Configuration] = None,
-            split_id: int = 0) -> BasePipeline:
+    def fit_pipeline(
+        self,
+        configuration: Configuration,
+        dataset: Optional[BaseDataset] = None,
+        X_train: Optional[Union[List, pd.DataFrame, np.ndarray]] = None,
+        y_train: Optional[Union[List, pd.DataFrame, np.ndarray]] = None,
+        X_test: Optional[Union[List, pd.DataFrame, np.ndarray]] = None,
+        y_test: Optional[Union[List, pd.DataFrame, np.ndarray]] = None,
+        dataset_name: Optional[str] = None,
+        resampling_strategy: Optional[Union[HoldoutValTypes, CrossValTypes]] = None,
+        resampling_strategy_args: Optional[Dict[str, Any]] = None,
+        run_time_limit_secs: int = 60,
+        memory_limit: Optional[int] = None,
+        eval_metric: Optional[str] = None,
+        all_supported_metrics: bool = False,
+        budget_type: Optional[str] = None,
+        include_components: Optional[Dict] = None,
+        exclude_components: Optional[Dict] = None,
+        search_space_updates: Optional[HyperparameterSearchSpaceUpdates] = None,
+        budget: Optional[float] = None,
+        pipeline_options: Optional[Dict] = None,
+        disable_file_output: Optional[List[str]] = None,
+    ) -> Tuple[Optional[BasePipeline], RunInfo, RunValue, BaseDataset]:
         """
         Fit a pipeline on the given task for the budget.
         A pipeline configuration can be specified if None,
@@ -1237,24 +1275,110 @@ class BaseTask:
         methods.
 
         Args:
-            dataset (Dataset):
-                The argument that will provide the dataset splits. It can either
-                be a dictionary with the splits, or the dataset object which can
-                generate the splits based on different restrictions.
-            split_id (int: default=0):
-                split id to fit on.
-            pipeline_config (Optional[Configuration]):
-                configuration to fit the pipeline with. If None,
-                uses default
+            X_train, y_train, X_test, y_test: Union[np.ndarray, List, pd.DataFrame]
+                A pair of features (X_train) and targets (y_train) used to fit a
+                pipeline. Additionally, a holdout of this pairs (X_test, y_test) can
+                be provided to track the generalization performance of each stage.
+            dataset_name (Optional[str]):
+                Name of the dataset, if None, random value is used.
+            resampling_strategy (Union[CrossValTypes, HoldoutValTypes]),
+                (default=HoldoutValTypes.holdout_validation):
+                strategy to split the training data.
+            resampling_strategy_args (Optional[Dict[str, Any]]):
+                Arguments required for the chosen resampling strategy. If None, uses
+                the default values provided in DEFAULT_RESAMPLING_PARAMETERS
+                in ```datasets/resampling_strategy.py```.
+            run_time_limit_secs (int: default=120):
+                Time limit for a single call to the machine learning model.
+                Model fitting will be terminated if the machine learning algorithm
+                runs over the time limit. Set this value high enough so that
+                typical machine learning algorithms can be fit on the training
+                data.
+            memory_limit (Optional[int]):
+                Memory limit in MB for the machine learning algorithm. autopytorch
+                will stop fitting the machine learning algorithm if it tries
+                to allocate more than memory_limit MB. If None is provided,
+                no memory limit is set. In case of multi-processing, memory_limit
+                will be per job. This memory limit also applies to the ensemble
+                creation process.
+            eval_metric (str):
+                Name of the metric that is used to evaluate a pipeline.
+            all_supported_metrics (bool: default=True):
+                if True, all metrics supporting current task will be calculated
+                for each pipeline and results will be available via cv_results
+            budget_type (str):
+                Type of budget to be used when fitting the pipeline.
+                It can be one of:
+
+                + `epochs`: The training of each pipeline will be terminated after
+                    a number of epochs have passed. This number of epochs is determined by the
+                    budget argument of this method.
+                + `runtime`: The training of each pipeline will be terminated after
+                    a number of seconds have passed. This number of seconds is determined by the
+                    budget argument of this method. The overall fitting time of a pipeline is
+                    controlled by func_eval_time_limit_secs. 'runtime' only controls the allocated
+                    time to train a pipeline, but it does not consider the overall time it takes
+                    to create a pipeline (data loading and preprocessing, other i/o operations, etc.).
+            include_components (Optional[Dict]):
+                If None, all possible components are used.
+                Otherwise specifies set of components to use.
+            exclude_components (Optional[Dict]):
+                If None, all possible components are used.
+                Otherwise specifies set of components not to use.
+                Incompatible with include components
+            search_space_updates(Optional[HyperparameterSearchSpaceUpdates]):
+                Updates to be made to the hyperparameter search space of the pipeline
+            budget (Optional[float]):
+                Budget to fit a single run of the pipeline. If not
+                provided, uses the default in the pipeline config
+            pipeline_options (Optional[Dict]):
+                Valid config options include "device",
+                "torch_num_threads", "early_stopping", "use_tensorboard_logger",
+                "metrics_during_training"
+            disable_file_output (Optional[List]):
+                Used as a list to pass more fine-grained
+                information on what to save. Allowed elements in the list are:
+
+                + `y_optimization`:
+                    do not save the predictions for the optimization set,
+                    which would later on be used to build an ensemble. Note that SMAC
+                    optimizes a metric evaluated on the optimization set.
+                + `pipeline`:
+                    do not save any individual pipeline files
+                + `pipelines`:
+                    In case of cross validation, disables saving the joint model of the
+                    pipelines fit on each fold.
+                + `y_test`:
+                    do not save the predictions for the test set.
+                + `all`:
+                    do not save any of the above.
+            configuration: (Configuration)
+                configuration to fit the pipeline with.
 
         Returns:
-            BasePipeline:
-                fitted pipeline
+            (BasePipeline): fitted pipeline
+            (RunInfo): Run information
+            (RunValue): Result of fitting the pipeline
+            (BaseDataset): Dataset created from the given tensors
         """
-        self.dataset_name = dataset.dataset_name
 
-        if self._logger is None:
-            self._logger = self._get_logger(str(self.dataset_name))
+        if dataset is None:
+            assert X_train is not None and \
+                   y_train is not None, "No dataset provided, must provide X_train, y_train tensors"
+            dataset = self.get_dataset(X_train=X_train,
+                                       y_train=y_train,
+                                       X_test=X_test,
+                                       y_test=y_test,
+                                       resampling_strategy=resampling_strategy,
+                                       resampling_strategy_args=resampling_strategy_args,
+                                       dataset_name=dataset_name
+                                       )
+
+        # TAE expects each configuration to have a config_id.
+        # For fitting a pipeline as it is not part of the
+        # search process, it makes sense to set it to 0
+        if hasattr(configuration, 'config_id') or configuration.config_id is None:
+            configuration.__setattr__('config_id', 0)
 
         # get dataset properties
         dataset_requirements = get_dataset_requirements(
@@ -1265,21 +1389,98 @@ class BaseTask:
         dataset_properties = dataset.get_dataset_properties(dataset_requirements)
         self._backend.save_datamanager(dataset)
 
-        # build pipeline
-        pipeline = self.build_pipeline(dataset_properties)
-        if pipeline_config is not None:
-            pipeline.set_hyperparameters(pipeline_config)
+        if self._logger is None:
+            # dataset_name is created inside the constructor of BaseDataset
+            # we expect it to be not None. This is for mypy
+            assert dataset.dataset_name is not None
+            self._logger = self._get_logger(dataset.dataset_name)
 
-        # initialise fit dictionary
-        X = self._get_fit_dictionary(
-            dataset_properties=dataset_properties,
-            dataset=dataset,
-            split_id=split_id)
+        if include_components is None:
+            include_components = self.include_components
+        if exclude_components is None:
+            exclude_components = self.exclude_components
+        if search_space_updates is None:
+            search_space_updates = self.search_space_updates
 
-        fit_and_suppress_warnings(self._logger, pipeline, X, y=None)
+        scenario_mock = unittest.mock.Mock()
+        scenario_mock.wallclock_limit = run_time_limit_secs
+        # This stats object is a hack - maybe the SMAC stats object should
+        # already be generated here!
+        stats = Stats(scenario_mock)
+
+        if memory_limit is None:
+            if hasattr(self, '_memory_limit') and self._memory_limit is not None:
+                memory_limit = self._memory_limit
+
+        metric = get_metrics(dataset_properties=dataset_properties,
+                             names=[eval_metric] if eval_metric is not None else None,
+                             all_supported_metrics=False).pop()
+
+        pipeline_options = self.pipeline_options.copy().update(pipeline_options) if pipeline_options is not None \
+            else self.pipeline_options.copy()
+
+        assert pipeline_options is not None
+
+        if budget_type is not None:
+            pipeline_options.update({'budget_type': budget_type})
+        else:
+            budget_type = pipeline_options['budget_type']
+
+        budget = budget if budget is not None else pipeline_options[budget_type]
+
+        if disable_file_output is None:
+            disable_file_output = self._disable_file_output if hasattr(self, '_disable_file_output') \
+                and self._disable_file_output is not None else []
+
+        stats.start_timing()
+
+        tae = ExecuteTaFuncWithQueue(
+            backend=self._backend,
+            seed=self.seed,
+            metric=metric,
+            logger_port=self._logger_port,
+            cost_for_crash=get_cost_of_crash(metric),
+            abort_on_first_run_crash=False,
+            initial_num_run=self._backend.get_next_num_run(),
+            stats=stats,
+            memory_limit=memory_limit,
+            disable_file_output=disable_file_output,
+            all_supported_metrics=all_supported_metrics,
+            budget_type=budget_type,
+            include=include_components,
+            exclude=exclude_components,
+            search_space_updates=search_space_updates,
+            pipeline_config=pipeline_options,
+            pynisher_context=self._multiprocessing_context
+        )
+
+        run_info, run_value = tae.run_wrapper(
+            RunInfo(config=configuration,
+                    budget=budget,
+                    seed=self.seed,
+                    cutoff=run_time_limit_secs,
+                    capped=False,
+                    instance_specific=None,
+                    instance=None)
+        )
+
+        fitted_pipeline: Optional[BasePipeline] = None
+        if 'all' in disable_file_output or 'pipeline' in disable_file_output:
+            self._logger.warning("File output is disabled. No pipeline can returned")
+        elif run_value.status == StatusType.SUCCESS:
+            if self.resampling_strategy in CrossValTypes:
+                load_function = self._backend.load_cv_model_by_seed_and_id_and_budget
+            else:
+                load_function = self._backend.load_model_by_seed_and_id_and_budget
+            fitted_pipeline = load_function(
+                seed=self.seed,
+                idx=run_info.config.config_id + tae.initial_num_run,
+                budget=float(run_info.budget),
+            )
 
         self._clean_logger()
-        return pipeline
+
+        return fitted_pipeline, run_info, run_value, dataset
 
     def predict(
         self,
--- a/autoPyTorch/api/base_task.py
+++ b/autoPyTorch/api/base_task.py
@@ -136,12 +136,15 @@ class BaseTask(ABC):
             Determines whether to delete the temporary directory,
             when finished
         include_components (Optional[Dict[str, Any]]):
-            If None, all possible components are used.
-            Otherwise specifies set of components to use.
+            Dictionary containing components to include. Key is the node
+            name and Value is an Iterable of the names of the components
+            to include. Only these components will be present in the
+            search space.
         exclude_components (Optional[Dict[str, Any]]):
-            If None, all possible components are used.
-            Otherwise specifies set of components not to use.
-            Incompatible with include components
+            Dictionary containing components to exclude. Key is the node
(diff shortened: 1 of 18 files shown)


autoPyTorch/evaluation/abstract_evaluator.py
```diff
@@ -835,19 +843,17 @@ def file_output(
...
-            else:
-                self.disabled_file_outputs = []
 
         # This file can be written independently of the others down bellow
-        if 'y_optimization' not in self.disabled_file_outputs:
+        if 'y_optimization' not in self.disable_file_output:
             if self.output_y_hat_optimization:
                 self.backend.save_targets_ensemble(self.Y_optimization)
 
         if hasattr(self, 'pipelines') and self.pipelines is not None:
```
nabenabe0928, line 854:
```suggestion
        if (
            'pipelines' in self.disable_file_output
            not hasattr(self, 'pipelines')
            or self.pipelines is None
            or self.pipelines[0] is None
            or len(self.pipelines) == 0
        ):
            pipelines = None
        elif self.task_type in CLASSIFICATION_TASKS:
            pipelines = VotingClassifier(estimators=None, voting='soft', )
            pipelines.estimators_ = self.pipelines
        elif self.task_type in REGRESSION_TASKS:
            pipelines = VotingRegressorWrapper(estimators=None)
            pipelines.estimators_ = self.pipelines
        else:
            raise ValueError(
                f'task_type must be in {REGRESSION_TASKS} or {CLASSIFICATION_TASKS}, '
                f'but got {self.task_type}'
            )
```

autoPyTorch/api/base_task.py
```diff
@@ -1219,10 +1314,29 @@ def refit(
...
+        memory_limit: Optional[int] = None,
+        eval_metric: Optional[str] = None,
+        all_supported_metrics: bool = False,
+        budget_type: Optional[str] = None,
+        include_components: Optional[Dict] = None,
+        exclude_components: Optional[Dict] = None,
+        search_space_updates: Optional[HyperparameterSearchSpaceUpdates] = None,
+        budget: Optional[float] = None,
+        pipeline_options: Optional[Dict] = None,
+        disable_file_output: Optional[List[str]] = None,
```
nabenabe0928, line 1338: Why do we need that many variables now?
Most information is in the BaseDataset, isn't it?
Let's try reducing variable numbers as much as possible,
because it will be painful when you create tests and it will increase line numbers and doc-string in many places.

[Clean code](https://www.amazon.com/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882) also says methods with 5 variables should be considered to be re-designed.
Basically, we use an object to wrap those variables in such cases.

nabenabe0928 (commented):
Hi, thanks for the PR.
I haven't really checked base_task.py, tabular_xxx.py, so I will check them later.

autoPyTorch/evaluation/abstract_evaluator.py
```diff
@@ -861,7 +867,7 @@ def file_output(
             pipelines = None
 
         if hasattr(self, 'pipeline') and self.pipeline is not None:
-            if 'pipeline' not in self.disabled_file_outputs:
+            if 'pipeline' not in self.disable_file_output:
                 pipeline = self.pipeline
             else:
                 pipeline = None
```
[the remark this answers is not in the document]
ravinkohli: Actually, I am making such refactoring changes to the abstract evaluator in an upcoming PR to fix #349 . In this PR, my focus was to add a method to fit a pipeline. I changed `disable_file_output` because it was very messy.

autoPyTorch/evaluation/abstract_evaluator.py
```diff
@@ -448,12 +461,7 @@ def __init__(self, backend: Backend,
         # Flag to save target for ensemble
         self.output_y_hat_optimization = output_y_hat_optimization
 
-        if isinstance(disable_file_output, bool):
-            self.disable_file_output: bool = disable_file_output
-        elif isinstance(disable_file_output, List):
-            self.disabled_file_outputs: List[str] = disable_file_output
-        else:
-            raise ValueError('disable_file_output should be either a bool or a list')
+        self.disable_file_output = disable_file_output if disable_file_output is not None else []
```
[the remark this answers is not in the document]
ravinkohli: yeah, we should check for it. Thanks for reminding

autoPyTorch/evaluation/abstract_evaluator.py
ravinkohli: I have refactored this also in the upcoming PR.

autoPyTorch/api/base_task.py
ravinkohli: I agree we can get rid of some of these variables for example, `include_components`, `exclude_components` and `search_space_updates`.  I have intentionally kept the other parameters because these are needed features to be able to fit a pipeline. These parameters allow for fitting a pipeline independent of the search parameters. For example, we used a different resampling strategy (`NoResampling`) in our experiments with `reg_cocktails` which allows for fitting a pipeline with the whole train data rather than making splits with them.  We used it to get the full test accuracy of the incumbent pipeline.

Btw, dataset and X_train, y_train, X_test, y_test, dataset_name, resampling_strategy, resampling_strategy_args are exclusive to each other. The user can either send a `BaseDataset` or the data separately and we make the dataset. Would you prefer just having `dataset` so the user always has to get a dataset before fitting the pipeline?

ravinkohli: Btw, let me know which other parameters are not necessary for this function and we can discuss them

autoPyTorch/evaluation/abstract_evaluator.py
```diff
@@ -835,19 +843,17 @@ def file_output(
...
+            if 'all' in self.disable_file_output:
                 return None, {}
-            else:
-                self.disabled_file_outputs = []
 
         # This file can be written independently of the others down bellow
-        if 'y_optimization' not in self.disabled_file_outputs:
+        if 'y_optimization' not in self.disable_file_output:
             if self.output_y_hat_optimization:
                 self.backend.save_targets_ensemble(self.Y_optimization)
```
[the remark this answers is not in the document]
ravinkohli: as it says here, 
```
Whether this worker should output the target predictions, so that they are
stored on disk. Fundamentally, the resampling strategy might shuffle the
Y_train targets, so we store the split in order to re-use them for ensemble
selection.
```
It seems to me that our code for building an ensemble wont work if we don't run `self.backend.save_targets_ensemble(self.Y_optimization)`. So I guess we should remove this condition. Before doing that, as this code was used from `autosklearn`, @eddiebergman do you know the reason for this `self.output_y_hat_optimization` flag?

autoPyTorch/evaluation/train_evaluator.py
```diff
@@ -79,10 +79,23 @@ class TrainEvaluator(AbstractEvaluator):
...
+                Used as a list to pass more fine-grained
+                information on what to save. Allowed elements in the list are:
+
+                + `y_optimization`:
+                    do not save the predictions for the optimization set,
+                    which would later on be used to build an ensemble. Note that SMAC
+                    optimizes a metric evaluated on the optimization set.
+                + `pipeline`:
+                    do not save any individual pipeline files
+                + `pipelines`:
```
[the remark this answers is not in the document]
ravinkohli: So currently `pipeline` is being used to save the model for holdout resampling and `pipelines` is being used to save the model for cross-validation. I think we can just have `pipeline` instead of the two. Actually, we had an unnoticed logical error before when using cross-validation. Where we were storing a pipeline that has not been fitted yet. In my upcoming PR, I have removed this functionality as it was confusing and unnecessary. So, I'll remove `pipelines` from the set of options in that PR.

autoPyTorch/evaluation/train_evaluator.py
```diff
@@ -107,7 +120,7 @@ def __init__(self, backend: Backend, queue: Queue,
                  num_run: Optional[int] = None,
                  include: Optional[Dict[str, Any]] = None,
                  exclude: Optional[Dict[str, Any]] = None,
-                 disable_file_output: Union[bool, List] = False,
+                 disable_file_output: Optional[List] = [],
```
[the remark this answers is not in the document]
ravinkohli: yeah I have changed the default value to None, I don't prefer having mutable objects as default values for function parameters

autoPyTorch/api/base_task.py
ravinkohli: > Basically, we use an object to wrap those variables in such cases.

Okay, I think we can do this. It will be helpful also for other functions like `search` and `fit_ensemble`. Do you mind if I do it in another PR? I think it is more appropriate to make a new PR so I can unify this for other functions as well.

examples/40_advanced/example_single_configuration.py
```diff
@@ -71,7 +67,7 @@
                                                                 configuration=configuration,
                                                                 budget_type='epochs',
                                                                 budget=20,
-                                                                run_time_limit_secs=70
+                                                                run_time_limit_secs=100
```
eddiebergman, line 70: Why was this increased?

autoPyTorch/api/base_task.py
```diff
@@ -229,7 +229,11 @@ def __init__(
                                  " HyperparameterSearchSpaceUpdates got {}".format(type(self.search_space_updates)))
 
     @abstractmethod
```
eddiebergman, line 232: It would be good to mark the class as ABC as it contains methods are now abstractmethods.

```python
class BaseTask(ABC):
    ...
```

autoPyTorch/api/base_task.py
```diff
@@ -240,7 +244,97 @@ def build_pipeline(self, dataset_properties: Dict[str, Any]) -> BasePipeline:
...
+                Testing feature set
+            y_test (Optional[Union[List, pd.DataFrame, np.ndarray]]):
+                Testing target set
+            resampling_strategy (Optional[Union[CrossValTypes, HoldoutValTypes]]):
+                Strategy to split the training data.
+            resampling_strategy_args (Optional[Dict[str, Any]]):
+                arguments required for the chosen resampling strategy. If None, uses
+                the default values provided in DEFAULT_RESAMPLING_PARAMETERS
+                in ```datasets/resampling_strategy.py```.
+            dataset_name (Optional[str], optional):
```
eddiebergman, line 282: Seems you have optional specified twice here

autoPyTorch/api/base_task.py
```diff
@@ -840,10 +934,10 @@ def _search(
             precision (int: default=32):
                 Numeric precision used when loading ensemble data.
                 Can be either '16', '32' or '64'.
-            disable_file_output (Union[bool, List]):
-                If True, disable model and prediction output.
-                Can also be used as a list to pass more fine-grained
-                information on what to save. Allowed elements in the list are:
+            disable_file_output (List[Union[str, DisableFileOutputParameters]]):
```
eddiebergman: This differs from the actual type signature

autoPyTorch/api/base_task.py
```diff
@@ -1233,24 +1349,116 @@ def fit(self,
...
+            dataset_name (Optional[str]):
+                Name of the dataset, if None, random value is used.
+            resampling_strategy (Union[CrossValTypes, HoldoutValTypes]),
+                (default=HoldoutValTypes.holdout_validation):
+                strategy to split the training data.
+            resampling_strategy_args (Optional[Dict[str, Any]]):
+                Arguments required for the chosen resampling strategy. If None, uses
+                the default values provided in DEFAULT_RESAMPLING_PARAMETERS
+                in ```datasets/resampling_strategy.py```.
+            run_time_limit_secs (int: default=120):
```
eddiebergman, line 1366: default is different

autoPyTorch/api/base_task.py
```diff
@@ -1233,24 +1349,116 @@ def fit(self,
...
-                configuration to fit the pipeline with. If None,
-                uses default
+            X_train, y_train, X_test, y_test: Union[np.ndarray, List, pd.DataFrame]
+                A pair of features (X_train) and targets (y_train) used to fit a
+                pipeline. Additionally, a holdout of this pairs (X_test, y_test) can
+                be provided to track the generalization performance of each stage.
+            dataset_name (Optional[str]):
+                Name of the dataset, if None, random value is used.
+            resampling_strategy (Union[CrossValTypes, HoldoutValTypes]),
+                (default=HoldoutValTypes.holdout_validation):
```
eddiebergman, line 1360: It is set to None in the type sig

autoPyTorch/api/tabular_classification.py
```diff
@@ -106,18 +104,82 @@ def __init__(
...
         )
 
-    def build_pipeline(self, dataset_properties: Dict[str, Any]) -> TabularClassificationPipeline:
+    def build_pipeline(
+        self,
+        dataset_properties: Dict[str, Any],
+        include_components: Optional[Dict] = None,
+        exclude_components: Optional[Dict] = None,
+        search_space_updates: Optional[HyperparameterSearchSpaceUpdates] = None
+    ) -> TabularClassificationPipeline:
```
eddiebergman, line 114: perhaps a docstring and some extra typing on Dict, at least to know if keys are `str`, so `Dict[str, Any]`

examples/40_advanced/example_single_configuration.py
ravinkohli: because the configuration did not finish on time. As its a random configuration, its better to take an upper bound

autoPyTorch/evaluation/abstract_evaluator.py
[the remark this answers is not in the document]
eddiebergman: We have this flag so that there is full control over what get's stored for memory purposes. We have a slighlty outdated PR that needs to be updated at somepoint but this is also all stored in memory during calculation which could be optimized.

Essentially, possibility to save memory if required

autoPyTorch/api/base_task.py
ravinkohli: yeah I fixed them

autoPyTorch/evaluation/abstract_evaluator.py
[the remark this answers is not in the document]
ravinkohli: but if you put `disable_file_output=['y_optimization`] the ensemble building fails and instead uses the single best model.

autoPyTorch/api/base_task.py
```diff
@@ -240,7 +245,98 @@ def build_pipeline(self, dataset_properties: Dict[str, Any]) -> BasePipeline:
...
+                Testing feature set
+            y_test (Optional[Union[List, pd.DataFrame, np.ndarray]]):
+                Testing target set
+            resampling_strategy (Optional[Union[CrossValTypes, HoldoutValTypes]]):
+                Strategy to split the training data.
+            resampling_strategy_args (Optional[Dict[str, Any]]):
+                arguments required for the chosen resampling strategy. If None, uses
+                the default values provided in DEFAULT_RESAMPLING_PARAMETERS
+                in ```datasets/resampling_strategy.py```.
+            dataset_name (Optional[str], optional):
```
ravinkohli, line 282:
```suggestion
            dataset_name (Optional[str]):
```

autoPyTorch/evaluation/abstract_evaluator.py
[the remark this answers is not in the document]
eddiebergman: Yup, but it's an option ... don't ask why ... I think the correct thing to do would be to check if is ensemble building is enabled and also check if `y_optimization` is included in `disable_file_output`.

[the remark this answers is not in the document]
ravinkohli: so you mean throw an error in this case?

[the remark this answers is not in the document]
eddiebergman: Yeah, or a warning, whatever you think is best.

test/test_api/test_base_api.py
```diff
@@ -20,6 +20,7 @@
 # ====
 @pytest.mark.parametrize("fit_dictionary_tabular", ['classification_categorical_only'], indirect=True)
 def test_nonsupported_arguments(fit_dictionary_tabular):
+    BaseTask.__abstractmethods__ = set()
```
eddiebergman, line 23: Interesting way to do this, I normally make a `MockBaseTask` class in some `mocks.py`, didn't know you could do this

autoPyTorch/evaluation/utils.py
```diff
@@ -102,3 +103,55 @@ def _predict(self, X: np.ndarray) -> np.ndarray:
...
+        do not save any of the above.
+    """
+    pipeline = 'pipeline'
+    pipelines = 'pipelines'
+    y_optimization = 'y_optimization'
+    y_test = 'y_test'
+    all = 'all'
+
+    @classmethod
+    def check_compatibility(cls, disable_file_output: List) -> None:
```
eddiebergman, line 134: It might also be handy to allow for a single param and not just a `List`

test/test_evaluation/test_utils.py
```diff
@@ -0,0 +1,20 @@
...
+
+def test_disable_file_output_string_error():
+    disable_file_output = ['model']
+    with pytest.raises(ValueError, match=r"Expected .*? to be in the members (.*?) of"
+                                         r" DisableFileOutputParameters or an instance."):
+        DisableFileOutputParameters.check_compatibility(disable_file_output=disable_file_output)
+
+
+def test_disable_file_output_enum_no_error():
+    disable_file_output = [DisableFileOutputParameters.pipeline, DisableFileOutputParameters.pipelines]
```
eddiebergman, line 19: This test and the first one can be combined with a `@pytest.mark.parametrize`

eddiebergman: Not essential, more just a statement

eddiebergman (commented):
It seems there's no mutual exclusivity between `include` and `exclude`, can a user provide both?

I've also started adding some documentation to tests after having some frustration with some older tests in autosklearn.

Have a look at [automl_common/tests/test_backend/test_contexts.py](https://github.com/automl/automl_common/blob/development/test/test_backend/test_context.py). It makes it easier to see what exactly is being tested by a function. While the function name helps, I keep having issues with it not really being descriptive enough.

ravinkohli:
> It seems there's no mutual exclusivity between `include` and `exclude`, can a user provide both?
There is mutual exclusivity, check [this](https://github.com/automl/Auto-PyTorch/blob/40a398743a4bbd6b186844fe46ce71101580b5f1/autoPyTorch/pipeline/components/base_choice.py#L113)
> I've also started adding some documentation to tests after having some frustration with some older tests in autosklearn.
> 
> Have a look at [automl_common/tests/test_backend/test_contexts.py](https://github.com/automl/automl_common/blob/development/test/test_backend/test_context.py). It makes it easier to see what exactly is being tested by a function. While the function name helps, I keep having issues with it not really being descriptive enough.
Sure, we'll start adding descriptions to the tests.

autoPyTorch/api/base_task.py
```diff
@@ -1265,21 +1508,102 @@ def fit(self,
...
-        fit_and_suppress_warnings(self._logger, pipeline, X, y=None)
+        pipeline_options = self.pipeline_options.copy().update(pipeline_options) if pipeline_options is not None \
+            else self.pipeline_options.copy()
 
+        assert pipeline_options is not None
+
+        if budget_type is not None:
+            pipeline_options.update({'budget_type': budget_type})
+        else:
+            budget_type = pipeline_options['budget_type']
```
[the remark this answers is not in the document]
ravinkohli: yeah, we provide default pipeline options with [this](https://github.com/automl/Auto-PyTorch/blob/development/autoPyTorch/configs/default_pipeline_options.json) file.

autoPyTorch/api/base_task.py
```diff
@@ -1265,21 +1508,102 @@ def fit(self,
...
+
+        budget = budget if budget is not None else pipeline_options[budget_type]
```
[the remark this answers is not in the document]
ravinkohli: yeah, actually we use `{'budget_type': 'epochs', 'epochs': 50}` to represent the pipeline options.

examples/40_advanced/example_single_configuration.py
```diff
@@ -0,0 +1,81 @@
...
+X_train, X_test, y_train, y_test = sklearn.model_selection.train_test_split(
+    X, y, test_size=0.5, random_state=3
+)
+
+############################################################################
+# Define an estimator
+# ===================
+
+estimator = TabularClassificationTask(
+    resampling_strategy=HoldoutValTypes.holdout_validation,
```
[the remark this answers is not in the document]
ravinkohli: No, we have `isinstance` checks.

autoPyTorch/evaluation/utils.py
```diff
@@ -102,3 +103,55 @@ def _predict(self, X: np.ndarray) -> np.ndarray:
...
+        for item in disable_file_output:
+            if item not in cls.__members__:
+                if not isinstance(item, cls):
+                    raise ValueError(f"Expected {item} to be in the members ("
+                                     f"{list(cls.__members__.keys())}) of {cls.__name__}"
+                                     f" or an instance.")
+
+    @staticmethod
+    def check_value_in_iterable(container: Iterable, parameter: "DisableFileOutputParameters") -> bool:
```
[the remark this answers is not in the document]
ravinkohli: In the future, if we have to use `enums` with such functionality, I would suggest making our own base class for enums that contain such utility methods.

test/test_api/test_base_api.py
ravinkohli: yeah I read that having `abstractmethods` is what causes an error while instantiating an abstract class. I thought for testing its fine to have this workaround.

autoPyTorch/evaluation/utils.py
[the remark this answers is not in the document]
eddiebergman: Makes sense, resolve whenever

autoPyTorch/evaluation/utils.py
```diff
@@ -102,3 +103,54 @@ def _predict(self, X: np.ndarray) -> np.ndarray:
...
+        for item in disable_file_output:
+            if item not in cls.__members__ and not isinstance(item, cls):
+                raise ValueError(f"Expected {item} to be in the members ("
+                                 f"{list(cls.__members__.keys())}) of {cls.__name__}"
+                                 f" or an instance.")
+
+    @staticmethod
+    def check_value_in_iterable(container: Iterable, parameter: "DisableFileOutputParameters") -> bool:
+        """
+        checks if parameter is in the container either as
```
nabenabe0928, line 144: I think this method should not be staticmethod, but it should look something like:
```python
def is_in_iterable(self, iter: Iterable) -> bool:
    return self in iter or self.value in iter
```

Btw, I think it is more appropriate to use `is_xxx` when we return `bool` and we should use `check_xxx` when we check if inputs are compatible and if not raise error.

nabenabe0928 (commented):
Sorry for the late work, but I will make steady progress whenever I have time

autoPyTorch/evaluation/abstract_evaluator.py
nabenabe0928: Hey, could you separate L843--L867 in the old code as a new method in the upcoming PR?

autoPyTorch/evaluation/abstract_evaluator.py
```diff
@@ -835,19 +850,17 @@ def file_output(
 
         # Abort if we don't want to output anything.
         if hasattr(self, 'disable_file_output'):
```
nabenabe0928, line 852: Define self.disable_file_output rather than just checking, because we somehow do not check `hasattr` below.

```
self.disable_file_output = [] if not hasattr(self, 'disable_file_output') else disable_file_output
```

autoPyTorch/evaluation/abstract_evaluator.py
ravinkohli: sure.

autoPyTorch/evaluation/utils.py
ravinkohli: it can also be `List[str]`. I'll add these.

autoPyTorch/evaluation/utils.py
```diff
@@ -102,3 +103,54 @@ def _predict(self, X: np.ndarray) -> np.ndarray:
...
+    + `pipelines`:
+        In case of cross validation, disables saving the joint model of the
+        pipelines fit on each fold.
+    + `y_test`:
+        do not save the predictions for the test set.
+    + `all`:
+        do not save any of the above.
+    """
+    pipeline = 'pipeline'
+    pipelines = 'pipelines'
```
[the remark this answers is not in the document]
ravinkohli: yeah but let me do that in the other PR, this has already gotten too big. Is that okay?

autoPyTorch/evaluation/utils.py
```diff
@@ -102,3 +103,54 @@ def _predict(self, X: np.ndarray) -> np.ndarray:
...
+    y_optimization = 'y_optimization'
+    y_test = 'y_test'
+    all = 'all'
+
+    @classmethod
+    def check_compatibility(cls, disable_file_output: List) -> None:
+        for item in disable_file_output:
+            if item not in cls.__members__ and not isinstance(item, cls):
```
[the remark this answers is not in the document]
ravinkohli: Actually, as I would also like to support passing a string, so we need the first condition.

autoPyTorch/evaluation/utils.py
```diff
@@ -102,3 +103,54 @@ def _predict(self, X: np.ndarray) -> np.ndarray:
...
+                raise ValueError(f"Expected {item} to be in the members ("
+                                 f"{list(cls.__members__.keys())}) of {cls.__name__}"
+                                 f" or an instance.")
```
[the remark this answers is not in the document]
ravinkohli: yeah I think this should be clearer, I want to support passing an enum (instance of `DisableFileOutputParameters`) as well as passing its corresponding string.

autoPyTorch/evaluation/utils.py
ravinkohli: yeah definitely makes sense, I'll change this.

autoPyTorch/evaluation/train_evaluator.py
```diff
@@ -79,10 +80,25 @@ class TrainEvaluator(AbstractEvaluator):
...
+                optimizes a metric evaluated on the optimization set.
+            + `pipeline`:
+                do not save any individual pipeline files
+            + `pipelines`:
+                In case of cross validation, disables saving the joint model of the
+                pipelines fit on each fold.
+            + `y_test`:
+                do not save the predictions for the test set.
+            + `all`:
+                do not save any of the above.
```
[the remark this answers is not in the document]
ravinkohli: I wanted to have this so it's clear what to pass in disable_file_output from the documentation of the API. It would be easier for the user than to find this somewhere else.

autoPyTorch/evaluation/utils.py
eddiebergman: If subclassing a self made enum class that provides an equality method might make things more intuitive.

Note: I havn't tested this but this is the main idea

```python
class MyEnum(Enum):

    def __eq__(self, other: Any) -> bool:
        if isinstance(other, MyEnum):
            return self == other
        elif isinstance(other, str):
            return self.value = other.value
        else:   
            raise RuntimeError("Some error")

class SubEnum(MyEnum):
    ... inherits __eq__
    

```.

```python
# Dict uses __eq__ in it's __contain__ method
if my_enum_val in my_dict:
    ...
    
# Same goes for list
```

autoPyTorch/evaluation/abstract_evaluator.py
ravinkohli: actually, we are assured that it is a list and we will always have the attribute `disable_file_output` as it is initialised in `__init__`. So, we can remove this `hasattr` check altogether

test/test_utils/test_common.py
```diff
@@ -0,0 +1,31 @@
...
+class SubEnum(autoPyTorchEnum):
+    x = "x"
+    y = "y"
+
+
+@pytest.mark.parametrize('iter',
+                         [[SubEnum.x],
+                          ["x"],
+                          {SubEnum.x: "hello"},
+                          {'x': 'hello'}])
```
nabenabe0928, line 18: I made it immutable just in case and stronger.
```suggestion
                         ([SubEnum.x],
                          ["x"],
                          {SubEnum.x: "hello"},
                          {'x': 'hello'},
                          SubEnum,
                          ["x", "y"]))
```

test/test_utils/test_common.py
```diff
@@ -0,0 +1,31 @@
...
+    can be used with strings.
+
+    Args:
+        iter (Iterable):
+            iterable to check for compaitbility
+    """
+
+    e = SubEnum.x
+
+    assert e in iter
```
nabenabe0928, line 31:
```suggestion
    assert e in iter

class DummyEnum(Enum):  # You need to move it on top
    x = "x"


@pytest.mark.parametrize('iter',
                         [[SubEnum.y],
                          ["y"],
                          {SubEnum.y: "hello"},
                          {'y': 'hello'}])
def test_autopytorch_enum_false(iter):
    """
    This test ensures that a subclass of `autoPyTorchEnum`
    can be used with strings.
    Args:
        iter (Iterable):
            iterable to check for compaitbility
    """

    e = SubEnum.x

    assert e not in iter


@pytest.mark.parametrize('others', (1, 2.0, SubEnum, DummyEnum.x))
def test_raise_errors_autopytorch_enum(others):
    """
    This test ensures that a subclass of `autoPyTorchEnum`
    raises error properly.
    Args:
        others (Any):
            Variable to compare with SubEnum.
    """

    with pytest.raises(RuntimeError):
        SubEnum.x == others


```

autoPyTorch/utils/common.py
```diff
@@ -75,6 +76,25 @@ def __str__(self) -> str:
...
 
+class autoPyTorchEnum(str, Enum):
+    """
+    Utility class for enums in autoPyTorch.
+    Allows users to use strings, while we internally use
+    this enum
+    """
+    def __eq__(self, other: Any) -> bool:
+        if isinstance(other, autoPyTorchEnum):
+            return type(self) == type(other) and self.value == other.value
```
nabenabe0928, line 87: Somehow this line is not passed through when you pass a list that contains the identical enum member.
For example, when you replace L87 with `raise NotImplementedError`, `APTEnum.x in [APTEnum.x]` should raise an error, but it does not.

Note that it raises an error when you do `APTEnum.x in [APTEnum.y]` or `APTEnum.x == APTEnum.x`.
Probably, `__contains__` method in `enum` or `str` or `list` is causing the issue, but I could not find out easily.

nabenabe0928 (commented):
Hi, really sorry for the late response.
I will check everything as soon as possible

Files to check:
1. base_api.py
2. test_api.py

autoPyTorch/utils/common.py
```diff
@@ -75,6 +76,25 @@ def __str__(self) -> str:
             self.hyperparameter, self.value_range, self.default_value, self.log)
 
 
+class autoPyTorchEnum(str, Enum):
+    """
+    Utility class for enums in autoPyTorch.
+    Allows users to use strings, while we internally use
+    this enum
+    """
+    def __eq__(self, other: Any) -> bool:
```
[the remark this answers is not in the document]
eddiebergman: Any particular reason to favor `object` over `Any`? `Any` is the more common type as it directly comes from the`typing` module.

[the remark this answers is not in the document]
nabenabe0928: This is a suggestion from mypy. But I do not mind either way.

autoPyTorch/api/base_task.py
```diff
@@ -233,18 +237,132 @@ def __init__(
...
+                name of the dataset, used as experiment name.
+
+        Returns:
+            BaseDataset:
+                the dataset object
+            BaseInputValidator:
+                fitted input validator
+        """
+        raise NotImplementedError("Function called on BaseTask, this can only be called by "
+                                  "specific task which is a child of the BaseTask")
```
nabenabe0928, line 318: You do not need this error message, because we cannot instantiate it without overriding abstract methods.

autoPyTorch/api/base_task.py
```diff
@@ -233,18 +237,132 @@ def __init__(
...
+
+    def get_dataset(
```
nabenabe0928, line 320: `_get_dataset`?

nabenabe0928 (commented):
minor comments

autoPyTorch/api/base_task.py
```diff
@@ -1237,24 +1384,119 @@ def fit(self,
...
+                                       y_test=y_test,
+                                       resampling_strategy=resampling_strategy,
+                                       resampling_strategy_args=resampling_strategy_args,
+                                       dataset_name=dataset_name
+                                       )
+
+        # TAE expects each configuration to have a config_id.
+        # For fitting a pipeline as it is not part of the
+        # search process, it makes sense to set it to 0
+        if hasattr(configuration, 'config_id') or configuration.config_id is None:
```
nabenabe0928, line 1498: or, but not and?
I do not get why it makes sense to overwrite `config_id`.

autoPyTorch/api/base_task.py
```diff
@@ -1237,24 +1384,119 @@ def fit(self,
...
+            (BaseDataset):
+                Dataset created from the given tensors
         """
-        self.dataset_name = dataset.dataset_name
 
-        if self._logger is None:
-            self._logger = self._get_logger(str(self.dataset_name))
+        if dataset is None:
+            assert X_train is not None and \
+                   y_train is not None, "No dataset provided, must provide X_train, y_train tensors"
```
nabenabe0928, line 1485:
```suggestion
            if X_train is None or y_train is None:
                raise ValueError("X_train and y_train must be provided")
```

nabenabe0928 (changes requested):

autoPyTorch/api/base_task.py
```diff
@@ -1265,21 +1507,102 @@ def fit(self,
...
+            # we expect it to be not None. This is for mypy
+            assert dataset.dataset_name is not None
+            self._logger = self._get_logger(dataset.dataset_name)
+
+        if include_components is None:
+            include_components = self.include_components
+        if exclude_components is None:
+            exclude_components = self.exclude_components
+        if search_space_updates is None:
+            search_space_updates = self.search_space_updates
```
nabenabe0928, line 1521:
```suggestion
        include_components = self.include_components if include_components is None else include_components
        exclude_components = self.exclude_components if exclude_components is None else exclude_components
        search_space_updates = self.search_space_updates if search_space_updates is None else search_space_updates
```

autoPyTorch/api/base_task.py
```diff
@@ -1265,21 +1507,102 @@ def fit(self,
...
+
+        scenario_mock = unittest.mock.Mock()
+        scenario_mock.wallclock_limit = run_time_limit_secs
+        # This stats object is a hack - maybe the SMAC stats object should
+        # already be generated here!
+        stats = Stats(scenario_mock)
+
+        if memory_limit is None:
+            if hasattr(self, '_memory_limit') and self._memory_limit is not None:
+                memory_limit = self._memory_limit
```
nabenabe0928, line 1531:
```suggestion
        if memory_limit is None and getattr(self, '_memory_limit', None) is not None:
            memory_limit = self._memory_limit
```

autoPyTorch/evaluation/abstract_evaluator.py
```diff
@@ -834,20 +849,17 @@ def file_output(
...
+        if 'all' in self.disable_file_output:
+            return None, {}
 
         # This file can be written independently of the others down bellow
-        if 'y_optimization' not in self.disabled_file_outputs:
+        if 'y_optimization' not in self.disable_file_output:
             if self.output_y_hat_optimization:
                 self.backend.save_targets_ensemble(self.Y_optimization)
 
         if hasattr(self, 'pipelines') and self.pipelines is not None:
```
nabenabe0928, line 860:
```suggestion
        if getattr(self, 'pipelines', None) is not None:
```

autoPyTorch/evaluation/abstract_evaluator.py
```diff
@@ -861,7 +873,7 @@ def file_output(
             pipelines = None
 
         if hasattr(self, 'pipeline') and self.pipeline is not None:
```
nabenabe0928, line 875:
```suggestion
        if getattr(self, 'pipeline', None) is not None:
```

autoPyTorch/api/base_task.py
```diff
@@ -1265,21 +1507,102 @@ def fit(self,
...
-        # initialise fit dictionary
-        X = self._get_fit_dictionary(
-            dataset_properties=dataset_properties,
-            dataset=dataset,
-            split_id=split_id)
+        pipeline_options = self.pipeline_options.copy().update(pipeline_options) if pipeline_options is not None \
+            else self.pipeline_options.copy()
 
-        fit_and_suppress_warnings(self._logger, pipeline, X, y=None)
+        assert pipeline_options is not None
```
nabenabe0928, line 1540: It does not make any bad effects other than on readability, but do we need it?

autoPyTorch/api/base_task.py
```diff
@@ -1265,21 +1507,102 @@ def fit(self,
...
+        if budget_type is not None:
+            pipeline_options.update({'budget_type': budget_type})
+        else:
+            budget_type = pipeline_options['budget_type']
+
+        budget = budget if budget is not None else pipeline_options[budget_type]
+
+        if disable_file_output is None:
+            disable_file_output = self._disable_file_output if hasattr(self, '_disable_file_output') \
+                and self._disable_file_output is not None else []
```
nabenabe0928, line 1551:
```suggestion
            disable_file_output = getattr(self, '_disable_file_output', [])
```

autoPyTorch/api/base_task.py
```diff
@@ -1265,21 +1507,102 @@ def fit(self,
...
+
+        run_info, run_value = tae.run_wrapper(
+            RunInfo(config=configuration,
+                    budget=budget,
+                    seed=self.seed,
+                    cutoff=run_time_limit_secs,
+                    capped=False,
+                    instance_specific=None,
+                    instance=None)
+        )
```
nabenabe0928, line 1583: Can we separate this process as one method?

autoPyTorch/api/base_task.py
```diff
@@ -1265,21 +1507,102 @@ def fit(self,
...
+                else:
+                    load_function = self._backend.load_model_by_seed_and_id_and_budget
+                fitted_pipeline = load_function(
+                    seed=self.seed,
+                    idx=run_info.config.config_id + tae.initial_num_run,
+                    budget=float(run_info.budget),
+                )
+        else:
+            warnings.warn(f"Fitting pipeline failed with status: {run_value.status}"
+                          f", additional_info: {run_value.additional_info}")
```
nabenabe0928, line 1602: Can we transplant?
Where should `self._clean_logger()` come? you can include it here if you think you need it here.
```python
    def _get_fitted_pipeline(...) -> ...:
        if run_value.status != StatusType.SUCCESS:
            warnings.warn(f"Fitting pipeline failed with status: {run_value.status}"
                          f", additional_info: {run_value.additional_info}")
            return None
        elif any(disable_file_output for c in ['all', 'pipeline']):
            self._logger.warning("File output is disabled. No pipeline can returned")
            return None

        if self.resampling_strategy in CrossValTypes:
            load_function = self._backend.load_cv_model_by_seed_and_id_and_budget
        else:
            load_function = self._backend.load_model_by_seed_and_id_and_budget

        return load_function(
            seed=self.seed,
            idx=run_info.config.config_id + tae.initial_num_run,
            budget=float(run_info.budget),
        )

```

nabenabe0928 (changes requested):

nabenabe0928:
I will check `test_api.py` after the comments are resolved.

autoPyTorch/utils/common.py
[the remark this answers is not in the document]
ravinkohli: I think I'd prefer to keep it as `Any`

autoPyTorch/utils/common.py
```diff
@@ -75,6 +76,25 @@ def __str__(self) -> str:
...
+        if isinstance(other, autoPyTorchEnum):
+            return type(self) == type(other) and self.value == other.value
+        elif isinstance(other, str):
+            return bool(self.value == other)
+        else:
+            raise RuntimeError(f"Unsupported type {type(other)}."
+                               f"{self} only supports `str` and `{self}`")
```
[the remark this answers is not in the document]
ravinkohli: I dont think we need `__mro__`. but I have added the name to the error output

autoPyTorch/api/base_task.py
ravinkohli: `config_id` is updated inside the `ConfigSpace.Configuration` class when `smac` queries a new config. However, in this case the configuration is coming from the user and it does not have a config id so we need to set it. In the case of fitting a pipeline we really dont care what the config id is, so setting it `0` seems fine to me.

nabenabe0928: I got what you meant in the comment out now 
Then we don't need the `if` itself?
Or does that raise any error?

ravinkohli: Actually, I dont think we need the `if` statement. I'll try the tests without the `if` statement and if its not necessary I'll remove it.

nabenabe0928: Thanks, it will look more clear to me

autoPyTorch/api/base_task.py
ravinkohli: sure.

autoPyTorch/api/base_task.py
ravinkohli: I think it's not going to be used anywhere else in the `BaseTask` and moreover, it's just 2 statements. I would prefer it to be as it is.

autoPyTorch/api/base_task.py
ravinkohli: I would like it to be a user-facing function.

autoPyTorch/api/base_task.py
ravinkohli: Actually, `get_dataset` expects there to be `X_train` and `y_train`, python itself will raise an error. I have this check because we allow the user to either send a dataset object or `X_train`, etc. 

I have done so because in case the user wants to get the default configuration or sample a configuration like [here](https://github.com/ravinkohli/Auto-PyTorch/blob/a0fef7700fd0be8696db1c8a477d4ddedf4f6754/examples/40_advanced/example_single_configuration.py#L59), they will have to create their own dataset object and hence they can use it to fit a pipeline. I have also allowed the option for `X_train`, etc. in case the user would like to fit an incumbent configuration after a search, they won't need to instantiate the dataset object themselves.

autoPyTorch/api/base_task.py
```diff
@@ -1237,24 +1384,119 @@ def fit(self,
...
+                Dataset created from the given tensors
         """
-        self.dataset_name = dataset.dataset_name
 
-        if self._logger is None:
-            self._logger = self._get_logger(str(self.dataset_name))
+        if dataset is None:
+            assert X_train is not None and \
+                   y_train is not None, "No dataset provided, must provide X_train, y_train tensors"
+            dataset = self.get_dataset(X_train=X_train,
```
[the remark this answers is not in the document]
ravinkohli: This would be possible if I take you suggestion of moving the check for `X_train` inside `get_dataset` but as explained in that conversation, we can encapsulate these lines to a different function.

autoPyTorch/api/base_task.py
ravinkohli: it's for `mypy`

autoPyTorch/utils/common.py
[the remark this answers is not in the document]
nabenabe0928: You are right, when we specify the choice in enum, we do not need `__mro__`

autoPyTorch/api/base_task.py
nabenabe0928: Yeah, if it is literally `get_dataset`, it is better to stick to your implementation, thanks for the explanation.
But could you change assert to raise error?

autoPyTorch/api/base_task.py
```diff
@@ -1223,10 +1351,29 @@ def refit(
 
         return self
 
-    def fit(self,
-            dataset: BaseDataset,
-            pipeline_config: Optional[Configuration] = None,
-            split_id: int = 0) -> BasePipeline:
+    def fit_pipeline(
+        self,
+        configuration: Configuration,
```
nabenabe0928, line 1355: Since we can take both `xxx_train` and `dataset`, it is better to force users to pass variables by keywords.
To achieve it, we need the `*` before the beginning of the keywords.
```suggestion
    def fit_pipeline(
        self,
        configuration: Configuration,
        *,
```

nabenabe0928 (changes requested):
Thanks for the commits.
Considering your comment, I added an additional suggestion.
It is better to have it for stable behavior from user side.

autoPyTorch/api/base_task.py
nabenabe0928: I know it is used only once, but since this process consumes many lines because of their arguments, but yeah it seems the method only for this purpose has a similar number of arguments...

autoPyTorch/api/base_task.py
ravinkohli: sure.

autoPyTorch/api/base_task.py
ravinkohli: Okay, I have implemented **dataset_kwargs

nabenabe0928: No no, you can just add * as in the suggestion and that works nicer than kwargs which hides information

ravinkohli: oh okay, I got it. Thanks, I actually did not know about this functionality.

Issue #350, opened 2021-12-01
plot_over_time might cause error in some environments


nabenabe0928:
NOTE: ISSUES ARE NOT FOR CODE HELP - Ask for Help at https://stackoverflow.com

Your issue may already be reported!
Also, please search on the [issue tracker](../) before creating one.

* **I'm submitting a ...**
- bug report
  
## Expected Behavior
Users can use `plot_over_time` anytime without any bugs.

## Current Behavior
`plt.show` might cause some issues especially in the environment where you are supposed to use non-GUI operations. 

## Possible Solution
Modify `results_visualizer` so that it can take a `savefig` option

nabenabe0928:
solved by PR#351.

PR #351, opened 2021-12-01
[feat] Add the option to save a figure in plot setting params

nabenabe0928:
Since non-GUI based environments would like to avoid the usage of
show method in the matplotlib, I added the option to savefig and
thus users can complete the operations inside AutoPytorch.

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/autoPyTorch/utils/results_visualizer.py
+++ b/autoPyTorch/utils/results_visualizer.py
@@ -1,6 +1,6 @@
 from dataclasses import dataclass
 from enum import Enum
-from typing import Any, Dict, Optional, Tuple
+from typing import Any, Dict, NamedTuple, Optional, Tuple
 
 import matplotlib.pyplot as plt
 
@@ -71,8 +71,7 @@ class ColorLabelSettings:
         return colors, labels
 
 
-@dataclass(frozen=True)
-class PlotSettingParams:
+class PlotSettingParams(NamedTuple):
     """
     Parameters for the plot environment.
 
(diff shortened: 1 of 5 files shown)


autoPyTorch/utils/results_visualizer.py
```diff
@@ -71,8 +71,7 @@ def extract_dicts(
         return colors, labels
 
 
-@dataclass(frozen=True)
-class PlotSettingParams:
+class PlotSettingParams(NamedTuple):
```
nabenabe0928, line 74: This change is due to the fact that `dataclass` does not allow to have `dict` or `list`.

autoPyTorch/utils/results_visualizer.py
```diff
@@ -201,17 +220,22 @@ def _set_plot_args(
...
+            ax.legend(**plot_setting_params.legend_kwargs)
 
         if plot_setting_params.title is not None:
-            ax.set_title(plot_setting_params.title)
-        if plot_setting_params.show:
+            ax.set_title(plot_setting_params.title, **plot_setting_params.title_kwargs)
+
+        if plot_setting_params.figname is not None:
+            plt.savefig(plot_setting_params.figname, **plot_setting_params.savefig_kwargs)
+        elif plot_setting_params.show:
```
ravinkohli, line 238: This means that show and figname would be mutually exclusive. Like we won't show the figure if there is a figure name even if `show==True`. Do you think we should do something like this:
```suggestion
        if plot_setting_params.show:
```

ravinkohli: ohh I see that is intended.

ravinkohli (approved):

nabenabe0928:
[issue#350](https://github.com/automl/Auto-PyTorch/issues/350) was solved here.

PR #363, opened 2021-12-15
Update workflow files

eddiebergman:
Updates the workflow files to be more similar to `auto-sklearn`, `ConfigSpace` and `SMAC`

* Adds some much needed spacing between steps of some files
* Fixes the double running of tests on every push to a PR
* Remove `scheduled_test.yaml` and added a `schedule` of daily to `pytest.yaml`
* Added `manual_dispatch` to some of the workflows (Enables manual triggering from the github UI)
* Adds a name to checkout step which didn't previously
* Added Python 3.10 to list of python versions tested. EDIT: This is excluded explicitly with the reason stated that Torch does not support python 3.10 yet
* Added windows and mac to `pytest.yaml` but it's currently disabled until we figure out how to run them.
* Added install from dist as one of the tests
* Gave name to each version test in `pytest.yaml`
* Moved testing args of `pytest` and `pytest-cov` as env variables


.github/workflows/dist.yml
```diff
@@ -1,31 +1,57 @@
...
+      - master
+      - development
+
+  # When a push occurs on a PR that targets these branches
+  pull_request:
+    branches:
+      - master
+      - development
+
+  schedule:
```
ravinkohli, line 19: actually why would you like this workflow on a schedule?

eddiebergman: There's no real need, there's also no real need to have it on anything but push to master or development. This workflow actually does very little compared to `pytest.yaml`, which is also build a dist version and then runs tests on it. The only difference here is that it runs [`twine check`](https://twine.readthedocs.io/en/latest/#twine-check) which checks that the long description will parse correctly on PyPi, that's it.

I would probably remove this workflow entirely upon second though.

eddiebergman: Also, I think the mypy compliance thing is wrong, I originally submitted a PR to autosklearn to check for it which I think got lost in translation to the workflow files in all our repos. It should be `mypy -c "import autosklearn"`

eddiebergman: Fixed the compliance check

ravinkohli (approved):
Thanks for this PR. I think it can be merged.

.github/workflows/dist.yml
```diff
@@ -1,31 +1,61 @@
...
     - name: Install dist
       run: |
         last_dist=$(ls -t dist/autoPyTorch-*.tar.gz | head -n 1)
         pip install $last_dist
+
     - name: PEP 561 Compliance
       run: |
         pip install mypy
+
         cd ..  # required to use the installed version of autosklearn
```
ravinkohli, line 57:
```suggestion
        cd ..  # required to use the installed version of autoPyTorch
```

eddiebergman:
Once the tests pass I think this can be merged

Issue #349, opened 2021-11-30
[Bug] Cross validation yields unexpected additional info


nabenabe0928:
# Issue Description
It seems abstract evaluator does not save cross-validated performance for opt/test datasets.
Also train_evaluator.py is too messy and requires major refactoring.

## Possible Solution
Modify `calculate_auxiliary_losses` in `abstract_evaluator.py`

## Your Code

When using the new results_manager, the following yields the issue.
(Although I fixed it)
```
python -m pytest test/test_api/test_api.py::test_tabular_classification[CrossValTypes.k_fold_cross_validation-resampling_strategy_args1-40981] -W ignore
```

PR #365, opened 2021-12-22
[fix] Add first draft of the PR for issue#349

nabenabe0928:
## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Checklist:

* [ ] Have you written new tests for your core changes, as applicable?

## Description
See [issue#349](https://github.com/automl/Auto-PyTorch/issues/349).
Note that although this PR has so many changes, most changes are gathered in three files:
1. tae.py
2. evaluator.py
3. abstract_evaluator.py

The changes in other files are mostly for tests or deletions because I integrated those features into the aforementioned three files.


autoPyTorch/evaluation/tae.py
```diff
@@ -62,34 +91,150 @@ def fit_predict_try_except_decorator(
...
+    if budget == 0 and status == StatusType.DONOTADVANCE:
+        status = StatusType.SUCCESS
+
+    if not isinstance(additional_run_info, dict):
+        additional_run_info = {'message': additional_run_info}
+
+    return cost, status, info, additional_run_info
+
+
+class TargetAlgorithmQuery(AbstractTAFunc):
```
ravinkohli, line 237: I think `ExecuteTAFuncWithQueue` was more appropriate as a name. Also, it keeps `AutoPyTorch` in line with `SMAC` and `auto-sklearn`.  Let's keep it that way.

autoPyTorch/evaluation/tae.py
```diff
@@ -110,90 +255,66 @@ def __init__(
         stats: Optional[Stats] = None,
         run_obj: str = 'quality',
         par_factor: int = 1,
-        output_y_hat_optimization: bool = True,
+        save_y_opt: bool = True,
```
ravinkohli, line 258: actually, I think `save_y_ensemble_optimization` or even `output_y_ensemble_optimization` as this flag also acts as a way to disable ensemble construction. As if it is `False` we cant build an ensemble. We can also add this to the description of the variable. Actually, it has got nothing to do with saving `y_opt`, these `true_targets_ensemble` are only used for ensemble construction

autoPyTorch/evaluation/tae.py
```diff
@@ -110,90 +255,66 @@ def __init__(
...
             abort_on_first_run_crash=abort_on_first_run_crash,
         )
 
-        self.backend = backend
+        self.fixed_pipeline_params = FixedPipelineParams.with_default_pipeline_config(
+            pipeline_config=pipeline_config,
+            backend=backend,
+            seed=seed,
+            metric=metric,
+            save_y_opt=save_y_opt,
```
ravinkohli, line 286: same here. please change the name.

autoPyTorch/evaluation/tae.py
```diff
@@ -242,86 +355,81 @@ def run_wrapper(
...
             capture_output=True,
             context=context,
         )
 
-        if isinstance(config, (int, str)):
-            num_run = self.initial_num_run
-        else:
-            num_run = config.config_id + self.initial_num_run
+        search_space_updates = self.fixed_pipeline_params.search_space_updates
+        self.logger.debug(f"Search space updates for {num_run}: {search_space_updates}")
```
ravinkohli, line 394: I dont think we need to specially create a variable just to log out the `search_space_updates`. I added it because I wanted to see if they are being passed correctly. especially when they will be printed for each run, when in fact they are the same for all num_runs.

ravinkohli, line 394: I dont think we need to specially create a variable just to log out the `search_space_updates`. I added it because I wanted to see if they are being passed correctly. especially when they will be printed for each run, when in fact they are the same for all num_runs.
```suggestion
```

autoPyTorch/evaluation/tae.py
```diff
@@ -62,34 +91,150 @@ def fit_predict_try_except_decorator(
...
+    return cost, status, info, additional_run_info
+
+
+def _process_exceptions(
+    obj: PynisherFunctionWrapperType,
+    queue: Queue,
+    budget: float,
+    worst_possible_result: float
+) -> ProcessedResultsType:
+    if obj.exit_status is TAEAbortException:
```
ravinkohli, line 193: I think this way of implementing the logic is a bit confusing. Could you rather maybe handle the exception first and then handle creating the `additional_run_info` based on that? Mainly I am struggling to find the use of `is_anything_exception`. Moreover, I also think that you are missing some info for example, `info_for_empty` in case of a `MEMOUT` previously also contained the memory limit as well. I'd suggest you to add both the `memory_limit` and `func_eval_time` which is available in `run_info.cutoff`.

autoPyTorch/api/base_task.py
```diff
@@ -43,8 +43,9 @@
 from autoPyTorch.datasets.resampling_strategy import CrossValTypes, HoldoutValTypes
 from autoPyTorch.ensemble.ensemble_builder import EnsembleBuilderManager
 from autoPyTorch.ensemble.singlebest_ensemble import SingleBest
-from autoPyTorch.evaluation.abstract_evaluator import fit_and_suppress_warnings
-from autoPyTorch.evaluation.tae import ExecuteTaFuncWithQueue, get_cost_of_crash
+from autoPyTorch.evaluation.abstract_evaluator import fit_pipeline
+from autoPyTorch.evaluation.pipeline_class_collection import get_default_pipeline_config
+from autoPyTorch.evaluation.tae import TargetAlgorithmQuery
```
ravinkohli, line 48: Also change back to `ExecuteTaFuncWithQueue`.

autoPyTorch/api/base_task.py
```diff
@@ -669,22 +670,23 @@ def _do_dummy_prediction(self) -> None:
...
             initial_num_run=num_run,
+            pipeline_config=get_default_pipeline_config(choice='dummy'),
             stats=stats,
             memory_limit=memory_limit,
             disable_file_output=self._disable_file_output,
             all_supported_metrics=self._all_supported_metrics
         )
 
-        status, _, _, additional_info = ta.run(num_run, cutoff=self._time_for_task)
+        status, _, _, additional_info = taq.run(num_run, cutoff=self._time_for_task)
```
ravinkohli, line 689: if you dont like `ta` you can maybe use `tae` which stands for target algorithm execution.

autoPyTorch/api/base_task.py
```diff
@@ -966,13 +968,13 @@ def _search(
                 information on what to save. Must be a member of `DisableFileOutputParameters`.
                 Allowed elements in the list are:
 
-                + `y_optimization`:
+                + `y_opt`:
                     do not save the predictions for the optimization set,
                     which would later on be used to build an ensemble. Note that SMAC
                     optimizes a metric evaluated on the optimization set.
-                + `pipeline`:
+                + `model`:
```
ravinkohli, line 975: I prefer the name `pipeline`.

autoPyTorch/api/base_task.py
```diff
@@ -966,13 +968,13 @@ def _search(
...
                     do not save any individual pipeline files
-                + `pipelines`:
+                + `cv_model`:
```
ravinkohli, line 977: you can use `cv_pipeline` instead.

autoPyTorch/api/base_task.py
```diff
@@ -1060,7 +1062,7 @@ def _search(
 
         # Here the budget is set to max because the SMAC intensifier can be:
         # Hyperband: in this case the budget is determined on the fly and overwritten
-        #            by the ExecuteTaFuncWithQueue
+        #            by the TargetAlgorithmQuery
```
ravinkohli, line 1065: change here as well.

autoPyTorch/api/base_task.py
```diff
@@ -1344,7 +1346,7 @@ def refit(
                 dataset_properties=dataset_properties,
                 dataset=dataset,
                 split_id=split_id)
-            fit_and_suppress_warnings(self._logger, model, X, y=None)
+            fit_pipeline(self._logger, model, X, y=None)
```
ravinkohli, line 1349: I think the previous name emphasised the fact that we are suppressing warnings, otherwise, we could have just used `model.fit(X, y)`. So could you change this name back to what it was? Or I dont mind `fit_pipeline_supress_warnings`

autoPyTorch/evaluation/abstract_evaluator.py
```diff
@@ -2,376 +2,138 @@
...
+    status: StatusType
+    pipeline: Optional[BaseEstimator] = None
+    valid_pred: Optional[np.ndarray] = None
+    test_pred: Optional[np.ndarray] = None
+    additional_run_info: Optional[Dict] = None
 
-    The provided configuration determines the type of pipeline created. For more
-    details, please read the get_pipeline() method.
 
+class FixedPipelineParams(NamedTuple):
```
ravinkohli, line 114: What do you mean by "fixed" pipeline params?

autoPyTorch/evaluation/abstract_evaluator.py
```diff
@@ -2,376 +2,138 @@
...
+    """
     Attributes:
         backend (Backend):
-            An object that allows interaction with the disk storage. In particular, allows to
+            An object to interface with the disk storage. In particular, allows to
             access the train and test datasets
-        queue (Queue):
-            Each worker available will instantiate an evaluator, and after completion,
-            it will append the result to a multiprocessing queue
         metric (autoPyTorchMetric):
```
ravinkohli, line 120:
```suggestion
        optimize_metric (autoPyTorchMetric):
```

autoPyTorch/evaluation/tae.py
nabenabe0928: Since this is not urgent, I will just reply to this message and another for now.

Although I agree that `TargetAlgorithmQuery` is not compatible with `auto-sklearn`, this name is still compatible with `SMAC` and the name itself is more appropriate because of the following problems in the previous name.

The problems of `ExecuteTAFuncWithQueue` are that:
1. the name starts from a verb, which is confusing for a class name except the class is callable,
2. the name itself is not correct because we **execute TA func** when we **query** it from **Queue**, but we never execute TA func with Queue,
3. instances of this class are stored in Queue as a query, so it will not be executed when it is instantiated

It is impossible to merge `BaseTask` with the one in ASK, so I did not feel the compatibility with ASK in `TAE` is necessary, but do you think we need it?

autoPyTorch/api/base_task.py
nabenabe0928: This is another question that is mentioned in the other comment.

This is maybe a stupid question, but why do we need to emphasize `suppress warnings`?
I understand it if we squeeze errors during the fit, but warnings will not be critical for running a script and we write them in logging anyways.
So it is more like `fit_pipeline_with_logging_warnings`?

PR #369, opened 2022-01-25
[FIX] Enable preprocessing in reg_cocktails

ravinkohli:
## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

* [ ] Have you written new tests for your core changes, as applicable?

## Description

Following the discussion in last weeks meeting, we would like to have preprocessing as part of the pipeline in regularisation cocktails. Therefore this PR makes the necessary changes to achieve that. These include uncommenting the various preprocessor transforms as well as code for learning embeddings. Moreover, this PR also removes the `is_small_preprocess` attribute from `BaseDataset` as we will always preprocess early in the `EarlyPreprocessing` node.

## Motivation and Context

It is required to allow searching for preprocessing hyperparameters in the reg_cocktails branch. Also, to avoid redundant preprocessing of the dataset for each epoch, we would like to preprocess the whole data all at once. 

## How has this been tested?

Previous tests for preprocessing and early preprocessing have been enabled. Also tests for embedding module have been enabled to check if embedding is compatible with the preprocessing approach


test/test_pipeline/components/preprocessing/test_scalers.py
```diff
@@ -175,7 +175,7 @@ def test_minmax_scaler(self):
 
 
 # TODO: fix in preprocessing PR
-@unittest.skip("Skipping tests as preprocessing is not finalised")
+# @unittest.skip("Skipping tests as preprocessing is not finalised")
```
nabenabe0928, line 178: Can we remove it rather than comment out because it is not relevant anymore?

autoPyTorch/pipeline/components/training/data_loader/image_data_loader.py
```diff
@@ -63,5 +63,5 @@ def _check_transform_requirements(self, X: Dict[str, Any], y: Any = None) -> Non
         if not X['image_augmenter'] and 'image_augmenter' not in X:
             raise ValueError("Cannot find the image_augmenter in the fit dictionary")
 
-        if not X['dataset_properties']['is_small_preprocess'] and 'preprocess_transforms' not in X:
-            raise ValueError("Cannot find the preprocess_transforms in the fit dictionary")
+        # if not X['dataset_properties']['is_small_preprocess'] and 'preprocess_transforms' not in X:
```
nabenabe0928, line 66: Can we continue this test by just removing `not X['dataset_properties']['is_small_preprocess']`?

autoPyTorch/pipeline/components/training/data_loader/feature_data_loader.py
```diff
@@ -93,5 +93,4 @@ def _check_transform_requirements(self, X: Dict[str, Any], y: Any = None) -> Non
                 mechanism, in which during a transform, a components adds relevant information
                 so that further stages can be properly fitted
         """
-        if not X['dataset_properties']['is_small_preprocess'] and 'preprocess_transforms' not in X:
```
nabenabe0928, line 96: Can we continue this test by just removing `not X['dataset_properties']['is_small_preprocess']`?

autoPyTorch/pipeline/components/training/data_loader/base_data_loader.py
```diff
@@ -226,9 +226,9 @@ def check_requirements(self, X: Dict[str, Any], y: Any = None) -> None:
         if 'backend' not in X:
             raise ValueError("backend is needed to load the data from disk")
 
-        if 'is_small_preprocess' not in X['dataset_properties']:
-            raise ValueError("is_small_pre-process is required to know if the data was preprocessed"
-                             " or if the data-loader should transform it while loading a batch")
+        # if 'is_small_preprocess' not in X['dataset_properties']:
+        #     raise ValueError("is_small_pre-process is required to know if the data was preprocessed"
+        #                      " or if the data-loader should transform it while loading a batch")
```
nabenabe0928, line 231: Can we remove it completely?

autoPyTorch/pipeline/components/training/data_loader/base_data_loader.py
```diff
@@ -57,7 +57,8 @@ def __init__(self, batch_size: int = 64,
         self.add_fit_requirements([
             FitRequirement("split_id", (int,), user_defined=True, dataset_property=False),
             FitRequirement("Backend", (Backend,), user_defined=True, dataset_property=False),
-            FitRequirement("is_small_preprocess", (bool,), user_defined=True, dataset_property=True)])
+            # FitRequirement("is_small_preprocess", (bool,), user_defined=True, dataset_property=True)
```
nabenabe0928, line 60: I do not know what this `is_small_preprocess` does, but can we remove it if we will not use it anymore?

autoPyTorch/pipeline/components/setup/early_preprocessor/EarlyPreprocessing.py
```diff
@@ -20,7 +20,7 @@ def __init__(self, random_state: Optional[np.random.RandomState] = None) -> None
         super().__init__()
```
nabenabe0928, line 20: Do you know why those components files have capital letters in the file names?
Just asking, but I am not requiring the fixes in this PR.

autoPyTorch/datasets/base_dataset.py
```diff
@@ -146,7 +146,7 @@ def __init__(
 
         # TODO: Look for a criteria to define small enough to preprocess
         # False for the regularization cocktails initially
-        self.is_small_preprocess = False
+        # self.is_small_preprocess = False
```
nabenabe0928, line 149: What is `small_preprocess`?

autoPyTorch/pipeline/components/setup/network_backbone/base_network_backbone.py
```diff
@@ -28,7 +28,7 @@ def __init__(self,
                  **kwargs: Any):
         super().__init__()
         self.add_fit_requirements([
-            FitRequirement('is_small_preprocess', (bool,), user_defined=True, dataset_property=True),
+            # FitRequirement('is_small_preprocess', (bool,), user_defined=True, dataset_property=True),
```
nabenabe0928, line 31: Can we remove it?

autoPyTorch/pipeline/components/setup/network_backbone/base_network_backbone.py
```diff
@@ -52,12 +52,7 @@ def fit(self, X: Dict[str, Any], y: Any = None) -> BaseEstimator:
         self.check_requirements(X, y)
         X_train = X['X_train']
```
nabenabe0928, line 53: Can we add a typing for this variable `X_train`?

autoPyTorch/pipeline/components/setup/network_embedding/base_network_embedding.py
```diff
@@ -31,22 +31,21 @@ def transform(self, X: Dict[str, Any]) -> Dict[str, Any]:
...
+                named_transformers_['numerical_pipeline']
+            num_numerical_columns = numerical_column_transformer.transform(
+                X_train[:, X['dataset_properties']['numerical_columns']]).shape[1]
+        num_input_features = np.zeros((num_numerical_columns + len(X['dataset_properties']['categorical_columns'])),
+                                      dtype=int)
+        categories = X['dataset_properties']['categories']
+        
+        for i, category in enumerate(categories):
+            num_input_features[num_numerical_columns + i, ] = len(category)
+        return num_numerical_columns, num_input_features
```
nabenabe0928, line 51: **COMMENT**

Shouldn't `num_input_feats`  be initialized by `np.ones` because numerical columns has 1 input feature?
```suggestion
        num_cols = num_numerical_columns + len(X['dataset_properties']['categorical_columns'])
        num_input_feats = np.zeros(num_cols, dtype=np.int32)
        categories = X['dataset_properties']['categories']

        for idx, cats in enumerate(categories, start=num_numerical_columns):
            num_input_feats[idx] = len(cats)
        return num_numerical_columns, num_input_feats
```

autoPyTorch/pipeline/components/setup/network_embedding/base_network_embedding.py
```diff
@@ -31,22 +31,21 @@ def transform(self, X: Dict[str, Any]) -> Dict[str, Any]:
...
-        # categories = X['dataset_properties']['categories']
-        #
-        # for i, category in enumerate(categories):
-        #     num_input_features[num_numerical_columns + i, ] = len(category)
-        # return num_numerical_columns, num_input_features
-        return None, None
+        if len(X['dataset_properties']['numerical_columns']) == 0:
+            num_numerical_columns = 0
+        else:
+            X_train = copy.deepcopy(X['backend'].load_datamanager().train_tensors[0][:2])
```
nabenabe0928, line 39: We do not have to deepcopy `X_train` because we do not change `X_train`?
Also it is not intuitive what `train_tensors[0][:2]` implies, so could you add the shape of `X_train` as a comment-out?

autoPyTorch/pipeline/components/setup/network_embedding/base_network_embedding.py
```diff
@@ -31,22 +31,21 @@ def transform(self, X: Dict[str, Any]) -> Dict[str, Any]:
     def build_embedding(self, num_input_features: np.ndarray, num_numerical_features: int) -> nn.Module:
         raise NotImplementedError
 
-    def _get_args(self, X: Dict[str, Any]) -> Tuple[None, None]:  # Tuple[int, np.ndarray]:
+    def _get_args(self, X: Dict[str, Any]) -> Tuple[int, np.ndarray]:
```
nabenabe0928, line 34: What `args` is it?

autoPyTorch/pipeline/components/setup/early_preprocessor/EarlyPreprocessing.py
```diff
@@ -32,14 +32,13 @@ def fit(self, X: Dict[str, Any], y: Any = None) -> "EarlyPreprocessing":
...
-                X_train = X['backend'].load_datamanager().train_tensors[0]
-
-            X['X_train'] = preprocess(dataset=X_train, transforms=transforms)
+        if 'X_train' in X:
+            X_train = X['X_train']
+        else:
+            # Incorporate the transform to the dataset
+            X_train = X['backend'].load_datamanager().train_tensors[0]
+
+        X['X_train'] = preprocess(dataset=X_train, transforms=transforms)
```
nabenabe0928, line 41: Do we need `deepcopy` here?

nabenabe0928 (commented):
Thanks for the PR, I reviewed all the files except tabular_feature_validator.py related (two) files.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -14,67 +14,51 @@
...
 def _create_column_transformer(
     preprocessors: Dict[str, List[BaseEstimator]],
-    numerical_columns: List[str],
     categorical_columns: List[str],
 ) -> ColumnTransformer:
     """
     Given a dictionary of preprocessors, this function
     creates a sklearn column transformer with appropriate
     columns associated with their preprocessors.
-
```
nabenabe0928, line 31: I think it is better to have a space between description, Args, and Returns (because Google style follows it).
There are many where you removed space, so could you get them back?
It also makes easier to focus on the core changes in the GitHub GUI

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -14,67 +14,51 @@
...
     Returns:
         Dict[str, List[BaseEstimator]]
     """
     preprocessors: Dict[str, List[BaseEstimator]] = dict()
 
     # Categorical Preprocessors
-    onehot_encoder = OneHotEncoder(categories='auto', sparse=False, handle_unknown='ignore')
+    ordinal_encoder = OrdinalEncoder(handle_unknown='use_encoded_value',
+                                                  unknown_value=-1)
```
nabenabe0928, line 58:
```suggestion
    ordinal_encoder = OrdinalEncoder(handle_unknown='use_encoded_value',
                                     unknown_value=-1)
```

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -14,67 +14,51 @@
...
     This function creates a Dictionary containing a list
     of numerical and categorical preprocessors
-
     Returns:
         Dict[str, List[BaseEstimator]]
     """
     preprocessors: Dict[str, List[BaseEstimator]] = dict()
 
     # Categorical Preprocessors
-    onehot_encoder = OneHotEncoder(categories='auto', sparse=False, handle_unknown='ignore')
```
nabenabe0928, line 69: Why do we remove onehot encoder?

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -14,67 +14,51 @@
...
+    ordinal_encoder = OrdinalEncoder(handle_unknown='use_encoded_value',
+                                                  unknown_value=-1)
     categorical_imputer = SimpleImputer(strategy='constant', copy=False)
 
-    # Numerical Preprocessors
-    numerical_imputer = SimpleImputer(strategy='median', copy=False)
-    standard_scaler = StandardScaler(with_mean=True, with_std=True, copy=False)
-
-    preprocessors['categorical'] = [categorical_imputer, onehot_encoder]
-    preprocessors['numerical'] = [numerical_imputer, standard_scaler]
```
nabenabe0928, line 77: Btw, I did not notice that `numerical_preprocessor` was deleted from the dev branch as well, but is there any reason?

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -161,31 +145,48 @@ def _fit(
 
             X = cast(pd.DataFrame, X)
 
-            self.all_nan_columns = set([column for column in X.columns if X[column].isna().all()])
```
nabenabe0928, line 164: Check from here

nabenabe0928 (commented):
Add some comments on tabular_feature_validator.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -161,31 +145,48 @@ def _fit(
```
nabenabe0928, line 145: Can we change the name of this method to either `fit_` or `fit_validator` because it is confusing to have `_fit`, which is an implicit private method, along with `fit` in the parent class?

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -264,6 +265,17 @@ def transform(
...
+                for column in X.columns:
+                    if column in self.all_nan_columns:
+                        if not X[column].isna().all():
+                            X[column] = np.nan
+                        X[column] = pd.to_numeric(X[column])
+            if len(self.categorical_columns) > 0:
+                categorical_columns = self.column_transformer.transformers_[0][-1]
+                for column in categorical_columns:
+                    if X[column].isna().all():
+                        X[column] = X[column].astype('object')
```
nabenabe0928, line 278: Can we just do the following here because `hasattr(X, "iloc") and not scipy.sparse.issparse(X)` is the exact condition that we will get `self.all_nan_columns is not None`?

```suggestion
            if self.all_nan_columns is None:
                raise ValueError('_fit must be called before calling transform')

            for col in list(self.all_nan_columns):
                X[col] = np.nan
                X[col] = pd.to_numeric(X[col])

            if len(self.categorical_columns) > 0:
                categorical_cols = self.column_transformer.transformers_[0][-1]
                for col in categorical_cols:
                    if X[col].isna().all():  # all nan yields numeric, so modify it here
                        X[col] = X[col].astype('object')
```

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -392,15 +398,12 @@ def _get_columns_info(
...
         # Register if a column needs encoding
-        numerical_columns = []
         categorical_columns = []
         # Also, register the feature types for the estimator
         feat_type = []
 
         # Make sure each column is a valid type
         for i, column in enumerate(X.columns):
-            if self.all_nan_columns is not None and column in self.all_nan_columns:
-                continue
```
nabenabe0928, line 403: Why do not we need this anymore?

nabenabe0928 (changes requested):
I checked the tabular_feature_validator

test/test_pipeline/components/preprocessing/test_scalers.py
ravinkohli: oh yeah. I forgot to delete them, I'll rerequest review after removing these and addressing your other comments

autoPyTorch/datasets/base_dataset.py
ravinkohli: So during last week's meeting we discussed about how to handle preprocessing in regularisation cocktails. If you remember, we had moved preprocessing to the tabular feature validator and disabled their hyperparameters. But we decided that we need to search the hyperparameters for preprocessing as well. So, in this PR I am making preprocessing back to how it is in development. Moreover, during our experiments for regularisation cocktails, we realised that a lot of time is taken due to transforming the dataset per epoch as it is currently implemented in the development branch. It's better to preprocess the whole dataset all at once per configuration. The idea behind `is_small_preprocess` was to use it as a flag that can be used to determine if the dataset can be preprocessed all at once or not. Based on this we either preprocessed the dataset in the early preprocessing node, or we move these to the transforms in the data loader. However, we never actually used this flag and always added it to the data loader transforms. With this PR, in conjunction with what we discussed in the meeting, I have removed this flag and now we preprocess the whole dataset in the early preprocessing node.

autoPyTorch/pipeline/components/setup/network_backbone/base_network_backbone.py
ravinkohli: sure

autoPyTorch/pipeline/components/setup/network_embedding/base_network_embedding.py
ravinkohli: So, in this node we are checking what the output shape would be for the training dataset based on the preprocessing applied to it. Again, this is not something new to `AutoPyTorch` as this is the same way it has been defined in development, see [here](https://github.com/automl/Auto-PyTorch/blob/06e67de5017b4cccad9398e24a3d9f0bd8176da3/autoPyTorch/pipeline/components/setup/network_embedding/base_network_embedding.py#L37-L40)

autoPyTorch/pipeline/components/setup/network_embedding/base_network_embedding.py
ravinkohli: I'll add comments

autoPyTorch/pipeline/components/setup/early_preprocessor/EarlyPreprocessing.py
ravinkohli: no we actually want to preprocess the dataset and reflect the changes also in datamanager.

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: yeah I'll fix this, thanks for pointing out.

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: To make it the same as `development` branch

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: It was never there in the development branch. It was an experimental implementation used only for regularisation cocktail experiments.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -161,31 +145,48 @@ def _fit(
...
-            )
+                encoded_categories = self.column_transformer.\
+                    named_transformers_['categorical_pipeline'].\
+                    named_steps['ordinalencoder'].categories_
+                self.categories = [
+                    # We fit an ordinal encoder, where all categorical
+                    # columns are shifted to the left
+                    list(range(len(cat)))
+                    for cat in encoded_categories
+                ]
```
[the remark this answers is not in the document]
ravinkohli: let's do these changes after `reg_cocktails` has been merged to `development`.

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: it will be none for sparse data. Again, we would like to keep it the same as development right now.

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: I think this is a common notation used also in `sklearn`. and again we would like to stick with what it is in development

autoPyTorch/pipeline/components/setup/early_preprocessor/EarlyPreprocessing.py
ravinkohli: It relates to the functionality of [`find_components`](https://github.com/automl/Auto-PyTorch/blob/06e67de5017b4cccad9398e24a3d9f0bd8176da3/autoPyTorch/pipeline/components/base_component.py#L21)

autoPyTorch/pipeline/components/setup/network_embedding/base_network_embedding.py
ravinkohli: We do perform feature preprocessing.

autoPyTorch/data/tabular_feature_validator.py
nabenabe0928: Ok, then I will make a new PR only for this change as it is way much quicker (around 100~1000x).

nabenabe0928: Hey, did you mean reg-cocktail dev?
I did not see this process in the dev branch.

autoPyTorch/data/tabular_feature_validator.py
nabenabe0928: Didn't know it.
Will it be re-appeared in the future?

autoPyTorch/pipeline/components/setup/early_preprocessor/EarlyPreprocessing.py
nabenabe0928: Thanks, it is also not a good convention, but let's stay as it is for now:(

autoPyTorch/datasets/base_dataset.py
nabenabe0928: Thanks, I will copy your comment to google doc.

autoPyTorch/data/tabular_feature_validator.py
nabenabe0928: I understood that you would like to stick to the current notation, so let's stay as it is.
But just for your information, this usage of `_fit` is not common.
More precisely, "this usage" means to define only `_fit` in a class and just inherit `fit` from a base (or parent) class. 

So I checked all the files in sklearn and the following files have `_fit` method:
```
sklearn/pipeline.py
sklearn/ensemble/_bagging.py
sklearn/neighbors/_base.py
sklearn/impute/_base.py
sklearn/cluster/_birch.py
sklearn/cluster/_bicluster.py
sklearn/linear_model/_least_angle.py
sklearn/linear_model/_stochastic_gradient.py
sklearn/feature_selection/_rfe.py
sklearn/neural_network/_rbm.py
sklearn/neural_network/_multilayer_perceptron.py
sklearn/manifold/_t_sne.py
sklearn/preprocessing/_encoders.py
sklearn/preprocessing/_data.py
sklearn/decomposition/_fastica.py
sklearn/decomposition/_pca.py
```

Except the following four classes, all of them have `fit` method in each class:
```
sklearn/cluster/_bicluster.py::SpectralCoclustering
sklearn/cluster/_bicluster.py::SpectralBiclustering
sklearn/impute/_base.py::NeighborsBase
sklearn/preprocessing/_encoders.py::_BaseEncoder
```
Two of them are base class, so we can ignore them (although it is strange to not have `def fit()` as an abstract method) and there are only two methods that only have `_fit`.
This is why it is not common to have only `_fit` in a child class.


**NOTE**
I used the following script to check all the files with the running command of `python ~/word_checker.py --path anaconda3/lib/python3.8/site-packages/sklearn --word "def _fit("`:

```python
import os
from argparse import ArgumentParser
from typing import Tuple


def _is_word_in_file(word: str, file: str) -> bool:
    with open(file, 'r') as f:
        for line in f:
            if word in line:
                return True
        else:
            return False


def _get_args() -> Tuple[str, str, str]:
    parser = ArgumentParser()
    parser.add_argument(
        '--path',
        required=True,
        type=str,
        help='The path of interest'
    )
    parser.add_argument(
        '--word',
        required=True,
        type=str,
        help='The word to check whether each file has'
    )
    parser.add_argument(
        '--ext',
        default='.py',
        type=str,
        help='The extension of files to check such as `.tex`, `.py`, `.csv`.\n'
             'You can omit the period and just parse such as `tex`, `py`, `csv` as well'
    )
    try:
        args = parser.parse_args()
    except:
        import sys
        parser.print_help()
        sys.exit()

    path, word, ext = args.path, args.word, args.ext
    if not ext.startswith('.'):
        ext = '.' + ext

    return path, word, ext


if __name__ == '__main__':
    path, word, ext = _get_args()
    path = path if 'home' in path else os.path.join(os.environ['HOME'], path)

    for root, dirs, files in os.walk(path): 
        if len(files) == 0:
            continue

        target_files = [os.path.join(root, fn) for fn in files if fn.endswith(ext)]
        passed_files = [fn for fn in target_files if _is_word_in_file(word, fn)]

        if len(passed_files) > 0:
            print('\n'.join(passed_files))
```

autoPyTorch/pipeline/components/setup/network_embedding/base_network_embedding.py
```diff
@@ -31,22 +31,37 @@ def transform(self, X: Dict[str, Any]) -> Dict[str, Any]:
...
+        else:
+            X_train = copy.deepcopy(X['backend'].load_datamanager().train_tensors[0][:2])
+
+            numerical_column_transformer = X['tabular_transformer'].preprocessor. \
+                named_transformers_['numerical_pipeline']
+            num_numerical_columns = numerical_column_transformer.transform(
+                X_train[:, X['dataset_properties']['numerical_columns']]).shape[1]
+
+        num_cols = num_numerical_columns + len(X['dataset_properties']['categorical_columns'])
+        num_input_feats = np.zeros(num_cols, dtype=np.int32)
```
nabenabe0928, line 61: In the tabular_feature_validator, we move categoricals to left, but it moves numericals to left.
Doesn't that matter?

test/test_data/test_feature_validator.py
```diff
@@ -678,16 +684,11 @@ def test_feature_validator_imbalanced_data():
...
+    assert train_feature_types == ['numerical', 'numerical', 'numerical', 'numerical']
     # validator will throw an error if the column types are not the same
     transformed_X_test = validator.transform(X_test)
     transformed_X_test = pd.DataFrame(transformed_X_test)
     assert sorted(validator.all_nan_columns) == sorted(['A', 'C', 'D'])
-    # as there are no categorical columns, we can make such an
-    # assertion. We only expect to drop the all nan columns
-    total_all_nan_columns = len(validator.all_nan_columns)
-    total_columns = len(validator.column_order)
-    assert total_columns - total_all_nan_columns == len(transformed_X_test.columns)
```
nabenabe0928, line 690: We need it, no?

autoPyTorch/datasets/resampling_strategy.py
```diff
@@ -39,13 +39,6 @@ def __call__(self, random_state: np.random.RandomState, val_share: float,
         ...
 
 
-class NoResamplingFunc(Protocol):
-    def __call__(self,
-                 random_state: np.random.RandomState,
-                 indices: np.ndarray) -> np.ndarray:
-        ...
```
nabenabe0928, line 46: What this change is supposed to mean? (I am lost)

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -264,6 +260,20 @@ def transform(
...
+                for column in X.columns:
+                    if column in self.all_nan_columns:
+                        if not X[column].isna().all():
+                            X[column] = np.nan
+                        X[column] = pd.to_numeric(X[column])
+            if len(self.categorical_columns) > 0:
+                if self.column_transformer is None:
+                    raise AttributeError("Expect column transformer to be built"
+                                         "if there are categorical columns")
+                categorical_columns = self.column_transformer.transformers_[0][-1]
```
nabenabe0928, line 273:
```suggestion
                # transformers_: https://scikit-learn.org/stable/modules/generated/sklearn.compose.ColumnTransformer.html
                categorical_columns = self.column_transformer.transformers_[0][-1]
```

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -264,6 +260,20 @@ def transform(
...
+                for column in categorical_columns:
+                    if X[column].isna().all():
+                        X[column] = X[column].astype('object')
```
nabenabe0928, line 276: This is quicker

```suggestion
                all_nan_columns = set() self.all_nan_columns is None else self.all_nan_columns
                for column in categorical_columns:
                    if column in all_nan_columns or X[column].isna().all():
                        X[column] = X[column].astype('object')
```

autoPyTorch/datasets/resampling_strategy.py
ravinkohli: these are removing some artifacts that came due to rebasing from development

autoPyTorch/pipeline/components/setup/network_embedding/base_network_embedding.py
ravinkohli: yeah it is only called once per configuration, when we are fitting the embedding layer.

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: No, I dont think we need it. We are doing those steps in the pipeline with simple imputer and scaling components. So, to explain the change, we are learning the hyperparameters for those preprocessors instead of using the default ones. This is the same as what is currently happening in `development`.

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: actually, this is the `transform` method. and `self.all_nan_columns` is built when we fit on the training dataset. So, basically the validation data or the test data all nan columns are not present in `self.all_nan_columns` and hence we need this check.

autoPyTorch/pipeline/components/setup/network_embedding/base_network_embedding.py
ravinkohli: yeah that is a good point but [here](https://github.com/automl/Auto-PyTorch/blob/e6c37a030223c9a22ec64be723577d0a2e68e4b0/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/TabularColumnTransformer.py#L61) sklearn moves them to the left. Basically, whichever columns are processed first in a column transformer, are moved to the left.

test/test_data/test_feature_validator.py
ravinkohli: Actually as I removed the numerical preprocessors, all_nan_columns are not dropped in the data validator. They are instead dropped in the simple imputer component.

autoPyTorch/pipeline/components/setup/network_embedding/base_network_embedding.py
nabenabe0928: It sounds more efficient in terms of cache efficiency.
Thanks for the info.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -366,10 +374,10 @@ def _check_data(
                 self.column_order = column_order
 
             dtypes = [dtype.name for dtype in X.dtypes]
-
+            diff_cols = X.columns[[s_dtype != dtype for s_dtype, dtype in zip(self.dtypes, dtypes)]]
             if len(self.dtypes) == 0:
                 self.dtypes = dtypes
-            elif self.dtypes != dtypes:
+            elif not self._is_datasets_consistent(diff_cols, X):
```
nabenabe0928, line 380: I have not understood how you are working on this PR, but do we need `_is_datasets_consistent` after all?
I thought that you changed files in a way that we need to check only dtype differences.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -517,11 +525,17 @@ def infer_objects(self, X: pd.DataFrame) -> pd.DataFrame:
...
+                # when train data has no object dtype, but test does
+                # we prioritise the datatype given in training data
+                for column, data_type in zip(X.columns, self.dtypes):
+                    X[column] = X[column].astype(data_type)
+            else:
+                # Calling for the first time to infer the categories
+                X = X.infer_objects()
+                for column, data_type in zip(X.columns, X.dtypes):
+                    if not is_numeric_dtype(data_type):
+                        X[column] = X[column].astype('category')
```
nabenabe0928, line 538:
```suggestion
        elif len(self.dtypes) != 0:  # when train data has no object dtype, but test does
            # we prioritise the datatype given in training data
            for column, data_type in zip(X.columns, self.dtypes):
                X[column] = X[column].astype(data_type)
        else:  # Calling for the first time to infer the categories
            X = X.infer_objects()
            for column, data_type in zip(X.columns, X.dtypes):
                if not is_numeric_dtype(data_type):
                    X[column] = X[column].astype('category')
```

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -517,11 +525,17 @@ def infer_objects(self, X: pd.DataFrame) -> pd.DataFrame:
...
-            # Calling for the first time to infer the categories
-            X = X.infer_objects()
-            for column, data_type in zip(X.columns, X.dtypes):
-                if not is_numeric_dtype(data_type):
-                    X[column] = X[column].astype('category')
+            if len(self.dtypes) != 0:
+                # when train data has no object dtype, but test does
+                # we prioritise the datatype given in training data
+                for column, data_type in zip(X.columns, self.dtypes):
+                    X[column] = X[column].astype(data_type)
```
nabenabe0928, line 532: It does not work if the number of features (let's say 10^5) is huge.
So could you change it to the following?
```suggestion
                dtype_dict = {col: dtype for col, dtype in zip(X.columns, self.dtypes)}
                X = X.astype(dtype_dict)
```

autoPyTorch/data/tabular_feature_validator.py
nabenabe0928, line 538: Here as well. It does not run when we have a huge number of features.

```suggestion
                dtype_dict = {col: 'category' for col, dtype in zip(X.columns, X.dtypes) if not is_numeric_dtype(dtype)}
                X = X.astype(dtype_dict)
```

autoPyTorch/data/tabular_feature_validator.py
nabenabe0928: Yeah, I knew it, but wouldn't we like to change all the columns if it has any possibilities that it has nan?

Btw, this part also needs to be changed for a huge number of features. (I will send another comment in the latest change)

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -265,6 +267,22 @@ def transform(
...
+                X[col] = pd.to_numeric(X[col])
+
+        if len(self.categorical_columns) > 0:
+            if self.column_transformer is None:
+                raise AttributeError("Expect column transformer to be built"
+                                     "if there are categorical columns")
+            categorical_columns = self.column_transformer.transformers_[0][-1]
+            for column in categorical_columns:
+                if X[column].isna().all():
+                    X[column] = X[column].astype('object')
```
nabenabe0928, line 284: Change for a huge number of features.
```suggestion
            cat_cols = self.column_transformer.transformers_[0][-1]  # categorical columns
            all_nan_cat_cols = set(X[X[cat_cols].isna().all()].columns())
            dtype_dict = {col: 'object' for col in cat_cols if col in all_nan_cat_cols}
            X = X.astype(dtype_dict)
```

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: hey, sorry for this. Actually, the issue I'm facing is that this PR is not running any checks. So the changes I made were causing some tests to fail. I ran the tests locally and found that we still need that check.

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: I think these are just preferences on where to start the comment.

nabenabe0928: Nono, it actually removed an indent level.

nabenabe0928:
Hey, thanks for your effort.
I think we can merge once you addressed [this](https://github.com/automl/Auto-PyTorch/pull/369#discussion_r797465995).

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: actually, if you notice we are also saving the dtypes in  `self.object_dtype_mapping` which should be done for both of the two conditions you moved back an indent level. So, I think its fine the way it is.

nabenabe0928: Oh yeah, I did not notice, but I also did not notice that we still have the same issue (which happens when we have a huge number of features) in this method.
Could you fix it?

```python
        if hasattr(self, 'object_dtype_mapping'):
            # Mypy does not process the has attr. This dict is defined below
            try:
                X = X.astype(self.object_dtype_mapping)
            except Exception as e:
                self.logger.warning(f'Casting the column {key} to {dtype} caused the exception {e}')
                pass
            return

        if len(self.dtypes) != 0:
            # when train data has no object dtype, but test does.  Prioritise the datatype given in training data
            dtype_dict = {col: dtype for col, dtype in zip(X.columns, self.dtypes)}
            X = X.astype(self.object_dtype_mapping)
        else:
            # Calling for the first time to infer the categories
            X = X.infer_objects()
            dtype_dict = {col: 'category' for col, dtype in X.items() if not is_numeric_dtype(dtype)}
            X = X.astype(dtype_dict)

        # only numerical attributes and categories
        self.object_dtype_mapping = {col: dtype for col, dtype in zip(X.columns, X.dtypes)}
```

nabenabe0928:
Hi, thanks for the changes and sorry for the late response.
I will approve the changes.

nabenabe0928 (approved):

autoPyTorch/datasets/resampling_strategy.py
ArlindKadra: It was added twice.

autoPyTorch/api/tabular_classification.py
```diff
@@ -14,7 +14,6 @@
 from autoPyTorch.datasets.base_dataset import BaseDatasetPropertiesType
 from autoPyTorch.datasets.resampling_strategy import (
     HoldoutValTypes,
-    CrossValTypes,
```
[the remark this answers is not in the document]
ravinkohli: no we did not remove cross validation. Actually this import was happening due to mypy but now I made a new type variable called `ResamplingStrategies` so we dont need it anymore

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -32,49 +31,36 @@ def _create_column_transformer(
...
 
     Returns:
         ColumnTransformer
     """
 
-    numerical_pipeline = 'drop'
-    categorical_pipeline = 'drop'
-    if len(numerical_columns) > 0:
-        numerical_pipeline = make_pipeline(*preprocessors['numerical'])
-    if len(categorical_columns) > 0:
```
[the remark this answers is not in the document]
ravinkohli: it is being checked before this function is called. see [here](https://github.com/automl/Auto-PyTorch/blob/4698386a91bd3ede75eec9258c4d2f73599cede6/autoPyTorch/data/tabular_feature_validator.py#L164). In case we dont have categorical columns no column transformer will be created.

autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -265,6 +267,22 @@ def transform(
...
+
+            for col in list(self.all_nan_columns):
+                X[col] = np.nan
+                X[col] = pd.to_numeric(X[col])
+
+        if len(self.categorical_columns) > 0:
+            if self.column_transformer is None:
+                raise AttributeError("Expect column transformer to be built"
+                                     "if there are categorical columns")
+            cat_cols = self.column_transformer.transformers_[0][-1]  # categorical columns
```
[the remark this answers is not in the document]
ravinkohli: true, we can make it more readable. I'll change this

autoPyTorch/data/tabular_feature_validator.py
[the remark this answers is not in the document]
ArlindKadra: True that, I thought we were going to check like before.

nabenabe0928:
Looks good to me now:)

autoPyTorch/api/tabular_classification.py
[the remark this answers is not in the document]
ArlindKadra: Hmm, but why was the later error message removed which notified the user if the resampling strategy was not supported?

ArlindKadra:
Thanks for the PR :). It looks good to me too now. I only have one last minor question.

PR #370, opened 2022-01-27
[fix] Hotfix debug no training in simple intensifier

nabenabe0928:
Fix the issue when using the simple intensifier in SMAC.

Currently, we can only train an initial configuration and this PR fixes this issue.


ravinkohli (approved):
Thanks for this fix. I think we can merge it when the checks pass.

PR #375, opened 2022-02-03
[FIX] Remove redundant categorical imputation

### LINKED ISSUE #374: Redundant imputation for categorical columns
NOTE: ISSUES ARE NOT FOR CODE HELP - Ask for Help at https://stackoverflow.com

Your issue may already be reported!
Also, please search on the [issue tracker](../) before creating one.

* **I'm submitting a ...**
  - [x] bug report

# Issue Description
We are already imputing categorical columns in the feature validator [here](https://github.com/automl/Auto-PyTorch/blob/bd4fabf4fbc8a1c4876f6901c8a3f06df213bb15/autoPyTorch/data/tabular_feature_validator.py#L62). However, we still use the SimpleImputer in the pipeline to impute categorical columns which are redundant because we would not have any nans in the categorical columns by this time, see [here](https://github.com/automl/Auto-PyTorch/blob/bd4fabf4fbc8a1c4876f6901c8a3f06df213bb15/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py#L70). This will also save us a hyperparameter during the search.

## Expected Behavior

The redundancy has no effect on the output of the simple imputer.

## Current Behavior

The redundancy has no effect on the output of the simple imputer.

## Possible Solution

Remove categorical imputer from `SimpleImputer`.

ravinkohli:
This PR removes categorical imputation from `SimpleImputer` to resolve #374 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

## Description

This PR removes the hyperparameter `categorical_strategy` from `SimpleImputer` as we are already imputing categorical columns in the feature validator.

## Motivation and Context

Resolves #374 

## How has this been tested?

I have adjusted the tests for SimpleImputer according to its new behavior.


autoPyTorch/optimizer/smbo.py
```diff
@@ -244,10 +244,12 @@ def __init__(self,
                                               port=self.logger_port)
         self.logger.info("initialised {}".format(self.__class__.__name__))
 
-        self.initial_configurations: Optional[List[Configuration]] = None
         if portfolio_selection is not None:
-            self.initial_configurations = read_return_initial_configurations(config_space=config_space,
-                                                                             portfolio_selection=portfolio_selection)
+            initial_configurations = read_return_initial_configurations(config_space=config_space,
```
eddiebergman, line 248: This way makes more sense but it means there is no longer and attribute `self.initial_configurations` if `portfolio_selection` is `None`. While unlikely to matter, this means code like `if self.initial_configurations:` happening elsewhere will now break as it doesn't exist. I'd keep the default thing that it is `None` and given a value `if portfolio_selection is not None`.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/TabularColumnTransformer.py
```diff
@@ -48,13 +48,16 @@ def fit(self, X: Dict[str, Any], y: Any = None) -> "TabularColumnTransformer":
...
+        # let the data in categorical columns pass through
+        numerical_pipeline = 'passthrough'
+        categorical_pipeline = 'passthrough'
 
         preprocessors = get_tabular_preprocessers(X)
-        if len(X['dataset_properties']['numerical_columns']):
+        if len(X['dataset_properties']['numerical_columns']) and len(preprocessors['numerical']):
             numerical_pipeline = make_pipeline(*preprocessors['numerical'])
-        if len(X['dataset_properties']['categorical_columns']):
+        if len(X['dataset_properties']['categorical_columns']) and len(preprocessors['categorical']):
```
eddiebergman, line 60: These lines make no sense to me, `if len(...) and <cond>`, I"m assuming it's checking for `len(...) > 0` but I would prefer to be explicit about that.

eddiebergman (commented):
While in principal, it makes sense and definitely means optimization is easier as there are less parameters, there's two points that cross my mind.

* In theory it would be much nicer to just have it as part of the pipeline but have no choice, it will always be performed. This would make preprocessing a configurable pipeline step where it would make more sense to me. For example, we have this problem now with Autosklearn where preprocessing done outside the pipeline is much harder to export.
* Have you given thought to any users who might want to configure this? Injecting it as a pipeline component is easier than exposing parameters that then get passed through.

As mentioned, we have the same problems too but I think rethinking the pipelines will eventually be on the agenda. Anywho, approved except for some minor comments.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/SimpleImputer.py
```diff
@@ -13,70 +13,42 @@
...
 
         Args:
-            X (np.ndarray):
-                The input features to fit on
-            y (Optional[np.ndarray]):
-                The labels for the input features `X`
+            X (Dict[str, Any]):
+                The fit dictionary
+            y (Optional[Any]):
+                Not Used -- to comply with API
```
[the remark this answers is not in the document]
ravinkohli: yeah, there is outdated documentation throughout the pipeline components. I'll raise an issue and fix them with a different PR.

autoPyTorch/optimizer/smbo.py
ravinkohli: yeah, thanks for pointing it out. I'll fix this.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/TabularColumnTransformer.py
ravinkohli: yeah, I agree. actually, we don't need to check for the length of columns as we will only have preprocessors if the length of columns is > 0. So, I'll change these conditions.

ravinkohli:
> While in principal, it makes sense and definitely means optimization is easier as there are less parameters, there's two points that cross my mind.
> 
>     * In theory it would be much nicer to just have it as part of the pipeline but have it as the only default choice, it will always be performed. This would make preprocessing a configurable pipeline step where it would make more sense to me. For example, we have this problem now with Autosklearn where preprocessing done outside the pipeline is much harder to export.
>
I agree if we can export everything that is done on the dataset (preprocessing, training, etc) that would be great and also useful for people to use our fitted pipelines directly. 

>     * Have you given thought to any users who might want to configure this? Injecting it as a pipeline component is easier than exposing parameters that then get passed through.
> 
Actually no, I have not yet considered this use case.

> As mentioned, we have the same problems too but I think rethinking the pipelines will eventually be on the agenda. Anywho, approved except for some minor comments.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/base_imputer.py
```diff
@@ -26,7 +25,7 @@ def transform(self, X: Dict[str, Any]) -> Dict[str, Any]:
         Returns:
             (Dict[str, Any]): the updated 'X' dictionary
         """
-        if self.preprocessor['numerical'] is None and self.preprocessor['categorical'] is None:
+        if self.preprocessor['numerical'] is None and len(X["dataset_properties"]["numerical_columns"]) != 0:
             raise ValueError("cant call transform on {} without fitting first."
```
nabenabe0928, line 29:
```suggestion
            raise ValueError("cannot call transform on {} without fitting first."
```

nabenabe0928 (commented):
Thanks for the PR.
It looks mostly good to me.
I still need to check the test file, but I did not see any big issues in other files.

test/test_pipeline/components/preprocessing/test_imputers.py
```diff
@@ -39,14 +39,14 @@ def test_get_config_space(self):
...
+                         [4.0, 5, np.nan],
                          [np.nan, 2, 3],
-                         ['7.0', np.nan, 9],
-                         ['4.0', np.nan, np.nan]], dtype=object)
-        numerical_columns = [1, 2]
-        categorical_columns = [0]
+                         [7.0, np.nan, 9],
+                         [4.0, np.nan, np.nan]])
+        numerical_columns = [0, 1, 2]
+        categorical_columns = []
```
nabenabe0928, line 49:
```suggestion
        categorical_columns = []  # This imputation is not for categoricals!    
```

test/test_pipeline/components/preprocessing/test_imputers.py
```diff
@@ -66,31 +66,29 @@ def test_mean_imputation(self):
...
 
     def test_median_imputation(self):
-        data = np.array([['1.0', np.nan, 3],
+        data = np.array([[1.0, np.nan, 3],
                          [np.nan, 8, 9],
-                         ['4.0', 5, np.nan],
+                         [4.0, 5, np.nan],
                          [np.nan, 2, 3],
-                         ['7.0', np.nan, 9],
-                         ['4.0', np.nan, np.nan]], dtype=object)
```
nabenabe0928, line 91: Because we would like to test the median is correct, shouldn't we test by the data whose median is different from mean?

test/test_pipeline/components/preprocessing/test_imputers.py
```diff
@@ -110,31 +108,29 @@ def test_median_imputation(self):
...
 
     def test_frequent_imputation(self):
-        data = np.array([['1.0', np.nan, 3],
+        data = np.array([[1.0, np.nan, 3],
                          [np.nan, 8, 9],
-                         ['4.0', 5, np.nan],
+                         [4.0, 5, np.nan],
                          [np.nan, 2, 3],
-                         ['7.0', np.nan, 9],
-                         ['4.0', np.nan, np.nan]], dtype=object)
```
nabenabe0928, line 135: here as well. We should modify the test dataset so that the imputation value is more obvious.
This example requires tie-breaking for the most frequent value.

test/test_pipeline/components/preprocessing/test_imputers.py
```diff
@@ -190,8 +183,7 @@ def test_constant_imputation(self):
             'X_train': data[train_indices],
             'dataset_properties': dataset_properties
         }
-        imputer_component = SimpleImputer(numerical_strategy='constant_zero',
-                                          categorical_strategy='constant_!missing!')
```
nabenabe0928, line 194: Where do we test `constant_missing` now?

nabenabe0928 (changes requested):
Hi I added some requests regarding the enhancements for the test!

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/imputation/base_imputer.py
eddiebergman: "can not" is two words :)

nabenabe0928: I did not know that "can not" exists, but "cannot" is preferred for formal settings?

- [article1](https://www.grammarly.com/blog/cannot-or-can-not/) --> "You’ll find cannot in formal writing and speech. Research studies, academic reports, and professional presentations are places where cannot is the most appropriate choice."
- [article 2](https://grammar.yourdictionary.com/vs/cannot-vs-can-not-vs-cant-differences-explained.html) --> "While can not is an acceptable alternate spelling, cannot is generally preferred by most writers."
- [article 3](https://www.merriam-webster.com/words-at-play/cannot-vs-can-not-is-there-a-difference) --> "Both cannot and can not are perfectly fine, but cannot is far more common and is therefore recommended"

test/test_pipeline/components/preprocessing/test_imputers.py
ravinkohli: we have removed it now, there is no categorical strategy for the simple imputer.

test/test_pipeline/components/preprocessing/test_imputers.py
ravinkohli: Actually none of the imputations are for categorical columns, that is the point of this PR

test/test_pipeline/components/preprocessing/test_imputers.py
nabenabe0928: Yeah, I know the background of this PR, but I mean we still have a categorical imputer somewhere, right?
Are we testing this imputer? This was my question.

ravinkohli: oh no sorry, that is being tested in feature validation.

PR #387, opened 2022-02-22
[ADD] dataset compression

ravinkohli:
reduces the precision of the dataset if it takes more memory than passed with `memory_allocation`.

## Types of changes

## Checklist:

## Description

This PR adds a new function `reduce_dataset_if_too_large`, which takes an argument `memory_allocation` and if the passed dataset takes more memory than the `memory_allocation`, we reduce the precision of the datasets by modifying the underlying data types. It only works for numerical columns. For numpy arrays, the dataset is mapped using `reduction_mappings` which is a `Mapping` object. `reduction_mappings` maps the dtypes to its less precise dtype for example, `np.float64` -> `np.float32`. For pandas data frames, we use the `downcast` parameter of the `to_numeric` function which reduces the precision of the data to the lowest possible type. For testing dataset, we use the final precision of the training dataset.

This reduction is performed in the `transform` method of `tabular_feature_validator` after all other steps are completed.

## Motivation and Context

This change will enable us to run `AutoPyTorch` with larger datasets as currently we face memory out errors.

## How has this been tested?

I have tested the functionality of `reduce_dataset_if_too_large` by ensuring that the memory footprint of the dataset decreases. I have also tested the different errors raised in the `validate_dataset_compression_args` as well as its behaviour when no args are passed. Additionally, I have tested fit and transform on numpy data with and without nans, pandas data with and without nans and data with various sparse formats.


autoPyTorch/data/tabular_feature_validator.py
```diff
@@ -272,6 +284,19 @@ def transform(
...
+            )
+            and self._dataset_compression is not None
+        ):
+            if self._precision is not None:
+                X = X.astype(self._precision)
+            else:
+                X = reduce_dataset_size_if_too_large(X, **self._dataset_compression)
+                self._precision = dict(X.dtypes) if hasattr(X, 'iloc') else X.dtype
+
         return X
```
nabenabe0928, line 300: We should separate it as a method (btw, Does this method really return `np.ndarray`?)
```suggestion
    def _compress_dataset(self, X: np.ndarray) -> np.ndarray:
        is_dataframe = hasattr(X, 'iloc')
        is_reducible_type = isinstance(X, np.ndarray) or scipy.sparse.issparse(X) or is_dataframe
        if not is_reducible_type or self._dataset_compression is None:
            return X
        elif self._precision is not None:
            X = X.astype(self._precision)
            return X
        else:
            X = reduce_dataset_size_if_too_large(X, **self._dataset_compression)
            self._precision = dict(X.dtypes) if is_dataframe else X.dtype
            return X
```

autoPyTorch/api/tabular_classification.py
```diff
@@ -368,11 +375,54 @@ def search(
...
+            if dataset_compression is True:
+                self._dataset_compression = default_dataset_compression_arg
+            else:
+                self._dataset_compression = None
+        else:
+            self._dataset_compression = dataset_compression
+
+        if self._dataset_compression is not None:
+            self._dataset_compression = validate_dataset_compression_arg(
+                self._dataset_compression, memory_limit=memory_limit)
```
nabenabe0928, line 425: This is the same processing as in tabular_regression, so we can make this process as a common function.

nabenabe0928 (changes requested):
Hi thanks for the PR:)
I left some comments and I will check utils.py from now.

autoPyTorch/data/utils.py
```diff
@@ -0,0 +1,309 @@
...
+from scipy.sparse import issparse, spmatrix
+
+
+# TODO: TypedDict with python 3.8
+#
+#   When upgrading to python 3.8 as minimum version, this should be a TypedDict
+#   so that mypy can identify the fields types
+DatasetCompressionSpec = Dict[str, Union[int, float, List[str]]]
+DatasetDTypeContainerType = Union[Type, Dict[str, Type]]
+DatasetCompressionInputType = Union[np.ndarray, spmatrix, pd.DataFrame]
```
nabenabe0928, line 32: Oh, I did not know that we have something like `spmatrix`...
Probably I should publish a PR to replace typing by this.

autoPyTorch/data/utils.py
```diff
@@ -0,0 +1,309 @@
...
+    "memory_allocation": 0.1,
+    "methods": ["precision"]
+}
+
+
+def validate_dataset_compression_arg(
+    dataset_compression: Mapping[str, Any],
+    memory_limit: int
+) -> DatasetCompressionSpec:
+    """Validates and return a correct dataset_compression argument
```
nabenabe0928, line 45:
```suggestion
    """Validate and return a correct dataset_compression argument
```

autoPyTorch/data/utils.py
```diff
@@ -0,0 +1,309 @@
...
+                f"\nmethods = {dataset_compression['methods']}"
+                f"\ndataset_compression = {dataset_compression}"
+            )
+
+        return cast(DatasetCompressionSpec, dataset_compression)
+    else:
+        raise ValueError(
+            f"Unknown type for `dataset_compression` {type(dataset_compression)}"
+            f"\ndataset_compression = {dataset_compression}"
+        )
```
nabenabe0928, line 127:
```suggestion
    if not isinstance(dataset_compression, Mapping):
        raise ValueError(
            f"Unknown type for `dataset_compression` {type(dataset_compression)}"
            f"\ndataset_compression = {dataset_compression}"
        )

    # Fill with defaults if they don't exist
    dataset_compression = {
        **default_dataset_compression_arg,
        **dataset_compression
    }

    # Must contain known keys
    if set(dataset_compression.keys()) != set(default_dataset_compression_arg.keys()):
        raise ValueError(
            f"Unknown key in dataset_compression, {list(dataset_compression.keys())}."
            f"\nPossible keys are {list(default_dataset_compression_arg.keys())}"
        )

    memory_allocation = validate_memory_allocation(dataset_compression["memory_allocation"], memory_limit)

    # "methods" must be non-empty sequence
    if (
        not isinstance(dataset_compression["methods"], Sequence)
        or len(dataset_compression["methods"]) <= 0
    ):
        raise ValueError(
            "key 'methods' must be a non-empty list"
            f"\nmethods = {dataset_compression['methods']}"
            f"\ndataset_compression = {dataset_compression}"
        )

    # "methods" must contain known methods
    if any(
        method not in cast(Sequence, default_dataset_compression_arg["methods"])  # mypy
        for method in dataset_compression["methods"]
    ):
        raise ValueError(
            f"key 'methods' can only contain {default_dataset_compression_arg['methods']}"
            f"\nmethods = {dataset_compression['methods']}"
            f"\ndataset_compression = {dataset_compression}"
        )

    return cast(DatasetCompressionSpec, dataset_compression)


def validate_memory_allocation(memory_allocation: Union[int, float], memory_limit: int) -> float:
    if not (isinstance(memory_allocation, float) or isinstance(memory_allocation, int)):
        raise ValueError(
            "'memory_allocation' must be an `int` or `float`"
            f"\nmemory_allocation = {memory_allocation}"
        )

    # "memory_allocation" if absolute, should be > 0 and < memory_limit
    if isinstance(memory_allocation, int) and not (0 < memory_allocation < memory_limit):
        raise ValueError(
            f"'memory_allocation' if int must be in (0, memory_limit={memory_limit})"
            f"\nmemory_allocation = {memory_allocation}"
        )

    if isinstance(memory_allocation, float):
        if not (0.0 < memory_allocation < 1.0):
            raise ValueError(
                "'memory_allocation' if float must be in (0, 1)"
                f"\nmemory_allocation = {memory_allocation}"
            )
        # convert to int so we can directly use
        memory_allocation = floor(memory_allocation * memory_limit)
    
    return memory_allocation
```

nabenabe0928 (changes requested):
I checked utils.py up to L127.

autoPyTorch/data/utils.py
```diff
@@ -0,0 +1,309 @@
...
+            if is_numeric_dtype(dtype):
+                if is_float_dtype(dtype):
+                    float_columns.append(col)
+                else:
+                    integer_columns.append(col)
+
+        if len(integer_columns) > 0:
+            X[integer_columns] = X[integer_columns].apply(lambda column: pd.to_numeric(column, downcast='integer'))
+        if len(float_columns) > 0:
+            X[float_columns] = X[float_columns].apply(lambda column: pd.to_numeric(column, downcast='float'))
```
nabenabe0928, line 232:
```suggestion
        integer_columns = X.columns[X.dtypes == int]
        float_columns = X.columns[X.dtypes == float]
        X[integer_columns] = X[integer_columns].apply(lambda column: pd.to_numeric(column, downcast='integer'))
        X[float_columns] = X[float_columns].apply(lambda column: pd.to_numeric(column, downcast='float'))
```

nabenabe0928 (changes requested):
I checked utils.py up to L240.

autoPyTorch/data/utils.py
```diff
@@ -0,0 +1,309 @@
...
+        precision = dict(X.dtypes)
+    else:
+        raise ValueError(f"Unrecognised data type of X, expected data type to "
+                         f"be in (np.ndarray, spmatrix, pd.DataFrame), but got :{type(X)}")
+
+    return X, precision, dtypes
```
nabenabe0928, line 238: The variable names are confusing?
It seems both precision and dtypes are precision and, at the same time, dtypes.
THe difference seems to be before or after, so we should rename them more informative...

autoPyTorch/data/utils.py
```diff
@@ -0,0 +1,309 @@
...
+
+def megabytes(arr: DatasetCompressionInputType) -> float:
+    if isinstance(arr, np.ndarray):
+        memory_in_bytes = arr.nbytes
+    elif issparse(arr):
+        memory_in_bytes = arr.data.nbytes
+    elif hasattr(arr, 'iloc'):
+        memory_in_bytes = arr.memory_usage(index=True, deep=True).sum()
+    else:
+        return 0
```
nabenabe0928, line 249: Isn't it better to raise an error because 0 is obviously wrong?

nabenabe0928 (commented):
I checked up to L252 of utils.py

autoPyTorch/data/utils.py
ravinkohli: no, actually pandas has a dtype `Float64DType` which is present when there are `nans`. I couldnt find another way to check for this dtype.

autoPyTorch/data/utils.py
ravinkohli: sure

autoPyTorch/data/utils.py
ravinkohli: true

autoPyTorch/api/tabular_regression.py
```diff
@@ -369,19 +376,64 @@ def search(
...
+            dataset_compression: Union[bool, Mapping[str, Any]] = True
+                We compress datasets so that they fit into some predefined amount of memory.
+                **NOTE**
+
+                Default configuration when left as ``True``:
+                .. code-block:: python
+                    {
+                        "memory_allocation": 0.1,
+                        "methods": ["precision"]
+                    }
```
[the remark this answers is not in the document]
ravinkohli: you can pass it as a dictionary

autoPyTorch/data/utils.py
```diff
@@ -0,0 +1,309 @@
...
+
+    if hasattr(np, 'float128'):
+        _mapping[np.float128] = np.float64
+
+    @classmethod
+    def __getitem__(cls, item: type) -> type:
+        for k, v in cls._mapping.items():
+            if k == item:
+                return v
+        raise KeyError(item)
```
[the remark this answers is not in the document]
ravinkohli: Actually, this comparison is the point of using this `Mapping` object. I will undo this change as it causes tests to fail

autoPyTorch/data/tabular_feature_validator.py
ravinkohli: > We should separate it as a method (btw, Does this method really return `np.ndarray`?)

This points to a very important fact which I don't know how I missed. We would and should never get a pandas dataframe here. It will always be a numpy array or a sparse matrix.

ravinkohli: I have removed the code to compress a pandas dataframe. For our record, it will be there in the commit history.

autoPyTorch/data/utils.py
ravinkohli: sure, but its not a priority.

autoPyTorch/data/utils.py
ravinkohli: I dont think this is necessary.

autoPyTorch/api/tabular_regression.py
```diff
@@ -402,20 +444,57 @@ def search(
             portfolio_selection=portfolio_selection,
         )
 
+    def _get_dataset_compression_mapping(
```
nabenabe0928, line 447: Is it a good idea to move it to BaseTask?
I mean BaseTask will be messier, but it can be one idea if it duplicates in the future.
Another idea will be to create a function that returns both compressed_dataset and reduced_dtype so that we do not require the access to self.

nabenabe0928 (commented):
Hi thanks for the changes, I will take a look at the tests next.

autoPyTorch/api/tabular_classification.py
```diff
@@ -453,6 +442,43 @@ def search(
             portfolio_selection=portfolio_selection,
         )
 
+    def _get_dataset_compression_mapping(
```
nabenabe0928, line 445: Actually we do not use self, so we can define it as a function and share it with other classes.

autoPyTorch/data/utils.py
nabenabe0928: Then could you at least perform beforehand so that we can reduce the nest?:
```python
    if not isinstance(dataset_compression, Mapping):
        raise ValueError(
            f"Unknown type for `dataset_compression` {type(dataset_compression)}"
            f"\ndataset_compression = {dataset_compression}"
        )
```

autoPyTorch/data/utils.py
nabenabe0928: Let me check then...

autoPyTorch/data/utils.py
ravinkohli: yeah sure, that makes sense.

autoPyTorch/api/tabular_classification.py
ravinkohli: True, that's better I think

autoPyTorch/data/utils.py
```diff
@@ -0,0 +1,344 @@
...
+
+        integer_columns = []
+        float_columns = []
+
+        for col, dtype in dtypes.items():
+            if is_numeric_dtype(dtype):
+                if is_float_dtype(dtype):
+                    float_columns.append(col)
+                else:
+                    integer_columns.append(col)
```
nabenabe0928, line 260: We would like to cover this line by a test.

autoPyTorch/data/utils.py
```diff
@@ -194,21 +194,21 @@ def reduce_precision(
...
     Returns:
         Tuple[DatasetCompressionInputType, DatasetDTypeContainerType, DatasetDTypeContainerType]
             Returns the reduced data X along with the dtypes it and the dtypes it was reduced to.
     """
-    precision: Optional[DatasetDTypeContainerType] = None
+    reduced_dtypes: Optional[DatasetDTypeContainerType] = None
     if isinstance(X, np.ndarray) or issparse(X):
         dtypes = X.dtype
         if X.dtype not in supported_precision_reductions:
             raise ValueError(f"X.dtype = {X.dtype} not equal to any supported"
```
nabenabe0928, line 245: We would like to test here using a data frame including categorical features

nabenabe0928 (changes requested):

test/test_data/test_feature_validator.py
```diff
@@ -557,3 +558,47 @@ def test_comparator():
...
+        input_data_featuretest, test_size=0.1, random_state=1)
+    validator = TabularFeatureValidator(dataset_compression={'memory_allocation': 0, 'methods': ['precision']})
+    validator.fit(X_train=X_train)
+    transformed_X_train = validator.transform(X_train.copy())
+
+    assert validator._reduced_dtype is not None
+    assert megabytes(transformed_X_train) < megabytes(X_train)
+
+    transformed_X_test = validator.transform(X_test.copy())
+    assert megabytes(transformed_X_test) < megabytes(X_test)
```
nabenabe0928, line 598: Let's make the test tighter.
```suggestion
    assert megabytes(transformed_X_train) < megabytes(X_train) / 2 + 1

    transformed_X_test = validator.transform(X_test.copy())
    assert megabytes(transformed_X_test) < megabytes(X_test) / 2 + 1
```

test/test_data/test_utils.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+    Tests the functionalities of `get_dataset_compression_mapping`
+    """
+    dataset_compression_mapping = get_dataset_compression_mapping(
+        dataset_compression=True,
+        memory_limit=memory_limit)
+    _verify_dataset_compression_mapping(dataset_compression_mapping)
+
+    dataset_compression_mapping = get_dataset_compression_mapping(
+        dataset_compression={'memory_allocation': 0.01, 'methods': ['precision']},
```
nabenabe0928, line 98: Can we test whether the mapping really follows this rule?

test/test_data/test_feature_validator.py
ravinkohli: actually its not necessary that we pass this check because we are only reducing the precision of integers

test/test_data/test_utils.py
ravinkohli: sure

autoPyTorch/data/utils.py
ravinkohli: this line is running for numpy arrays, but I can add a test that checks this error.

autoPyTorch/data/utils.py
```diff
@@ -0,0 +1,344 @@
...
+        return iter(cls._mapping.keys())
+
+    @classmethod
+    def __len__(cls) -> int:
+        return len(cls._mapping)
+
+
+reduction_mapping = _DtypeReductionMapping()
+supported_precision_reductions = list(reduction_mapping)
+
```
nabenabe0928, line 223:
```suggestion
FLOAT_DTYPES = tuple(dt for dt in reduction_mapping._mapping.keys() if dt.__name__.startswith("float"))
INT_DTYPES = tuple(dt for dt in reduction_mapping._mapping.keys() if dt.__name__.startswith("int"))
```

autoPyTorch/data/utils.py
nabenabe0928:
```python
        col_names = X.dtypes.index
        float_cols = col_names[[dt in FLOAT_DTYPES for dt in X.dtypes.values]]
        int_cols = col_names[[dt in INT_DTYPES for dt in X.dtypes.values]]
        X[int_cols] = X[int_cols].apply(lambda column: pd.to_numeric(column, downcast='integer'))
        X[float_cols] = X[float_cols].apply(lambda column: pd.to_numeric(column, downcast='float'))
```

autoPyTorch/data/utils.py
ravinkohli: we are not going to be using these anywhere else. Lets keep it the way you suggested on mattermost.

autoPyTorch/data/utils.py
```diff
@@ -246,23 +245,17 @@ def reduce_precision(
...
-        float_columns = []
+        col_names = X.dtypes.index
 
-        for col, dtype in dtypes.items():
-            if is_numeric_dtype(dtype):
-                if is_float_dtype(dtype):
-                    float_columns.append(col)
-                else:
-                    integer_columns.append(col)
+        float_cols = col_names[[dt.name.startswith("float") for dt in X.dtypes.values]]
```
nabenabe0928, line 254: I sent those comments before your commit
[comment 1](https://github.com/automl/Auto-PyTorch/pull/387#discussion_r814767978)
[commen 2](https://github.com/automl/Auto-PyTorch/pull/387#discussion_r814771826)

nabenabe0928 (changes requested):

nabenabe0928 (approved):

PR #395, opened 2022-02-28
[FIX] SWA and SE with non cyclic schedulers

ravinkohli:
This PR fixes a bug in swa and se when they are used with noncyclic schedulers. Also enables learned entity embeddings for reg_cocktails.

### DIFF
--- a/autoPyTorch/pipeline/components/setup/network_embedding/__init__.py
+++ b/autoPyTorch/pipeline/components/setup/network_embedding/__init__.py
@@ -155,8 +155,10 @@ class NetworkEmbeddingChoice(autoPyTorchChoice):
                     default = default_
                     break
 
-        categorical_columns = dataset_properties['categorical_columns'] \
-            if isinstance(dataset_properties['categorical_columns'], List) else []
+        if isinstance(dataset_properties['categorical_columns'], list):
+            categorical_columns = dataset_properties['categorical_columns']
+        else:
+            categorical_columns = []
 
         updates = self._get_search_space_updates()
         if '__choice__' in updates.keys():
--- a/autoPyTorch/pipeline/components/training/trainer/base_trainer.py
+++ b/autoPyTorch/pipeline/components/training/trainer/base_trainer.py
@@ -338,8 +338,8 @@ class BaseTrainerComponent(autoPyTorchTrainingComponent):
         """
         perform swa model update
         """
-        assert self.swa_model is not None, "SWA model can't be none when" \
-                                           " stochastic weight averaging is enabled"
+        if self.swa_model is None:
+            raise ValueError("SWA model cannot be none when stochastic weight averaging is enabled")
         self.swa_model.update_parameters(self.model)
         self.swa_updated = True
 
@@ -350,8 +350,8 @@ class BaseTrainerComponent(autoPyTorchTrainingComponent):
             epoch (int):
                 current epoch
         """
-        assert self.model_snapshots is not None, "model snapshots container can't be " \
-                                                 "none when snapshot ensembling is enabled"
+        if self.model_snapshots is None:
+            raise ValueError("model snapshots cannot be None when snapshot ensembling is enabled")
         is_last_epoch = (epoch == self.budget_tracker.max_epochs)
         if is_last_epoch and self.use_stochastic_weight_averaging:
             model_copy = deepcopy(self.swa_model)


nabenabe0928 (changes requested):
Hi Thanks for the PR.
I checked except base_trainer.

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -334,6 +334,35 @@ def on_epoch_start(self, X: Dict[str, Any], epoch: int) -> None:
         """
         pass
 
+    def _swa_update(self) -> None:
+        """
+        perform swa model update
+        """
+        assert self.swa_model is not None, "SWA model can't be none when" \
+                                           " stochastic weight averaging is enabled"
```
nabenabe0928, line 342:
```suggestion
        if self.swa_model is None:
            raise ValueError("SWA model cannot be none when stochastic weight averaging is enabled")
```

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -334,6 +334,35 @@ def on_epoch_start(self, X: Dict[str, Any], epoch: int) -> None:
...
+
+    def _se_update(self, epoch: int) -> None:
+        """
+        Add latest model or swa_model to model snapshot ensemble
+        Args:
+            epoch (int):
+                current epoch
+        """
+        assert self.model_snapshots is not None, "model snapshots container can't be " \
+                                                 "none when snapshot ensembling is enabled"
```
nabenabe0928, line 354:
```suggestion
        if self.model_snapshots is None:
            raise ValueError("model snapshots cannot be None when snapshot ensembling is enabled")
```

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -344,39 +373,19 @@ def on_epoch_end(self, X: Dict[str, Any], epoch: int) -> bool:
...
-                if self.use_stochastic_weight_averaging:
-                    assert self.swa_model is not None, "SWA model can't be none when" \
-                                                       " stochastic weight averaging is enabled"
-                    self.swa_model.update_parameters(self.model)
-                    self.swa_updated = True
-                if self.use_snapshot_ensemble:
-                    assert self.model_snapshots is not None, "model snapshots container can't be " \
-                                                             "none when snapshot ensembling is enabled"
-                    model_copy = deepcopy(self.swa_model) if self.use_stochastic_weight_averaging \
-                        else deepcopy(self.model)
```
nabenabe0928, line 375: Is it ok to add the condition `is_last_epoch` in the new implementation?

autoPyTorch/pipeline/tabular_classification.py
```diff
@@ -289,6 +289,29 @@ def _get_hyperparameter_search_space(self,
...
                             cs.get_hyperparameter('network_embedding:__choice__').default_value = default
 
+        # Disable CyclicLR until todo is completed.
+        if 'lr_scheduler' in self.named_steps.keys() and 'trainer' in self.named_steps.keys():
+            trainers = cs.get_hyperparameter('trainer:__choice__').choices
+            for trainer in trainers:
+                available_schedulers = self.named_steps['lr_scheduler'].get_available_components(
+                    dataset_properties=dataset_properties,
+                    exclude=exclude if bool(exclude) else None,
+                    include=include if bool(include) else None)
```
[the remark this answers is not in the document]
ravinkohli: yes because our `get_available_components` logic checks if `include` or `exclude` are `None` or not. I think though we dont need to use `get_available_components` as I only need the string names. I'll fix this.

autoPyTorch/pipeline/tabular_classification.py
```diff
@@ -289,6 +289,29 @@ def _get_hyperparameter_search_space(self,
...
+                    cs.add_forbidden_clause(ForbiddenAndConjunction(
+                        ForbiddenEqualsClause(cs.get_hyperparameter(
+                            f'trainer:{trainer}:use_snapshot_ensemble'), True),
+                        ForbiddenEqualsClause(cs.get_hyperparameter('lr_scheduler:__choice__'), cyclic_lr_name)
+                    ))
+                    cs.add_forbidden_clause(ForbiddenAndConjunction(
+                        ForbiddenEqualsClause(cs.get_hyperparameter(
+                            f'trainer:{trainer}:use_stochastic_weight_averaging'), True),
+                        ForbiddenEqualsClause(cs.get_hyperparameter('lr_scheduler:__choice__'), cyclic_lr_name)
+                    ))
```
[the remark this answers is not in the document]
ravinkohli: sure

autoPyTorch/pipeline/components/setup/network_embedding/__init__.py
```diff
@@ -148,71 +148,62 @@ def get_hyperparameter_search_space(
...
+                if include is not None and default not in include:
+                    raise ValueError("Provided {} in include, however, the dataset "
+                                     "is incompatible with it".format(include))
+                embedding = CSH.CategoricalHyperparameter('__choice__',
+                                                          ['NoEmbedding'],
+                                                          default_value=default)
+            else:
+                embedding = CSH.CategoricalHyperparameter('__choice__',
+                                                          list(available_embedding.keys()),
+                                                          default_value=default)
```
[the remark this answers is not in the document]
ravinkohli: let me do that after we are done with the experiments. I'll raise an issue.

ArlindKadra (approved):
I had one small question and except that everything looks good, especially refactoring to remove the code duplication.

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
ArlindKadra: He caught a bug, when there was a wrong implementation initially for se and swa when both activated and it was fixed, it was not fixed for the case of a non-cyclic scheduler. Hence now that you fixed it, it is different from the previous implementation since you had no if last epoch check to add the swa model only in the end, but you add 3 copies of it throughout.

ArlindKadra: Basically, now it is ok and correct.

autoPyTorch/pipeline/base_pipeline.py
```diff
@@ -295,6 +297,67 @@ def _get_hyperparameter_search_space(self,
...
+                    while True:
+                        try:
+                            cs.add_forbidden_clause(ForbiddenAndConjunction(
+                                ForbiddenEqualsClause(cs.get_hyperparameter(
+                                    'network_embedding:__choice__'), 'LearnedEntityEmbedding'),
+                                ForbiddenEqualsClause(cs.get_hyperparameter('encoder:__choice__'), encoder)
+                            ))
+                            break
+                        except ValueError:
+                            # change the default and try again
```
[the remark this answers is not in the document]
ravinkohli: we will get a value error in case `LearnedEntityEmbedding` was the default value for the `__choice__` hyperparameter

PR #401, opened 2022-03-03
[refactor] Separate all_nan_columns detection

nabenabe0928:
Refactoring


nabenabe0928:
Check
- test/test_pipeline/components/preprocessing/test_feature_preprocessor.py::TestFeaturePreprocessors::test_pipeline_fit_include[Nystroem-classification_numerical_and_categorical]

ravinkohli (changes requested):
The logic for all nans will be changed once the reg cocktails PR is merged. So I think this PR is not needed in the long term. Please close it if you agree.

PR #408, opened 2022-03-09
add modifications that are common for all experiments

ravinkohli:
Modifications needed for all experiments

1. Allows passing search space updates via `get_search_space_updates`
2. logs learning curves
3. does not raise error when incompatible update is sent to the pipeline, instead gives a warning
4. remove update from search space updates

### DIFF
--- a/autoPyTorch/pipeline/base_pipeline.py
+++ b/autoPyTorch/pipeline/base_pipeline.py
@@ -346,16 +346,18 @@ class BasePipeline(Pipeline):
                 cyclic_lr_name = 'CyclicLR'
                 if cyclic_lr_name in available_schedulers:
                     # disable snapshot ensembles and stochastic weight averaging
-                    cs.add_forbidden_clause(ForbiddenAndConjunction(
-                        ForbiddenEqualsClause(cs.get_hyperparameter(
-                            f'trainer:{trainer}:use_snapshot_ensemble'), True),
-                        ForbiddenEqualsClause(cs.get_hyperparameter('lr_scheduler:__choice__'), cyclic_lr_name)
-                    ))
-                    cs.add_forbidden_clause(ForbiddenAndConjunction(
-                        ForbiddenEqualsClause(cs.get_hyperparameter(
-                            f'trainer:{trainer}:use_stochastic_weight_averaging'), True),
-                        ForbiddenEqualsClause(cs.get_hyperparameter('lr_scheduler:__choice__'), cyclic_lr_name)
-                    ))
+                    snapshot_ensemble_hyperparameter = cs.get_hyperparameter(f'trainer:{trainer}:use_snapshot_ensemble')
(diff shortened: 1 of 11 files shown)


autoPyTorch/pipeline/base_pipeline.py
```diff
@@ -465,12 +465,12 @@ def _get_base_search_space(
     def _check_search_space_updates(self, include: Optional[Dict[str, Any]],
                                     exclude: Optional[Dict[str, Any]]) -> None:
         assert self.search_space_updates is not None
+        ignore_updates = []
         for update in self.search_space_updates.updates:
             if update.node_name not in self.named_steps.keys():
                 raise ValueError("Unknown node name. Expected update node name to be in {} "
                                  "got {}".format(self.named_steps.keys(), update.node_name))
             node = self.named_steps[update.node_name]
-            node_name = node.__class__.__name__
```
ArlindKadra, line 473: It was kinda better to have it as `node_name`, it is shorter, instead of putting the whole long version 4 times.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/encoding/__init__.py
```diff
@@ -86,15 +87,18 @@ def get_hyperparameter_search_space(self,
                                  "choices in {} got {}".format(self.__class__.__name__,
                                                                available_preprocessors,
                                                                choice_hyperparameter.value_range))
-            if len(choice_hyperparameter) == 0:
```
ArlindKadra, line 89: what was the case before that `choice_hyperparameter` was equal to 0? Since if there was a value range it should have not been 0?

ArlindKadra (commented):
I only have a few minor things, otherwise looks good :)

autoPyTorch/utils/hyperparameter_search_space_update.py
```diff
@@ -122,6 +122,29 @@ def append(self, node_name: str, hyperparameter: str, value_range: Union[List, T
                                                             value_range=value_range,
                                                             default_value=default_value,
                                                             log=log))
+    def remove_update(self, update: HyperparameterSearchSpaceUpdate) -> None:
+        """
+        removes an update
```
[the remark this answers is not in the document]
ravinkohli: yeah I'll update these things

autoPyTorch/pipeline/base_pipeline.py
ravinkohli: its an unintended code change

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/encoding/__init__.py
ravinkohli: yeah, I don't know why it was done this way. But it should be checking for the length of categorical columns

ArlindKadra (approved):

Issue #180, opened 2021-04-20
[memo] High memory consumption and the places of doubts


nabenabe0928:
I write down the current memory usage as a memo just in case when we encounter memory leak issues in the future.
This post is based on [the current implementation](https://github.com/automl/Auto-PyTorch/commit/a4e08e297dc298b53d5962f0f4b4e55b5e9ded8c).

When we run a dataset with the size of 300B, AutoPytorch consumes ~1.5GB and the followings are the major reason for the memory consumption:
|Source| Consumption [GB] | 
|-| :-:|
|Import modules| 0.35 |
|Dask Client | 0.35 |
|Logger | 0.4 |
| Running of context.Process in multiprocessing module | 0.4 |
| Model | 0 ~ inf |
|**Total**| 1.5 ~ inf |

When we run a dataset with the size of 300MB (400,000 instances x 80 features) such as Albert, AutoPytorch consumes ~1.7GB and the followings are the major reason for the memory consumption:
|Source| Consumption [GB] | 
|-| :-:|
|Import modules| 0.35|
|Dask Client | 0.35 |
|Logger | 0.4 |
| Dataset itself | 0.3 |
|self.categories in InputValidator| 0.3 |
| Running of context.Process in multiprocessing module | 0.4 |
| Model (e.g. LightGBM) | 0.4 ~ inf |
|**Total**| 2.5 ~ inf |

ArlindKadra:
Interesting :), I think the analysis in the future should also be extended to the following datasets:
 
 https://archive.ics.uci.edu/ml/datasets/covertype  
 https://archive.ics.uci.edu/ml/datasets/HIGGS
 https://archive.ics.uci.edu/ml/datasets/Poker+Hand
 
 They proved tricky.

nabenabe0928:
When we call the following, we consume 4GB even before the model instantiation:
```
python -m pytest test -vs -k <test_predict>
```

nabenabe0928:
FYI, when we use [Optuna](https://github.com/optuna/optuna) with a tiny model, we consume around only 150MB.
This module is also thread safe.

```
import optuna


def objective(trial):
    x0 = trial.suggest_uniform('x0', -10, 10)
    x1 = trial.suggest_uniform('x1', -10, 10)
    return x0 ** 2 + x1 ** 2


if __name__ == '__main__':
    study = optuna.create_study()
    study.optimize(objective, n_trials=5000, n_jobs=4)
```

[...]

PR #425, opened 2022-03-31
nn.Embedding to avoid OneHotEncoding all categorical columns

ravinkohli:
This PR replaces the `LearnedEntityEmbedding` with pytorch's `nn.Embedding` which implicitly one hot encodes categorical columns. This leads to a reduction in memory usage compared to the old version.

## Types of changes

## Motivation and Context

One hot encoding can lead to explosion in memory when the categories per column is high. Using `nn.Embedding` for such categorical columns will significantly reduce memory usage. Moreover, it is a more robust and simpler implementation of the embedding module. To do this, I have introduced a new pipeline step called `ColumnSplitter` (I am up for better name suggestions) which has `min_values_for_embedding` as a hyperparameter. 
It also makes minor changes which optimise some parts of the library. These include
1. We were loading the data from the datamanager and preprocessing some part of it when we needed the shape of the data after preprocessing. Loading data from disk is a time heavy process, and preprocessing even a small part of it is unnecessary. Now, the shape after preprocessing is passed from the `EarlyPreprocessing` node making it more efficient. 
2. remove `self.categories` from tabular feature validator which according to #180 takes a lot of memory. We dont really need to store all the categories anyways we only need `num_categories_per_col`.

## How has this been tested?

I have successfully run `example_tabular_classification` on `Australian` datasets where the default configuration allows us to verify the features introduced in this PR.


autoPyTorch/pipeline/components/setup/network_embedding/LearnedEntityEmbedding.py
```diff
@@ -94,28 +85,30 @@ def __init__(self, random_state: Optional[np.random.RandomState] = None, **kwarg
...
             for i in range(len(dataset_properties['categorical_columns'])
                            if isinstance(dataset_properties['categorical_columns'], List) else 0):
+                # currently as we dont have information about the embedding columns
+                # we search for more dimensions than necessary. This can be solved by
+                # not having `min_unique_values_for_embedding` as a hyperparameter and
+                # instead passing it as a parameter to the feature validator, which
+                # allows us to pass embed_columns to the dataset properties.
+                # TODO: test the trade off
+                # Another solution is to combine `OneHotEncoding`, `Embedding` and `NoEncoding` in one custom transformer.
+                # this will also allow users to use this transformer outside the pipeline
```
ravinkohli, line 111:
```suggestion
                # this will also allow users to use this transformer outside the pipeline, see [this](https://github.com/manujosephv/pytorch_tabular/main/pytorch_tabular/categorical_encoders.py#L132)
```

autoPyTorch/api/base_task.py
```diff
@@ -120,8 +120,8 @@ def get_search_updates(categorical_indicator: List[bool]):
         The search space updates like setting different hps to different values or ranges.
     """
 
-    has_cat_features = any(categorical_indicator)
-    has_numerical_features = not all(categorical_indicator)
+    # has_cat_features = any(categorical_indicator)
+    # has_numerical_features = not all(categorical_indicator)
```
theodorju, line 124: I think this should be removed.

autoPyTorch/pipeline/components/setup/network_embedding/LearnedEntityEmbedding.py
```diff
@@ -19,69 +19,59 @@
 class _LearnedEntityEmbedding(nn.Module):
     """ Learned entity embedding module for categorical features"""
 
-    def __init__(self, config: Dict[str, Any], num_input_features: np.ndarray, num_numerical_features: int):
+    def __init__(self, config: Dict[str, Any], num_categories_per_col: np.ndarray, num_features_excl_embed: int):
         """
         Args:
             config (Dict[str, Any]): The configuration sampled by the hyperparameter optimizer
             num_input_features (np.ndarray): column wise information of number of output columns after transformation
```
theodorju, line 26: I think `num_input_features` should be replaced with `num_categories_per_col (np.ndarray): number of categories for categorical columns that will be embedded`

autoPyTorch/pipeline/tabular_classification.py
```diff
@@ -289,6 +277,7 @@ def _get_pipeline_steps(
             ("imputer", SimpleImputer(random_state=self.random_state)),
             # ("variance_threshold", VarianceThreshold(random_state=self.random_state)),
             # ("coalescer", CoalescerChoice(default_dataset_properties, random_state=self.random_state)),
+            ("column_splitter", ColumnSplitter(random_state=self.random_state)),
```
theodorju, line 280: I think the docstring of the class should be updated to also include `column_splitter` as a step.

autoPyTorch/pipeline/tabular_regression.py
```diff
@@ -235,8 +232,9 @@ def _get_pipeline_steps(
 
         steps.extend([
             ("imputer", SimpleImputer(random_state=self.random_state)),
-            ("variance_threshold", VarianceThreshold(random_state=self.random_state)),
-            ("coalescer", CoalescerChoice(default_dataset_properties, random_state=self.random_state)),
+            # ("variance_threshold", VarianceThreshold(random_state=self.random_state)),
+            # ("coalescer", CoalescerChoice(default_dataset_properties, random_state=self.random_state)),
+            ("column_splitter", ColumnSplitter(random_state=self.random_state)),
```
theodorju, line 237: Same as `tabular_classification.py`, it would be nice to add this step in the docstring as well.

theodorju (commented):
As discussed in the meeting, I reviewed the changes.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/column_splitting/ColumnSplitter.py
```diff
@@ -0,0 +1,83 @@
...
+                    self.special_feature_types['embed_columns'].append(column)
+                    # we only care about the categories for columns to be embedded
+                    self.num_categories_per_col.append(categories_per_column)
+                else:
+                    self.special_feature_types['encode_columns'].append(column)
+
+        return self
+
+    def transform(self, X: Dict[str, Any]) -> Dict[str, Any]:
+        if self.num_categories_per_col is not None:
```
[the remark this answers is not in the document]
ravinkohli: it will be none when there were no categorical column, see line [38](https://github.com/automl/Auto-PyTorch/blob/95733585f38bb6d4d7c92782a56faebc05710693/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/column_splitting/ColumnSplitter.py#L38)

autoPyTorch/api/base_task.py
```diff
@@ -267,7 +249,8 @@ def __init__(
 
         self.input_validator: Optional[BaseInputValidator] = None
 
-        self.search_space_updates = search_space_updates if search_space_updates is not None else get_search_updates(categorical_indicator)
+        # if search_space_updates is not None else get_search_updates(categorical_indicator)
```
ravinkohli, line 252:
```suggestion
```

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/column_splitting/ColumnSplitter.py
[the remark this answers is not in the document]
theodorju: Hm, but line [38](https://github.com/automl/Auto-PyTorch/blob/95733585f38bb6d4d7c92782a56faebc05710693/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/column_splitting/ColumnSplitter.py#L38) sets this to an empty list and `[] is not None` returns `True`.

ravinkohli:
This branch will be merged to reg_cocktails. Therefore, this PR has been shifted to #451

PR #435, opened 2022-06-02
[ADD] Reduce search space for embedding

ravinkohli:
This PR reduces the number of hyperparameters when using the embedding module from 1 per categorical column to 2 for all of them. The new technique to calculate the dimension reduction for each column is based on a heuristic used in AutoGluon.

### DIFF
--- a/autoPyTorch/pipeline/components/setup/network_embedding/LearnedEntityEmbedding.py
+++ b/autoPyTorch/pipeline/components/setup/network_embedding/LearnedEntityEmbedding.py
@@ -3,8 +3,9 @@ from typing import Any, Dict, List, Optional, Union
 
 from ConfigSpace.configuration_space import ConfigurationSpace
 from ConfigSpace.hyperparameters import (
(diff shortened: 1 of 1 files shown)


theodorju:
Some of my (minor) findings; all on `LearnedEntityEmbedding.py`:
- On line 1: `ceil` is imported but never used.
- On line 2: `List` is imported but never used.
- Method `get_num_output_dimensions` has no type hints. Also, the docstring could possibly be a bit enhanced to mention that the method returns 1 for numerical features.
- On line 68: I think there is no need to initialize `x_pointer` there. 
- On `get_hyperparameter_search_space` argument `embedding_size_factor`, I think it might be better to sort `value_range` values to enhance code readability.

nabenabe0928 (approved):
Hi thanks for the PR, it looks good to me.

theodorju:
@ravinkohli yeap, thanks for the feedback. I'll add them directly to the line of code.

